PR #51, opened 2017-02-20
Fixes to custom resolver docs

nathforge:
* Changed super class call.
* Updated config example.

### DIFF
--- a/docs/advanced_patterns.rst
+++ b/docs/advanced_patterns.rst
@@ -60,7 +60,7 @@ The following python module template can be copied and used:
   class CustomResolver(Resolver):
 
       def __init__(self, *args, **kwargs):
-        super(EnvironmentVariable, self).__init__(*args, **kwargs)
+        super(CustomResolver, self).__init__(*args, **kwargs)
 
       def resolve(self):
         """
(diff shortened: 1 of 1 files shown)


jamesroutley:
Hey @nathforge thanks for the PR! I've added a comment but it looks good to me apart from that.

James

docs/advanced_patterns.rst
```diff
@@ -83,11 +83,10 @@ This resolver can be used in a stack config file with the following syntax::
 
   template_path: <...>
   parameters:
-    param1:
-      <your resolver name>: <value>  # <your resolver name>
-                                     # should match Hook.name.
-                                     # <value> will be passed to the
-                                     # resolver's resolve() method.
+    param1: !<your_resolver_name>: <value> # <your resolver name> is the lower camel-case version
```
theseanything: this needs to be `!<your_resolver_name> <value>` - without the colon between resolver name and value

theseanything:
@nathforge thanks! Just a small error in the syntax, but good spot!

nathforge:
Thanks both, have updated!

PR #53, opened 2017-02-21
Update circle.yml to deploy docs based on tags

theseanything:
Changed the circleci deployment to only update docs at /docs/ with a new tag version and /dev/doc/ with the latest from master branch.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -17,12 +17,10 @@ test:
     - make docs
 deployment:
     production:
-        branch: master
+        tag: /v*/
         commands:
-            - make docs
             - aws s3 sync ~/sceptre/docs/_build/html s3://sceptre.cloudreach.com/docs/ --delete
     develop:
-        branch: develop
+        branch: master
         commands:
-            - make docs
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -17,12 +17,10 @@ test:
     - make docs
 deployment:
     production:
-        branch: master
+        tag: /v*/
         commands:
-            - make docs
             - aws s3 sync ~/sceptre/docs/_build/html s3://sceptre.cloudreach.com/docs/ --delete
```
jamesroutley: @theseanything Sorry - I'd written this comment a while ago but hadn't submitted it. Still getting used to GH I guess.

theseanything: Ah do you mean - that only certain tagged commits get push to docs or make sure tagged commits ALSO get put on dev?

jamesroutley: Make sure tagged commits also get pushed to dev

theseanything: Yup so thats what would happen - only commits with `v*` tags would go to `/docs` and all commits regardless of tags goes to `/dev/docs`

jamesroutley: cool

Issue #63, opened 2017-02-27
Remove python module support


jamesroutley:
Sceptre officially supports use via the CLI and as a Python module. Use via the CLI is far more common (I'm not sure if anyone uses Sceptre as a Python module). We guarantee that both our CLI and public Python APIs won't change. Guaranteeing the Python API leads to two problems:

- It makes it difficult to refactor the inner workings of Sceptre
- It makes our versioning unusual, as we are effectively versioning two independent APIs with a single version number. If we commit a breaking change to Sceptre's Python internals, we must bump our major version number, even though users of Sceptre as a CLI tool (most of our users) would see no change.

This would be a relatively superficial change. Our Python function and classes would still be public, and people could still use Sceptre as a module, but we wouldn't officially support it or guarantee its API.

## Implementation

Remove references to use as a Python module from the documentation.

## Pros
- Reduces our public API
- Makes it easier to develop Sceptre as we have fewer contracts to uphold
- Makes our versioning less strange

## Cons
- Discontinued support for Python module

theseanything:
Not sure it would be fair to say no ones using it - I've seen it used a lot in custom hooks. I think we still need to strive for consistency as it make developing a lot easier. 

Do you see this just in the scope of documentation and versioning - or any code structure changes you see happening? 

Maybe good to look at Hashicorp Vault and see how they manage a CLI and restful API in the same code base. Possibly could do a complete separation with sceptre-core and sceptre-cli? We also need to consider how this restricts future extension to the project e.g. what if someone wants to wrap sceptre with flask etc...

What issues have we run into so far with this supporting both?

emiliogarcia:
In some customers we are seriously considering using the module. I would MUCH rather prefer you don't remove it.

jamesroutley:
Hm okay if we keep it we still need some way of fixing the versioning issue. Separating the CLI from the Python core would do that as @seanrankine mentioned. Sean how do Hashicorp do it? 

I guess we could have two repositories, `sceptre` and `sceptre-core`. `sceptre-core` would implement the Python module, and `sceptre` would depend on `sceptre-core` and implement the CLI. Each would have their own version numbers. It does add complexity though.

I think we still need the freedom to be able to change the python internals. I imagine we'd go though `sceptre-core` version number relatively quickly, and we should warn users that using the CLI will likely give more stable results.

henrybell:
@jamesroutley I agree with @emiliogarcia and think this would be a nice way to do it.

[...]

PR #72, opened 2017-03-03
Rename cfn attribute of Template to body

theseanything:
Simply renaming `cfn` to `body` in the Template class as it better defines the attribute.


jamesroutley:
Happy to put this through once tests pass but I'd like to note that this emphasises the issue in #63. This is a breaking change on our Python API so should be a major version bump according to semver.

jamesroutley:
(clarifying point) I don't think we should do a major version bump, I think a patch is suitable. I guess our version num should change w.r.t. the CLI API until we sort this out.

PR #75, opened 2017-03-06
Refactor upload to s3

theseanything:
Refactored the upload_to_s3 and create_bucket function in Template class. Introduced a bucket_exists function to encapsulate some the create_bucket logic.

### DIFF
--- a/sceptre/template.py
+++ b/sceptre/template.py
@@ -151,20 +151,16 @@ class Template(object):
         :raises: botocore.exception.ClientError
 
         """
+        self.logger.debug(
+            "%s - Attempting to find template bucket '%s'",
+            self.name, bucket_name
+        )
         try:
-            self.logger.debug(
-                "%s - Attempting to find template bucket '%s'",
-                self.name, bucket_name
-            )
             connection_manager.call(
                 service="s3",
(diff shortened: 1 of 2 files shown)


theseanything:
Missing s3.bucket_lock - will add and reopen.

sceptre/template.py
```diff
@@ -132,9 +134,47 @@ def upload_to_s3(
...
+        :type bucket_name: str
+        :param connection_manager: The connection manager used to make
+            AWS calls.
+        :type connection_manager: sceptre.connection_manager.ConnectionManager
+        :returns: Boolean whether the bucket exists
+        :rtype: bool
+        :raises: botocore.exception.ClientError
+
+        """
+        try:
```
jamesroutley: I think it's conventional for Python `try` blocks to only contain the code we're expecting might fail. Could we switch this to:

```Python
log("Attempting to find bucket")
try:
    connection_manager.call(...)
except ClientError:
    ...
log("Found template bucket")
return True
```

(I realise this PR is just copy-pasting but I think we might as well fix this now)

theseanything: Yup agree! will change

PR #79, opened 2017-03-15
Parallelise circle

jamesroutley:
This PR makes use of CircleCI's parallel tests [feature](https://circleci.com/docs/1.0/parallel-manual-setup/#using-configuration-files-modifier) to speed up our integration tests. 

Circle's simple parallelism is file based. You supply a command and a list of files to run that command on. In our case, the command is `behave`, and the list of files is the result of the glob `integration-tests/features/*.feature`.

This command has to be run from the root directory, `sceptre/`. Previously, our integration tests were run from `sceptre/integration-tests`, where we store the config files and templates used by sceptre to run the tests. To get `behave` to run from the root dir, the integration tests have been modified so all sceptre commands are explicitly supplied with the sceptre directory to use. 

For example:
```
sceptre create-stack dev vpc
```

Has been changed to:
```
sceptre --dir /path/to/sceptre/integration-tests dev vpc
```

This parallelisation means our int tests are run over 4 containers rather than 1. This has sped up the integration tests from 22 mins to 7 mins, and the overall test from 25 mins to 10 mins. (all times approximate)

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -9,8 +9,9 @@ dependencies:
     - pip install -r requirements_tests.txt
     - python setup.py install
 test:
-  post:
+  override:
     - make lint
+    - make test-all
     - behave:
         parallel: true
         files:
--- a/circle.yml
+++ b/circle.yml
@@ -12,7 +12,7 @@ test:
   override:
     - make lint
     - make test-all
-    - behave --junit --junit-directory:
+    - behave --junit --junit-directory $CIRCLE_TEST_REPORTS/behave/junit.xml:
         parallel: true
         files:
           - integration-tests/features/*.feature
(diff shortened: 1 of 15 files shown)


circle.yml
```diff
@@ -7,13 +7,15 @@ dependencies:
   override:
     - pip install -r requirements.txt
     - pip install -r requirements_tests.txt
-test:
-  pre:
     - python setup.py install
```
jamesroutley: Running `python setup.py install` as a pre-test step only runs it in a single container. I moved it to `dependences/override` so it's run on all containers.

theseanything: why not `make install` ?

jamesroutley:
@seanrankine I've switched it to `make install`

Issue #78, opened 2017-03-13
Better Installation instructions


johnculkin:
Sceptre needs some better installation instructions. My recent experience installing Sceptre from Github wasn't as smooth as I would have liked. 

I'd like something that talks about the recommended tools to install beforehand - Ex: virtualenv/virtualenvwrapper - Troposphere (with awacs). I could create a shell script to run these - thoughts? 

I'll write something up and create a PR

theseanything:
Thanks @johnculkin! I would list the steps, rather than creating a script - as the steps I ran through with you were recommended, but not mandatory to get sceptre running.

Those steps were - for MacOS:
`brew install python`
`pip install virtualenv`
`pip install virtualenvwrapper`
(add lines to bashrc)
`mkvirtualenv sceptre`
`pip install sceptre`
`pip install troposphere` - if your using troposphere templates. (Optional!)

johnculkin:
ok I like it -  Here is my first quick and dirty attempt - I'll see if Mr Mossman would like to test it for me 

`brew install python

pip install virtualenv
pip install virtualenvwrapper

mkdir ~/virtualenvs

cat <<EOF >> ~/.bash_profile 
export WORKON_HOME=~/virtualenvs
source /usr/local/bin/virtualenvwrapper.sh
EOF

source ~/.bash_profile

mkvirtualenv venv1
workon venv1

pip install troposphere
pip install awacs
pip install sceptre`

jamesroutley:
I think we'd do better to just link to a proper Python install guide than rewrite it here. Something like [this](http://docs.python-guide.org/en/latest/starting/installation/) and [this](http://docs.python-guide.org/en/latest/dev/virtualenvs/) would be more thorough and more likely to be kept up to date.

theseanything:
+1 for link to proper install guide - but we may want to look at adding a note the readme, saying this the ideal way for running sceptre etc...

PR #80, opened 2017-03-15
Update installation docs [skip-ci]

jamesroutley:
#78

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -6,6 +6,8 @@
 Installation
 ============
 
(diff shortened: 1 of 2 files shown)


docs/installation.rst
```diff
@@ -6,13 +6,16 @@
 Installation
 ============
 
+This assumes that you have Python installed. A thorough guide on installing Python can be found `here <http://docs.python-guide.org/en/latest/starting/installation/>`_. Sceptre works best when run in a virtual environment. Notes on installing and setting up ``virtualenv`` can be found `here <http://docs.python-guide.org/en/latest/dev/virtualenvs/>`_.
```
theseanything: not keen on the wording - 

> Sceptre works best

would change to something like the 

> We strongly recommend using a virtualenv for installation of Sceptre.

jamesroutley:
@seanrankine switched.

PR #82, opened 2017-03-16
Respect --dir when loading custom resolvers and hooks

connec:
Rather than always loading custom resolvers and hooks from the current directory, they should instead be loaded from the sceptre dir along with config and templates.

### DIFF
--- a/sceptre/config.py
+++ b/sceptre/config.py
@@ -231,7 +231,7 @@ class Config(dict):
         self.add_yaml_constructors(
             resolvers_folder, Resolver, resolver_constructor_factory
         )
-        external_resolver_folder = os.path.join(os.getcwd(), "resolvers")
+        external_resolver_folder = os.path.join(self.sceptre_dir, "resolvers")
         self.add_yaml_constructors(
             external_resolver_folder, Resolver, resolver_constructor_factory
         )
@@ -270,7 +270,7 @@ class Config(dict):
             library_hook_folder, Hook, hook_constructor_factory
         )
 
-        project_hook_folder = os.path.join(os.getcwd(), "hooks")
+        project_hook_folder = os.path.join(self.sceptre_dir, "hooks")
         self.add_yaml_constructors(
             project_hook_folder, Hook, hook_constructor_factory
         )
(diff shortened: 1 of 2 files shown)


sceptre/config.py
```diff
@@ -231,7 +231,7 @@ def resolver_constructor_factory(node_class):
         self.add_yaml_constructors(
             resolvers_folder, Resolver, resolver_constructor_factory
         )
-        external_resolver_folder = os.path.join(os.getcwd(), "resolvers")
+        external_resolver_folder = os.path.join(str(self.sceptre_dir), "resolvers")
```
connec: I had to use `str(...)` to pass the check in `get_subclasses`, which expects a `str` argument. Let me know if there's a better way of dealing with this.

jamesroutley: Am I correct in thinking that's because `type(self.sceptre_dir) == unicode`?

jamesroutley: Yep so (assuming it is the unicode thing), we check that the directory passed in is a `str`. In fact it should check to make sure it's a `basestring` to avoid encoding issues like this. 

https://github.com/cloudreach/sceptre/blob/master/sceptre/helpers.py#L221

theseanything: The type casting shouldn't not be required with the change proposed in PR #83

sceptre/config.py
```diff
@@ -270,7 +270,7 @@ def hook_constructor_factory(node_class):
             library_hook_folder, Hook, hook_constructor_factory
         )
 
-        project_hook_folder = os.path.join(os.getcwd(), "hooks")
+        project_hook_folder = os.path.join(str(self.sceptre_dir), "hooks")
```
theseanything: Again remove the type casting.

theseanything:
This cannot be merged until PR #83 is merged.

connec:
I'll hold off rebase+updates until #83 is merged.

connec:
I've updated this now that #83 has been merged. Turns out there was a broken test which I've now fixed 😅

connec:
🎉 

What's the release cycle likely to be for this? I suppose it's either a breaking change or a bug fix 😄

jamesroutley:
I'll push it when the tests pass

jamesroutley:
@connec it's up: https://pypi.python.org/pypi/sceptre/1.1.1

connec:
Awesome, thanks!

PR #83, opened 2017-03-16
Update string type check in get_subclasses to account for unicode

connec:
The string check in `helpers.get_subclasses` was checking against `str`, meaning that passing a unicode string would fail.

The simple fix is to change the `isinstance` check to match against `basestring`.

**Note:** This isn't Python 3 compatible, but I didn't see any evidence that it's supported currently. Let me know if that's not the case.

### DIFF
--- a/sceptre/helpers.py
+++ b/sceptre/helpers.py
@@ -218,10 +218,12 @@ def get_subclasses(class_type, directory=None):
     :returns: A dict of classes found.
     :rtype: dict
     """
-    if not isinstance(directory, str):
-        raise TypeError("'directory' object should be a str")
+    try:
+        glob_expression = os.path.join(directory, "*.py")
+    except (AttributeError, TypeError):
+        raise TypeError("'directory' object should be a string")
 
-    module_paths = glob.glob(os.path.join(directory, "*.py"))
+    module_paths = glob.glob(glob_expression)
 
     sys.path.append(directory)
 


jamesroutley:
We do support py3 actually - @seanrankine can you take a look at this [line of code](https://github.com/cloudreach/sceptre/blob/master/sceptre/helpers.py#L221)? What's the type check for? Tracing the code back I don't think there's any way `directory` could not be a string, and type checking [isn't really Python's game](http://www.voidspace.org.uk/python/articles/duck_typing.shtml). Also the default value for `directory` is `None`, which would fail the type check.

I'm wondering if we should just remove the line, but there may be some reason for its inclusion I'm not seeing.

jamesroutley:
Thanks for your PRs by the way @connec!

theseanything:
So technically this correct it should be basestring to include unicode - but you right @jamesroutley, there should be no type checking. The proper way would be to use a try, except error handling.

sceptre/helpers.py
```diff
@@ -218,8 +218,8 @@ def get_subclasses(class_type, directory=None):
     :returns: A dict of classes found.
     :rtype: dict
     """
-    if not isinstance(directory, str):
-        raise TypeError("'directory' object should be a str")
+    if not isinstance(directory, basestring):
+        raise TypeError("'directory' object should be a string")
```
theseanything: Remove entirely.

sceptre/helpers.py
```diff
@@ -218,8 +218,8 @@ def get_subclasses(class_type, directory=None):
...
 
     module_paths = glob.glob(os.path.join(directory, "*.py"))
```
theseanything: Wrap above and below in try/except clause - and catch a type error.

sceptre/helpers.py
```diff
@@ -218,10 +218,12 @@ def get_subclasses(class_type, directory=None):
     :returns: A dict of classes found.
     :rtype: dict
     """
-    if not isinstance(directory, str):
-        raise TypeError("'directory' object should be a str")
+    try:
+        glob_expression = os.path.join(directory, "*.py")
+    except (AttributeError, TypeError):
```
connec: Python 2 and 3 throw different exceptions when `directory` is not a string, so I covered both.

```sh
# Python 2
>>> os.path.join(1, '')
AttributeError: 'int' object has no attribute 'endswith'

# Python 3.5.2
>>> os.path.join(1, '')
TypeError: join() argument must be str or bytes, not 'int'
```

connec:
@jamesroutley I've updated with a Python 2+3 compatible `try`/`except` version. That said, I get 18 test failures on master when running in Python 3.5.2.

connec:
Any chance this can go through? We'd love to have #82 for a workflow we're building around sceptre.

jamesroutley:
Yep - sorry about the delay

connec:
No worries! Thanks.

PR #102, opened 2017-04-25
Fix dependency namming when using dependencies

padilo:
This is how it works now. I don't see the point to have here the environment, but maybe it can be optional.


theseanything:
Thanks for this @padilo, good catch!

PR #105, opened 2017-05-04
user_variables mentioned

padilo:
This is an undocumented feature.


jamesroutley:
Hey @padilo. Thanks for the PR! Unfortunately I don't think this isn't an undocumented feature, but an implementation detail of how we let users pass variables in from the CLI. It's not meant to be used in this way, and I don't think it'll behave as expected.

PR #114, opened 2017-05-13
added mfa support to role assumption

### LINKED ISSUE #104: MFA support for iam_role
It would be great to be able to provide iam roles that require MFA authentication in the environment config for Sceptre to switch to.
If you are open for the option I'd be happy to help out with this.

As a proposal for the new behaviour for assuming the role:
1. Try to assume to role without providing MFA information. This will either work, in which case there is nothing to do, or it will error (sadly it doesn't tell us if it's an MFA related error or just lack of permissions).
2. If there was an AccessDenied error (which we get for lack of MFA), it will contain the arn of the current user, which contains the account id and the username. This is enough to construct the MFA device for the user since it's `arn:aws:iam::{account_id}:mfa/{username}`.
3. Prompt the user for an MFA code
4. Attempt to assume the role again, providing the MFA token and device from `2.` and `3.`
5. Either succeed or inform the user of the error.

What do you guys think?

mitom:
Closes #104 

It adds a new config in the environment `require_mfa` (boolean). When turned on and there is an `iam_role` set it will try to get the current user's information using the `get_user` api and builds the virtual MFA device's arn. When `get_user` errors due to MFA limitation, it will still contain enough information to create the MFA arn.

### DIFF
--- a/docs/environment_config.rst
+++ b/docs/environment_config.rst
@@ -13,6 +13,8 @@ An environment config file is a yaml object of key-value pairs configuring Scept
 
 - `iam_role`_ *(optional)*
 
+- `require_mfa`_ *(optional)*
+
 - `project_code`_ *(required)*
 
 - `region`_ *(required)*
@@ -30,6 +32,11 @@ iam_role
 ````````
 
 The ARN of a role for Sceptre to assume before interacting with the environment. If not supplied, Sceptre uses the user's AWS CLI credentials.
+
+require_mfa
+````````
+
+Whether or not the `iam_role` given requires MFA to assume it. It is only used when `iam_role` is set. If not set, it defaults to `false`.
 
 project_code
 ````````````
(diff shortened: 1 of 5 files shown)


mitom:
I didn't incorporate any way to persist the token between requests as it was brought up in #104 @seanrankine . Do you have any suggestions to solve that? My best is so far a file-cache but generally I'd be against that path.

theseanything:
@mitom thanks for this! Was wondering if there was away to not need the "mfa_required" flag in stack config? Is it possible to detect it automatically or ask for mfa after trying to assume without?

mitom:
@theseanything That was my original idea and I'd prefer it as well but I found it confusing. The error for not having the permission to assume a role and for requiring MFA appear to be exactly the same. I found that even during testing this I had to verify that I can assume the role by using an other script to check because I had trouble figuring out which part was the culprit for a problem.
If you think it's okay to simply ask for MFA and then error I'm happy to update it.

mitom:
any update on this?

theseanything:
Is this issue alleviated with return of profile support and doing assume roles via profiles?

patrickli:
I feel this PR is no longer useful with profile support.

theseanything:
Issue solved with #143

Issue #116, opened 2017-05-15
Deprecate project_variables resolver


jamesroutley:
## Overview

I think we should deprecate the [`project_variables`](https://sceptre.cloudreach.com/docs/stack_config.html#project-variables) resolver for the following reasons:

1. It doesn't fulfil it's intended use. The resolver was originally added to allow Sceptre users a way to store configuration items shared by several stacks in a single place. The current implementation only allows for a single configuration item to be passed to each stack (#115), which isn't useful if you want to have more than one piece of common config. 
2. It's syntax is overly verbose. You have to namespace your environment and stack, leading to highly nested files.
3. It confuses users #48 . By introducing yet another way to dynamically add configuration to stack config files (alongside user variables and environment variables), it is unclear which the 'right' method is.

## Todo

Throw a deprecation warning when the `project_variables` resolver is used.

## Implementation

The warning should be done with the [`warnings`](https://docs.python.org/2/library/warnings.html) library. A [deprecation warning](https://docs.python.org/2/library/exceptions.html#exceptions.DeprecationWarning) should be raised. Because deprecation warnings are ignored by default, they will need to be [switched on](http://stackoverflow.com/a/20960427). The warning should print `The project_variables resolver is being deprecated, and will be removed in Sceptre version 2.0`. 

Bonus points for printing the warning in yellow - we use [`colorama`](https://pypi.python.org/pypi/colorama) for colouring text.

Code should be added to the beginning of the `resolve()` [function](https://github.com/cloudreach/sceptre/blob/master/sceptre/resolvers/project_variables.py#L27).

## Definition of done

- Using the `project_variables` resolver should print out a warning that it'll be deprecated in version 2.0
- A deprecation notice should be added to the relevant section of the documentation

amongil:
I would like to take this one.

jamesroutley:
@albertoal Nice one - thanks!

PR #117, opened 2017-05-16
Deprecate project_variables resolver #116

amongil:
Added deprecation warning for the "project_variables" resolver as on Issue #116 and updated documentation accordingly.

### DIFF
--- a/sceptre/resolvers/project_variables.py
+++ b/sceptre/resolvers/project_variables.py
@@ -4,6 +4,8 @@ import os
 import yaml
 
 from sceptre.resolvers import Resolver
+from warnings import warn, simplefilter
+from colorama import Fore, init
 
 
 class ProjectVariables(Resolver):
@@ -24,6 +26,12 @@ class ProjectVariables(Resolver):
         :returns: Value at argument location.
         :rtype: str
         """
+        init(autoreset=True)
+        simplefilter('always', DeprecationWarning)
+        warn(Fore.YELLOW + (
+                "The project_variables resolver is being deprecated, and will "
+                "be removed in Sceptre version 2.0"
+             ), DeprecationWarning, stacklevel=2)
         file_path = os.path.join(
             self.environment_config.sceptre_dir,
             self.argument
--- a/sceptre/resolvers/project_variables.py
+++ b/sceptre/resolvers/project_variables.py
@@ -2,10 +2,11 @@
 
 import os
 import yaml
+from warnings import warn, simplefilter
+
+from colorama import Fore, Style
 
 from sceptre.resolvers import Resolver
-from warnings import warn, simplefilter
-from colorama import Fore, init
 
 
 class ProjectVariables(Resolver):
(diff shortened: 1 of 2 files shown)


sceptre/resolvers/project_variables.py
```diff
@@ -4,6 +4,8 @@
 import yaml
 
 from sceptre.resolvers import Resolver
+from warnings import warn, simplefilter
```
theseanything: Does warnings need to be added to requirements.txt?

sceptre/resolvers/project_variables.py
```diff
@@ -24,6 +26,12 @@ def resolve(self):
         :returns: Value at argument location.
         :rtype: str
         """
+        init(autoreset=True)
+        simplefilter('always', DeprecationWarning)
+        warn(Fore.YELLOW + (
+                "The project_variables resolver is being deprecated, and will "
```
theseanything: The project_variables is ~~being~~ deprecated, and will be remove in later versions. (I don't want to make any guarantees of which version it will go.) Maybe also nice to offer an alternative solution to using project_variables. - but know its tricky as its own purpose is unclear. Possibly use --vars instead?

sceptre/resolvers/project_variables.py
```diff
@@ -24,6 +26,12 @@ def resolve(self):
         :returns: Value at argument location.
         :rtype: str
         """
+        init(autoreset=True)
```
jamesroutley: `autoreset` is a really nice feature, which I didn't actually know about. However, we already run `colorama.init()` in [cli.py](https://github.com/cloudreach/sceptre/blob/master/sceptre/cli.py#L123) - could you remove the `init()` from here and add a `Style.RESET_ALL` at the end of the warning?

sceptre/resolvers/project_variables.py
```diff
@@ -4,6 +4,8 @@
 import yaml
 
 from sceptre.resolvers import Resolver
+from warnings import warn, simplefilter
+from colorama import Fore, init
```
jamesroutley: Could you move these imports up? The general rule we follow is:

```
python stdlib imports
<newline>
external library imports
<newline>
sceptre specific imports
```

So something like:
```
import os
import yaml
from warnings import warn, simplefilter

from colorama import Fore, init
  		  
from sceptre.resolvers import Resolver
```

sceptre/resolvers/project_variables.py
jamesroutley: I like that you've used brackets to join the long string here. Could we de-indent lines `32-34` by 4 spaces?

sceptre/resolvers/project_variables.py
```diff
@@ -24,6 +26,12 @@ def resolve(self):
         :returns: Value at argument location.
         :rtype: str
         """
+        init(autoreset=True)
+        simplefilter('always', DeprecationWarning)
```
jamesroutley: Could we use double quotes (") instead of single (') - both are equally valid and which one you choose is a personal preference but we use doubles elsewhere in the codebase.

sceptre/resolvers/project_variables.py
amongil: I think so. I will modify that aswell.

sceptre/resolvers/project_variables.py
amongil: I tried that and the lint complained about it. This is the only indentation and parenthesis setup that made the lint come clear. How should we go about this?

sceptre/resolvers/project_variables.py
jamesroutley: nop stdlib

sceptre/resolvers/project_variables.py
jamesroutley: Try:

```
warn(
    "{0}The project_variables ..."
    "be removed... {1}".format(Fore.YELLOW, Style.RESET_ALL)
)
```

I prefer [string.format()](https://docs.python.org/2/library/string.html#string.Formatter.format) over string concatination with `+`, and we use it more through the rest of the codebase. 

+1 for linting!

sceptre/resolvers/project_variables.py
```diff
@@ -2,6 +2,9 @@
 
 import os
 import yaml
+from warnings import warn, simplefilter
+
```
theseanything: extra line?

jamesroutley: We tend to put a newline in between stdlib imports and external imports [e.g.](https://github.com/cloudreach/sceptre/blob/master/sceptre/cli.py#L17)

jamesroutley:
lgtm

sceptre/resolvers/project_variables.py
theseanything: oh didn't realise warnings was inbuilt

Issue #120, opened 2017-05-17
Stack config key `protect` raises exception when used


afdezl:
When using the `protect` key in a stack config, the exception `AttributeError: 'Stack' object has no attribute 'full_stack_name'` is raised.  The decorator method uses a deprecated version of the stack name attribute. Should be replaced with `self.name` in the helpers file.

jamesroutley:
@afdezl - thanks for pointing that out - you got time to submit a PR? dw if not - let me know and I can put one in

jamesroutley:
Haha looks like you just put one in - thanks!

afdezl:
Done.

PR #122, opened 2017-05-18
Simplify execution protection

jamesroutley:
This pull request aims to simplify execution protection. It is motivated by #120 and it's PR #121, and is a "proposal" PR.

The previous implementation used a decorator, `@execution_protection`, which was applied to all "unsafe" commands (`create`, `delete` etc). The decorator code was very tightly coupled with the stack code. The decorator function assumed that: 

- the first argument passed was `self` (i.e. it could only be applied to an object)
- the object has a `dict` attribute `config`
- the object has an attribute `name`

This tight coupling makes the function difficult to test. The previous implementation mocked out `Stack` completely, so when changes were made to `Stack` (`full_stack_name` -> `name`), the fact that this decorator still used the old value wasn't picked up (#121)

This new implementation removes the decorator, and replaces it with a method. This new method: 

- removes the coupling (and the `Mock`)
- is easier to test
- is easier to understand

### DIFF
--- a/tests/test_stack.py
+++ b/tests/test_stack.py
@@ -857,7 +857,7 @@ environment_config={'key': 'val'}, connection_manager=connection_manager)"
 
     def test_protect_execution_without_explicit_protection(self):
         self.stack._config = {}
-        # Function should do nothing if protect == False
+        # Function should do nothing if protect isn't explicitly set
         self.stack._protect_execution()
 
     def test_protect_execution_with_protection(self):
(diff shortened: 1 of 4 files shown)


tests/test_stack.py
```diff
@@ -849,6 +850,16 @@ def test_get_role_arn_with_role(self):
...
+    def test_protect_execution_without_protection(self):
+        self.stack._config = {"protect": False}
+        # Function should do nothing if protect == False
+        self.stack._protect_execution()
+
+    def test_protect_execution_with_protection(self):
+        self.stack._config = {"protect": True}
+        with pytest.raises(ProtectedStackError):
+            self.stack._protect_execution()
+
```
theseanything: Can we add a test where "protect" isn't in the config, as thats typical for most stacks.

jamesroutley: added - `858`

jamesroutley: @seanrankine I've added the changes

PR #123, opened 2017-05-19
Add confirmation step when deleting a stack.

### LINKED ISSUE #61: Add confirmation step when deleting
Sceptre should ask for user confirmation when deleting stacks or environments. A CLI flag `--yes/-y` will be added, allowing scriptable deletion of resources. This confirmation will only happen on the CLI, not when delete commands are run from 

## Implementation

- Add a [confirmation prompt](http://click.pocoo.org/5/prompts/#confirmation-prompts) with Click to the function `sceptre.cli.delete_stack`.
- Add a [Click option](http://click.pocoo.org/5/options/) `-y`, which skips the confirmation prompt and runs `env.stacks[stack].delete()`.

## Pros
- Syntax is in keeping with Unix standards (e.g. `yum install`)

## Cons

rmbennett:
Fixes #61  

To automate your stack deletions  add --yes to your delete stack command.

This uses [Click: Yes Parameters](http://click.pocoo.org/5/options/#yes-parameters) [further detail can be found here](http://click.pocoo.org/5/api/#click.confirmation_option)

### DIFF
--- a/docs/get_started.rst
+++ b/docs/get_started.rst
@@ -130,7 +130,7 @@ Delete Stack
 
 Delete the stack::
 
-  $ sceptre delete-stack dev vpc
+  $ sceptre delete-stack dev vpc [--yes]
 
 
 Next Steps
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -128,7 +128,7 @@ class TestCli(object):
     @patch("sceptre.cli.get_env")
     def test_delete_stack(self, mock_get_env, mock_getcwd):
         mock_getcwd.return_value = sentinel.cwd
-        self.runner.invoke(cli, ["delete-stack", "dev", "vpc"])
+        self.runner.invoke(cli, ["delete-stack", "dev", "vpc", "--yes"])
         mock_get_env.assert_called_with(sentinel.cwd, "dev", {})
         mock_get_env.return_value.stacks["vpc"].delete\
             .assert_called_with()
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -135,6 +135,13 @@ class TestCli(object):
 
     @patch("sceptre.cli.os.getcwd")
     @patch("sceptre.cli.get_env")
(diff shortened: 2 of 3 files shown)


jamesroutley:
Thanks - this is a breaking change, so we can only merge it when we 2.0, so we'll hold off until then.

tests/test_cli.py
```diff
@@ -128,7 +128,7 @@ def test_create_stack(self, mock_get_env, mock_getcwd):
     @patch("sceptre.cli.get_env")
     def test_delete_stack(self, mock_get_env, mock_getcwd):
         mock_getcwd.return_value = sentinel.cwd
-        self.runner.invoke(cli, ["delete-stack", "dev", "vpc"])
+        self.runner.invoke(cli, ["delete-stack", "dev", "vpc", "--yes"])
         mock_get_env.assert_called_with(sentinel.cwd, "dev", {})
         mock_get_env.return_value.stacks["vpc"].delete\
             .assert_called_with()
```
theseanything: Can we have a test to make sure the "--yes" flag works? as in delete_stack without it won't delete

rmbennett: Sure, I'll add this.

rmbennett: So I've added this, however, I'm not 100% sure this is the best way to do it. Any suggestions?

docs/get_started.rst
```diff
@@ -130,7 +130,7 @@ Delete Stack
 
 Delete the stack::
 
-  $ sceptre delete-stack dev vpc
+  $ sceptre delete-stack dev vpc [--yes]
```
rmbennett: @theseanything  the --help option for delete stack now shows that you should pass --yes for delete-stack. 

However, I'm not sure where in the documentation this should be placed - I don't see any clear location for it. (I.e. there aren't any sceptre-command  specific pages for this anywhere at the moment in the docs.)

PR #127, opened 2017-05-24
125 cmd hook

jamesroutley:


docs/stack_config.rst
```diff
@@ -348,6 +348,26 @@ Example:
...
+********
+
+Executes string as a shell or PowerShell command.
+
+Syntax:
+
+.. code-block:: yaml
+
+    <hook_point>:
+      - !cmd <bash_command>
```
theseanything: I assume was renamed to show that it support Windows commands....

in which case we should remove `bash` from `<bash_command>`

docs/stack_config.rst
```diff
@@ -348,6 +348,26 @@ Example:
...
+
+Example:
```
theseanything: Maybe add windows example?

sceptre/hooks/cmd.py
```diff
@@ -0,0 +1,26 @@
+import subprocess
+from sceptre.hooks import Hook
+from sceptre.exceptions import InvalidHookArgumentTypeError
+
+
+class Cmd(Hook):
+    """
+    Cmd implements a Sceptre hook which can run arbitrary shell commands.
```
theseanything: again is shell a linux only term?

sceptre/hooks/cmd.py
```diff
@@ -0,0 +1,26 @@
...
+    """
+
+    def __init__(self, *args, **kwargs):
+        super(Cmd, self).__init__(*args, **kwargs)
+
+    def run(self):
+        """
+        Runs the argument string in child process.
```
theseanything: Runs the argument string in *sub* process.

tests/test_hooks/test_cmd.py
```diff
@@ -0,0 +1,23 @@
...
+    def test_run_with_non_str_argument(self):
+        self.cmd.argument = None
+        with pytest.raises(InvalidHookArgumentTypeError):
+            self.cmd.run()
+
+    @patch('sceptre.hooks.bash.subprocess.check_call')
+    def test_run_with_str_argument(self, mock_call):
+        self.cmd.argument = u"echo hello"
+        self.cmd.run()
+        mock_call.assert_called_once_with(['/bin/bash', '-c', "echo hello"])
```
theseanything: Maybe add another test with a command give a non-zero exit?

jamesroutley:
Agree with changes, I'll add them tomorrow.

tests/test_hooks/test_cmd.py
jamesroutley: added below

jamesroutley:
@seanrankine added those changes

sceptre/hooks/cmd.py
```diff
@@ -0,0 +1,26 @@
...
+
+        :raises: sceptre.exceptions.InvalidTaskArgumentTypeException
+        :raises: subprocess.CalledProcessError
+        """
+        if not isinstance(self.argument, basestring):
+            raise InvalidHookArgumentTypeError(
+                'The argument "{0}" is the wrong type - cmd hooks require '
+                'arguments of type string.'.format(self.argument)
+            )
+        subprocess.check_call(["/bin/bash", "-c", self.argument])
```
theseanything: Remove the /bin/bash -c

theseanything: Remove the /bin/bash -c

sceptre/hooks/cmd.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def run(self):
+        """
+        Runs the argument string in a subprocess.
+
+        :raises: sceptre.exceptions.InvalidTaskArgumentTypeException
+        :raises: subprocess.CalledProcessError
+        """
+        try:
+            subprocess.check_call(self.argument, shell=True)
```
theseanything: Sorry, again we did we opt for `shell=True`?

jamesroutley: Is that 'did we' or 'why did we'?

- did we: yes
- why did we: it's gives us access to constructs defined at the shell level, such as piping. It's considered a security risk if you pass unfiltered external input to `subprocess.check_call` but we decided it was alright here because it'd only run in an environment where the user already had access.

theseanything: Ahah sorry - meant "why did we". So what about build servers? Someone could push and execute potentially malicious code. If it were `shell=False`, do we still have the ability to invoke a shell via `/bin/bash -c`?

theseanything: Decided parsing hook arguments would be difficult and possible security risks with sceptre aren't solved by shell=False.

PR #128, opened 2017-05-24
Sptr 6 - Describe-stack events

rmbennett:
Adds the functionality to describe stack events from the sceptre CLI. 

Potential issue - outputs are not coloured in the way create-stack or delete-stack outputs are.


jamesroutley:
Nice! 

So the code that we use to format events is in the function `_log_new_events()`. How about pulling the formatting code into a new function `_format_event()`, which formats a single item in the list of events returned. 

You can then create a new function `describe_all_events()` which calls `describe_events()`, formats and returns the output. I wouldn't worry about colouring the outputs for now, I've had a look at it and it's a large piece of work - we can implement it later. 

Finally, I'd look at the the "NextToken" property of the returned events - it's possible that a single call won't return all events.

rmbennett:
Hi @jamesroutley finally had some time to work on your suggestions.
I've just pushed an updated _log_new_events and created a _format_event in the format you requested. 

I then created a _log_all_events rather than a describe_all_events as it seemed to fit in better with naming convention to me. (Let me know what you think). 

I still need to look at the NextToken Property

sceptre/stack.py
```diff
@@ -766,6 +767,20 @@ def _get_simplified_status(status):
                 "{0} is unknown".format(status)
             )
 
+    def _format_event(self, event):
+        """
+        Formats a stack event
+        """
```
theseanything: doc string missing parameter info

sceptre/stack.py
```diff
@@ -777,16 +792,19 @@ def _log_new_events(self):
...
-                event["ResourceType"],
-                event["ResourceStatus"],
-                event.get("ResourceStatusReason", "")
-            ]))
+            self._format_event(event)
             self.most_recent_event_datetime = event["Timestamp"]
 
+    def _log_all_events(self):
+        """
+        Log the all stack events while the stack is being built.
```
theseanything: doc string missing parameter info

theseanything:
@rmbennett are you able to update this PR? If not, will close this in 5 days.

rmbennett:
@theseanything will aim to take a look this week.

theseanything:
Closing stale PR as because it has not had recent activity. Thanks for your contributions.

PR #131, opened 2017-05-27
Move warnings init to cli

jamesroutley:
- Deprecation warnings are [ignored](https://docs.python.org/2/library/warnings.html#default-warning-filters) by Python by default. To make sure they're printed, we add `warnings.simplefilter("always", DeprecationWarning)`. This was previously invoked in `ProjectVariables.resolve`. As we may deprecate other things, I've moved the `simplefilter` line to the `cli()` func, where it only needs to be run once.
- Fix grammar in warning message
- Remove `stacklevel` argument. `stacklevel` changes the stack trace printed with the warning. It is used in this example [here](https://docs.python.org/2/library/warnings.html#available-functions). However, it's used in the example because the `warn` is done is a separate function. In that example, we want to print the stack location of where the deprecation function is called, not of where the `warn` func is invoked. Because we don't wrap the warn function, we don't want to change the stack trace; the default behaviour of printing the stack trace of where `warn` was invoked is correct.

Stack trace with `stacklevel`:
```
/Users/helsinki/code/sceptre/sceptre/resolvers/__init__.py:91: DeprecationWarning:
```

Stack trace without `stacklevel`:
```
/Users/helsinki/code/sceptre/sceptre/resolvers/project_variables.py:36: DeprecationWarning:
```


theseanything:
Is there way to test that the CLI will print out warnings?

jamesroutley:
@seanrankine is this good to merge?

PR #134, opened 2017-06-04
103 jinja stack templates

jamesroutley:
Add jinja2 templating support to YAML and JSON templates


sceptre/template.py
```diff
@@ -261,3 +263,30 @@ def _create_bucket(self, region, bucket_name, connection_manager):
                     }
                 }
             )
+
+
+def _render_jinja_template(template_dir, filename, jinja_vars):
```
theseanything: Compromise with static method - another PR for module functions and global loggers.

sceptre/template.py
```diff
@@ -62,17 +62,19 @@ def body(self):
...
 
-            if file_extension in {".json", ".yaml"}:
-                self.logger.debug("%s - Opening file %s", self.name, self.path)
-                with open(self.path) as template_file:
-                    self._body = template_file.read()
+            if file_extension in {".json", ".yaml", "j2"}:
+                self._body = self._render_jinja_template(
+                    os.path.dirname(self.path),
+                    os.path.basename(self.path),
+                    {"sceptre_user_data": self.sceptre_user_data}
```
jamesroutley: yep

sceptre/template.py
```diff
@@ -261,3 +263,30 @@ def _create_bucket(self, region, bucket_name, connection_manager):
                     }
                 }
             )
+
+    @staticmethod
+    def _render_jinja_template(template_dir, filename, jinja_vars):
```
jamesroutley: So I considered this, but I think we need the `jinja` in there to correctly namespace it against the `Template` class. `_render_template` sounds like it should render the CFN template, whereas we're actually rendering a Jinja template

PR #136, opened 2017-06-06
Add simplified Config and ConnectionManager inits

jamesroutley:
Config and ConnectionManager both require quite a lot of
context to initialise. This commit adds new functions for
initialising them. The new function require only the sceptre
directory, an environment path or stack name. This should
make it easier to create config and connection managers
without having to instantiate an environment or stack
first. All new functions implement memoization. If a
particular Config or ConnectionManager is requested twice,
a cached version is returned the second time. The functions
aren't being used at the moment, but we should note that they
aren't necessarily thread safe.

The code here hasn't been integrated into the rest of the codebase yet. I'm opening this PR to show my initial sketch of how I think it should work.

### DIFF
--- a/sceptre/config.py
+++ b/sceptre/config.py
@@ -38,7 +38,7 @@ def environment(sceptre_dir, environment_path, user_variables=None):
     :returns: The Config for the environment
     :rtype: sceptre.config.Config
     """
-    user_variables = user_variables if user_variables != None else {}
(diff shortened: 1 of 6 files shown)


jamesroutley:
@seanrankine What were the requested changes for this?

theseanything:
I feel like we need to re-visit how we want to structure things - this may not be needed after the StackSets feature is properly released.

PR #137, opened 2017-06-10
fixed misleading \"The template does not have the required..\" error

rafcio19:
A fix for if we have a class which calls a non-existing function:

    class SomeClass(object):
        def __init__(self, sceptre_user_data):
            self.template = Template()
            self.sceptre_user_data = sceptre_user_data
            self.non_existing_function()

The template generating class outputs a a misleading error:

`The template does not have the required 'sceptre_handler(sceptre_user_data)' function.`

### DIFF
--- a/sceptre/template.py
+++ b/sceptre/template.py
@@ -111,7 +111,13 @@ class Template(object):
         try:
             body = module.sceptre_handler(self.sceptre_user_data)
         except AttributeError as e:
-            raise e
+            if 'sceptre_handler' in e.message:
+                raise TemplateSceptreHandlerError(
+                    "The template does not have the required "
+                    "'sceptre_handler(sceptre_user_data)' function."
+                )
+            else:
+                raise e
         for directory in relpaths_to_add:
             sys.path.remove(os.path.join(os.getcwd(), directory))
         return body
--- a/sceptre/template.py
+++ b/sceptre/template.py
@@ -111,15 +111,7 @@ class Template(object):
         try:
             body = module.sceptre_handler(self.sceptre_user_data)
         except AttributeError as e:
-            if 'sceptre_handler' in e.message:
-                raise TemplateSceptreHandlerError(
-                    "The template does not have the required "
-                    "'sceptre_handler(sceptre_user_data)' function."
-                )
-            else:
-                raise TemplateSceptreHandlerError(
-                    "Error generating the template: {}".format(e.message)
-                )
+            raise e
         for directory in relpaths_to_add:
             sys.path.remove(os.path.join(os.getcwd(), directory))
         return body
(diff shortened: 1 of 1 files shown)


theseanything:
@rafcio19 I haven't been able to recreate this issue. Could you link me to a full gist of your template file? Possibly could be a whitespace issue.

jamesroutley:
@seanrankine This makes sense to me say the template is:

```
class VPC(object):
    def __init__(self, sceptre_user_data):
        self.template = Template()
        self.sceptre_user_data = sceptre_user_data
        self.non_existing_function()  # This throws an AttributeError

sceptre_handler(sceptre_user_data):
    vpc = VPC()
```

Then line [112](https://github.com/cloudreach/sceptre/blob/master/sceptre/template.py#L112) will raise an `AttributeError`. We catch this error and assume it's because `sceptre_handler` doesn't exist. In fact, `self.non_existing_function()` doesn't exist.

sceptre/template.py
```diff
@@ -110,11 +110,16 @@ def _call_sceptre_handler(self):
...
-                "'sceptre_handler(sceptre_user_data)' function."
-            )
+        except AttributeError as e:
+            if 'sceptre_handler' in e.message:
+                raise TemplateSceptreHandlerError(
+                    "The template does not have the required "
+                    "'sceptre_handler(sceptre_user_data)' function."
+                )
+            else:
+                raise TemplateSceptreHandlerError(
```
jamesroutley: Could we switch this line to just `raise`? Python will raise the original exception, which I think is the preferred behaviour here. The template is written by the user so I think it makes sense to bubble up the same error that they would see if they compiled the template without Sceptre.

There's an example of an re-raising an exception [here](https://docs.python.org/3/tutorial/errors.html#raising-exceptions).

rafcio19: `raise` is definitely a better than the original option

rafcio19:
@theseanything my bad with poor error description, @jamesroutley got there :) PR updated as per @jamesroutley's comments

sceptre/template.py
```diff
@@ -110,11 +110,8 @@ def _call_sceptre_handler(self):
 
         try:
             body = module.sceptre_handler(self.sceptre_user_data)
-        except AttributeError:
-            raise TemplateSceptreHandlerError(
-                "The template does not have the required "
-                "'sceptre_handler(sceptre_user_data)' function."
-            )
+        except AttributeError as e:
+            raise e
```
theseanything: I think @jamesroutley meant to keep the if/else statement but instead of raising:
```
raise TemplateSceptreHandlerError("Error generating the template: {}".format(e.message))
````
just raise the existing Exception 
```
raise e
```

rafcio19: done

mikkoc:
Thanks all, this was a pain to debug before :)

PR #141, opened 2017-06-10
140 Nonscalar resolvers and hooks

### LINKED ISSUE #140: Nonscalar resolver and hook arguments
YAML's tag syntax supports nonscalar tag arguments, such as lists or mappings (or even other tags). For example:

    !sometag
      foo: bar
      baz: 123

However, Sceptre doesn't allow this for the arguments of resolvers and hooks. The resolver and hook constructors added to the YAML loader assume a scalar argument when calling `loader.construct_scalar(node)` (for both [resolvers](https://github.com/cloudreach/sceptre/blob/7ff56cd58f6885847d8577dade2524663958fb72/sceptre/config.py#L224) and [hooks](https://github.com/cloudreach/sceptre/blob/7ff56cd58f6885847d8577dade2524663958fb72/sceptre/config.py#L262)).

It would be useful to support nonscalar arguments for both resolvers and hooks for more configurable cases such as something like this:

    !s3_upload
      path: some/file/path
      s3_bucket: !stack_output another_stack::BucketName
      s3_object_key: some_key

(Note the nested resolver! That's a bonus!)

There's a few ways to work around this at the moment, such as by passing things through `sceptre_user_data` or by passing strings that you parse into multiple arguments, but they typically result in some shenanigans involving using sceptre internals. Nonscalar arguments would be very neat and tidy!

PR incoming...

sftrabbit:
Resolves #140 

This adds support for nonscalar arguments to resolvers and hooks, including arguments that themselves contain other resolvers. As a hypothetical example:

```
!s3_upload
  path: some/file/path
  s3_bucket: !stack_output another_stack::BucketName
  s3_object_key: some_key
```

Took me a while to find a solution that felt right, but I finally came to simply changing both `Resolver` and `Hook`'s `argument` properties to be `ResolvableProperty`s.

When creating the YAML tag constructors, we replace the call to `loader.construct_scalar` with `loader.construct_object` (which can handle any type of value). However, just doing this meant that `loader.construct_object` would complain that is seeing the resolver/hook node twice, because when constructing those resolvers/hooks we want to reevaluate the node to pass as the resolver/hook's argument. To fix this, I've added a function `resolve_node_tag` which copies the node and re-resolves its tag to match whatever value it has, removing the explicit resolver/hook tag.

A few thoughts/concerns:

- `ResolvableProperty` is a little bit scary - I've never seen this descriptor protocol in Python before, but it's a little odd that it's basically affecting the object that its a property of. Not only that, but the specific implementation of `ResolvableProperty` relies on the *containing* object to have a `config` property. I'm not 100% sure what that's for (it seems like it basically would provide a default value for the resolvable property?), but it feels a bit awkward.
- There's [a couple of lines](https://github.com/cloudreach/sceptre/blob/7ff56cd58f6885847d8577dade2524663958fb72/sceptre/resolvers/__init__.py#L69-L70) in `ResolvableProperty` that effectively coerce a `None` value into an empty dict. This becomes a problem for cases where the argument of a resolver/hook is actually meant to be `None`. However, there's a test for it so I assume if I changed it it would be considered a breaking change (although, I don't know what depends on it since removing it breaks only that one test). So instead, I've updated the `EnvironmentVariable` resolver (which has a test for when its argument is `None`) to turn the argument back into `None` before using it. Not super nice, but it avoids any breaking changes.
- There are not a lot of integration tests for config files with resolvers and hooks, so I would suggest testing this change on a few existing and complex sceptre stacks before merging! There's also some of the deeper parts of `Config` that aren't tested much (such as where the call to `loader.construct_object` now is).
- There's nothing in the documentation (as far as I can tell) that suggests that nonscalar arguments shouldn't be possible, so I saw this more as removing an unexpected restriction rather than adding a new feature, so I'm not sure whether documentation specifically for this is necessary.

Commit b1a4be0 is just to make the give the second commit a nicer diff - `Resolver` now has a `ResolvableProperty`, so needed to reorder the classes.

### DIFF
--- a/sceptre/config.py
+++ b/sceptre/config.py
@@ -7,6 +7,7 @@ This module implements a Config class, which stores a stack or environment's
 configuration.
 """
 
+import copy
 import logging
 import os
 import yaml
@@ -221,7 +222,7 @@ class Config(dict):
             :rtype: func
             """
             return lambda loader, node: node_class(
-                loader.construct_scalar(node),
+                loader.construct_object(self.resolve_node_tag(loader, node)),
                 connection_manager,
                 environment_config,
                 self
@@ -258,8 +259,9 @@ class Config(dict):
             :returns: A lambda that constructs hook objects.
             :rtype: func
             """
+
             return lambda loader, node: node_class(
-                loader.construct_scalar(node),
+                loader.construct_object(self.resolve_node_tag(loader, node)),
                 connection_manager,
                 environment_config,
                 self
@@ -300,3 +302,9 @@ class Config(dict):
                 "Added constructor for %s with node tag %s",
                 str(node_class), node_tag
             )
+
+
+    def resolve_node_tag(self, loader, node):
+        node = copy.copy(node)
+        node.tag = loader.resolve(type(node), node.value, (True, False))
+        return node
--- a/sceptre/hooks/__init__.py
+++ b/sceptre/hooks/__init__.py
@@ -2,6 +2,8 @@ import abc
 import logging
 from functools import wraps
 
+from ..resolvers import ResolvableProperty
+
 
 class Hook(object):
     """
@@ -21,6 +23,8 @@ class Hook(object):
     :type connection_manager: sceptre.connection_manager.ConnectionManager
     """
     __metaclass__ = abc.ABCMeta
+    argument = ResolvableProperty("argument")
+    config = {}
 
     def __init__(
             self, argument=None, connection_manager=None,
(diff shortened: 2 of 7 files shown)


theseanything:
Huge thanks for this @sftrabbit... This is something I've definitely been wanting to do. I'll give a more indepth reply/review later this week. At the moment I can say that much better and comprehensive integration tests are on the way. (So will probably hold out until thats in)

sftrabbit:
Just a thought - not sure how well this plays with automatically working out which stacks depend on each other when using `stack_output` as a nested resolver. Would need testing.

ngfgrant:
Sorry @sftrabbit this will need to be reimplemented 😢

PR #146, opened 2017-06-16
111 md docs

jamesroutley:
A big old pull request. A lot of the files are config stuff that's autogenerated by `Gem`, `Sphinx`, `Jekyll`. The new website can be run locally with `make website-latest` and navigating to `localhost:4000/latest/`


theseanything:
There is a lot of stuff going in here and concerned is that much of it isn't for Sceptre itself. Would it be possible to keep the website templates/config/jekyll project in a separate repo and then only keep the website content within this repo?

jamesroutley:
I think the majority of the suff (everything under `website/_api`, which is most of the non-sceptre-specific things) is required by Sphinx, and is the same as what we already have in `docs/`. I think it's neater to leave it here rather than pulling it out to its own repo. Sphinx needs access to the source code to autogenerate the API docs, and the docs should be re-launched when Sceptre is updated. 

I feel that it's cleaner to have a couple of conf files included here (which we have previously included), to avoid messing around with having multiple repos (if we update docs config we'd need to upload the updated docs to S3. To do that we'd need to set up a separate Circle job for the other repo, which git cloned sceptre itself to get the source code for sphinx).

Makefile
```diff
@@ -82,6 +82,43 @@ docs:
...
+	rm -f website/_api/modules.rst
+	sphinx-apidoc -o website/_api sceptre
+	$(MAKE) -C website/_api clean
+	$(MAKE) -C website/_api html
+	mkdir -p website/docs/api
+	rm -rf website/docs/api/_static
+	mkdir -p website/docs/api/_static/
+	cp -r website/_api/_build/html/_static website/docs/api/
+	rm -f website/docs/api/sceptre.html
+	cp website/_api/_build/html/sceptre.html website/docs/api/sceptre.html
```
theseanything: Why only this file? Doesn't the sphinx api pages need sceptre.resolvers.html, sceptre.hooks.html etc..?

theseanything:
I feel like the getting started guide should be apart of the main documentation section. Also could the intro just go on the main page?

theseanything:
The old RST documentation still exists - not sure if this was intentional. We should just keep the md version of the documentation.

theseanything:
There also seems to be syntax highlight issue here:
image

Makefile
jamesroutley: Added

theseanything:
CI failed due to parallelisation of the integration tests.

PR #153, opened 2017-06-24
Increase max_retries

jamesroutley:
This pull request:

- Increases the maximum number of Boto call retries from 5 to 30
- Switches from an exponential backoff to a linear backoff
- Renames exponential_backoff to _retry (as it's no longer exponential)
- Moves function definition from helpers.py to connection_manager.py
because it's only used by the connection manager


sceptre/connection_manager.py
```diff
@@ -7,14 +7,55 @@
...
 
 import boto3
+from botocore.exceptions import ClientError
 
 from .helpers import mask_key
-from .helpers import exponential_backoff
+from .exceptions import RetryLimitExceededError
+
+
+def _retry(func):
```
theseanything: Why do we have decorator when we only apply it only function?

jamesroutley: For separation of concerns - the `call` code makes boto calls, while the retry retries stuff. It also makes it easier to test - we can test the retry functionality without instantiating a boto session.

It's implemented as a decorator because we need to run code before and after the decorated function is run.

PR #155, opened 2017-07-03
Adding the ability to generate CFN parameters file from sceptre config files

adamtemple:
Sceptre generates the CFN parameters file at dynamically at stack creation time - it would be useful for some use cases to be able to generate the raw CFN template and parameters file from the sceptre configuration, such as when providing the templates and inputs to users that need to make use of the templates outwith the scope of Sceptre.

It slightly goes against the practices of having Sceptre control the full stack deployment but this content is generated by the tool internally and it would be useful to have access to this content directly.

Took a stab at adding this based on the existing code but very likely missed something, so any input you might have would be most appreciated!

### DIFF
--- a/integration-tests/features/test-generate-parameters-file.feature
+++ b/integration-tests/features/test-generate-parameters-file.feature
@@ -1,0 +1,6 @@
+Feature: test sceptre's generate-parameter-file
+
+  Scenario: check sceptre can create a json parameters file
+    Given the generate-parameter-file command is run
+    Then the parameters file json syntax is correct
+    And the parameters file contains a valid cidr range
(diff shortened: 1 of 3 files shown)


jamesroutley:
@seanrankine does this handle resolver resolution?

theseanything:
Yup it should work with resolution. However can we hold off until new integration tests are in?

theseanything:
Hi @adamtemple, thanks for the PR. I'd like to merge #164 before this. Could you update the integrations tests accordingly?

integration-tests/features/test-generate-parameters-file.feature
```diff
@@ -0,0 +1,6 @@
+Feature: test sceptre's generate-parameter-file
```
theseanything: Needs updating with new integration tests.

theseanything:
@adamtemple are you able to update this PR? If not, will close this in 5 days.

adamtemple:
Sure - forgot about this one. Let me take a look today and over the weekend.

Cheers,

theseanything:
Will close this PR as stale because it has not had recent activity. Thanks for contributions.

PR #164, opened 2017-07-10
Improved integration tests

theseanything:
This PR is for improved integration test. Feature files have been re-written to be better formatted and more descriptive of what is happening in the test. Test are independent - as they can be run in any order without dependencies on other tests. Now all stacks spun up in testing only use CloudFormation Wait Handles, which allows us to use a more restrictive IAM policy - potentially allow PRs from forks to run integration tests.

All tests previously covered are still included with the exception of continue-update-rollback. This because it was difficult to induce a UPDATE_ROLLBACK_FAILED, it required the use of AWS Resources other than WaitHandles and it significantly increased the length of testing. 

Unfortunately, I have had to serialise the integration test - due to many API calls cause AWS to drop the connections. (Even at 2x parallelisation.) However, we can look at remedying this in a future PR.

Apologies for the size - all of it are file deletions and CloudFormation templates/Sceptre project files.


theseanything:
Currently failing due to bug introduced in [commit](https://github.com/cloudreach/sceptre/commit/970c118f2a18105e13e5bb370930f6e2de2ed674), once #166 is merged these tests should pass.

integration-tests/features/dependency-resolution.feature
```diff
@@ -0,0 +1,39 @@
...
+  Scenario: launch a environment with dependencies that is partially complete
+    Given stack "3/A" exists in "CREATE_COMPLETE" state
+    and stack "3/B" exists in "CREATE_COMPLETE" state
+    and stack "3/C" does not exist
+    When the user launches environment "3"
+    Then all the stacks in environment "3" are in "CREATE_COMPLETE"
+    and that stack "3/A" was created before "3/B"
+    and that stack "3/B" was created before "3/C"
+
+  # Need to fix bug to pass the scenario
```
jamesroutley: Does this need to be done?

integration-tests/steps/change_sets.py
```diff
@@ -0,0 +1,224 @@
...
+@then('change set "{change_set_name}" for stack "{stack_name}" is described')
+def step_impl(context, change_set_name, stack_name):
+    full_name = get_cloudformation_stack_name(context, stack_name)
+    response = retry_boto_call(
+        context.client.describe_change_set,
+        StackName=full_name,
+        ChangeSetName=change_set_name
+    )
+    response.pop("ResponseMetadata")
+    context.output.pop("ResponseMetadata")
```
jamesroutley: Why do we `pop` these when we just key into them elsewhere (e.g. line 173 of this file)

integration-tests/steps/change_sets.py
```diff
@@ -0,0 +1,224 @@
...
+
+@then('the change sets for stack "{stack_name}" are listed')
+def step_impl(context, stack_name):
+    full_name = get_cloudformation_stack_name(context, stack_name)
+    response = retry_boto_call(
+        context.client.list_change_sets,
+        StackName=full_name
+    )
+
+    assert response["Summaries"] == context.output["Summaries"]
```
jamesroutley: Where is `context.output["Summaries"]` set?

integration-tests/steps/change_sets.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+@then('no change sets for stack "{stack_name}" are listed')
+def step_impl(context, stack_name):
```
jamesroutley: what's happening here? Doesn't seem to be listing change sets?

circle.yml
```diff
@@ -10,13 +10,8 @@ dependencies:
     - make install
 test:
   override:
-    - make lint
-    - make test-all
-    - behave --junit --junit-directory $CIRCLE_TEST_REPORTS/behave/junit.xml:
+    - case $CIRCLE_NODE_INDEX in 0) make lint; make test-all; make docs ;; 1) behave integration-tests/ --junit --junit-directory $CIRCLE_TEST_REPORTS/behave/junit.xml ;; esac:
```
jamesroutley: what was the issue with parallelising the tests?

theseanything: Essentially running parallel just hammers the API and causes lots of dropped connections - so lots of tests fail. I tried putting retry function around the boto calls, which works, but I can't wrap the boto calls Sceptre makes. So when Sceptre experiences a dropped connection - it fails. I'm sure there is a way, but might save it for another PR.

The line above essential runs the lint, unit-tests, docs in a separate container to integration tests - which save a little bit of time.

integration-tests/features/dependency-resolution.feature
theseanything: Its a bug we've had for a while, so I'd rather just go ahead and get these new tests in. I'll make a issue with the bug and remove these comments from the code.

integration-tests/steps/change_sets.py
theseanything: in function `@when('the user lists change sets for stack "{stack_name}"')`

integration-tests/steps/change_sets.py
theseanything: yeah - need to remove line 178 - should only assert that `Summaries` is empty

integration-tests/steps/change_sets.py
theseanything: We pop the `ResponseMetadata` because it will change on every request - as it contains attribute such as `RequestId` which is unique and will never be the same. However the rest of the response should be the same as the response provided by Sceptre. I will change above to be consistent.

circle.yml
jamesroutley: Will increasing max retries (which I think we have an open PR for) help? (happy for it to go though in a later PR)

theseanything: Yeah should do! Its the same way I'm retrying the other boto calls.

PR #165, opened 2017-07-10
Bugfix for environment paths with backslashes

theseanything:
This PR make sure Environment objects being given a path with forward slashes. Fixes a bug whereby launching nested environments in a Windows environment would fail due to environment paths containing backslashes.

### DIFF
--- a/sceptre/environment.py
+++ b/sceptre/environment.py
@@ -369,19 +369,18 @@ class Environment(object):
         config_files = glob(os.path.join(
             self.sceptre_dir, "config", self.path, "*.yaml"
         ))
-        base_config_file_names = [
-            os.path.basename(config_file).split(".")[0]
+        stack_basenames = [
+            os.path.splitext(os.path.basename(config_file))[0]
             for config_file in config_files
         ]
-        stack_basenames = [
-            basename for basename in base_config_file_names
-            if basename != "config"
+
+        if "config" in stack_basenames:
+            stack_basenames.remove("config")
+
+        stack_names = [
+            "/".join([self.path, basename]) for basename in stack_basenames
         ]
-        available_stacks = [
-            "/".join([self.path, basename])
-            for basename in stack_basenames
-        ]
-        return available_stacks
+        return stack_names
 
     def _load_stacks(self):
         """
--- a/sceptre/environment.py
+++ b/sceptre/environment.py
@@ -53,10 +53,8 @@ class Environment(object):
         self.logger = logging.getLogger(__name__)
 
         self.sceptre_dir = sceptre_dir
-        self.path = os.path.normpath(environment_path).replace('\\', '/')
+        self.path = self._validate_path(environment_path)
         self._options = {} if options is None else options
-
-        self._check_env_path_valid(self.path)
 
         self._is_leaf = None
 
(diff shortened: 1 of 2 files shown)


theseanything:
Closed - as introduced a delete-env bug

sceptre/environment.py
```diff
@@ -369,19 +372,18 @@ def _get_available_stacks(self):
...
-            basename for basename in base_config_file_names
-            if basename != "config"
+            os.path.splitext(os.path.basename(config_file))[0]
+            for config_file in config_files
         ]
-        available_stacks = [
-            "/".join([self.path, basename])
-            for basename in stack_basenames
+
+        if "config" in stack_basenames:
```
jamesroutley: `stack_basenames` doesn't word as a variable name if it includes a `config` item. Can we move this if statement to a filter in the list comp on 375? Or change the name

sceptre/environment.py
```diff
@@ -74,20 +72,25 @@ def __repr__(self):
...
         Raises an InvalidEnvironmentPathError if the path has a leading or
         trailing slash.
 
         :param path: A directory path
         :type path: str
         :raises: sceptre.exceptions.InvalidEnvironmentPathError
+        :returns: A normalised path with forward slashes.
+        :returns: string
         """
+        path = path.replace('\\', '/')
```
jamesroutley: Could we add a comment as to why we're doing this - I don't think the code is self-explanatory

sceptre/environment.py
```diff
@@ -74,20 +72,25 @@ def __repr__(self):
...
         """
+        Converts backslashes to forward slashes.
         Raises an InvalidEnvironmentPathError if the path has a leading or
         trailing slash.
 
         :param path: A directory path
         :type path: str
         :raises: sceptre.exceptions.InvalidEnvironmentPathError
+        :returns: A normalised path with forward slashes.
+        :returns: string
```
jamesroutley: I think this should be `rtype`

tests/test_environment.py
```diff
@@ -74,21 +75,42 @@ def test_repr(self):
...
-            self.environment._check_env_path_valid(
+            self.environment._validate_path(
                 "/this/environment/path/is/invalid"
             )
 
-    def test_check_env_path_valid_with_invalid_path_trailing_slash(self):
+    def test_validate_path_with_leading_backslash(self):
+        with pytest.raises(InvalidEnvironmentPathError):
+            self.environment._validate_path(
+                '\\this\environment\path\is\invalid'
```
jamesroutley: Can we double quote strings

tests/test_environment.py
```diff
@@ -74,21 +75,42 @@ def test_repr(self):
...
         with pytest.raises(InvalidEnvironmentPathError):
-            self.environment._check_env_path_valid(
+            self.environment._validate_path(
                 "this/environment/path/is/invalid/"
             )
 
+    def test_validate_path_with_trailing_backslash(self):
+        with pytest.raises(InvalidEnvironmentPathError):
+            self.environment._validate_path(
+                'this\environment\path\is\invalid\\'
```
jamesroutley: Why do we need to escape the trailing slash (if I understand what's happening here), but not the others?

jamesroutley: Oh it's so as to not escape the closing quote - I get it.

sceptre/environment.py
jamesroutley: Do we need to worry about environment paths that contain single letter directories which happen to form special chars?

e.g.
```
In [12]: "\a\n\c".replace("\\", "/")
Out[12]: '\x07\n/c'
```

theseanything: Added comment in the doctring.

theseanything: So above you are using a string literal and should be aware that you haven't escaped your backslashes. When paths for Windows are read by python - they will have escaped backslashes e.g. "\\a\\n\\c".

theseanything: I mean e.g. "\\\\a\\\\n\\\\c" , github escaped my slashes ahah

PR #171, opened 2017-07-19
CircleCI 2.0 Functionality

theseanything:
Adds support for CircleCI 2.0. Tests now runs in parallel and finish around 6 minutes. Also Dockerfile included for custom Docker image used for testing - based of CircleCI's Python image, but adds Ruby.


.circleci/config.yml
```diff
@@ -0,0 +1,133 @@
...
+          command: |
+            SHORT_SHA1=$(echo $CIRCLE_SHA1 | cut -c -7)
+            GIT_TAG=$(git describe --abbrev=0 --tags | sed 's/\./-/g')
+            . ./venv/bin/activate
+            if [[ "${CIRCLE_TAG}" =~ ^v[0-9]+\.[0-9]+\.?[0-9]*$ ]]; then
+              make docs-latest
+              aws s3 sync ~/project/docs/_site s3://sceptre.cloudreach.com/latest/ --delete
+              make docs-tag
+              aws s3 sync ~/project/docs/_site s3://sceptre.cloudreach.com/$GIT_TAG/ --delete
+            elif [[ "${CIRCLE_BRANCH}" == "circle-ci-2.0" ]]; then
```
jamesroutley: I'm guessing `circle-ci-2.0` should be `master`?

.circleci/config.yml
```diff
@@ -0,0 +1,133 @@
...
+  build:
+    docker:
+      - image: cloudreach/sceptre-circleci:0.1
+    steps:
+      - checkout
+      - run:
+          name: 'Creating Virtualenv'
+          command: virtualenv venv
+      - restore_cache:
+          key: v2-requirements-{{ checksum "requirements.txt" }}
```
jamesroutley: Why **v2**-requirements?

theseanything: It v2 because caches are immutable and I updated the contents of what gets cached from v1 - so we can't use v1

jamesroutley:
Is this ready to go? Looks like one of the tests is failing

PR #172, opened 2017-07-19
Indent directory structure properly

stig:
It is clear from the surrounding text that the example directory structure was not indented correctly.


theseanything:
Thanks @stig!

PR #174, opened 2017-07-21
Jinja templating for only `.j2.` files.

theseanything:
This PR is fix an oversight whereby Jinja templating is enabled by default for file with the extension `.json` and `.yaml`. This is not desirable as there are cases in which Jinja templating syntax maybe used in template, but not for Jinja templating purposes. This PR simply restricts Jinja templating to files with `.j2` extension to provide a quick fix. Also contains documentation updates.


docs/docs/templates.md
```diff
@@ -16,17 +16,22 @@ Sceptre uses CloudFormation or Troposphere templates to launch AWS Stacks. Conve
...
 
-Templates with `.json` or `.yaml` extensions are treated as CloudFormation templates. They are read in and launched without modification.
+Templates with `.json` or `.yaml` extensions are treated as CloudFormation templates. They are read in and used without modification.
 
 
-## Troposphere
+## Jinja
 
-Templates with a `.py` extension are treated as Troposphere templates. They should implement a function named `sceptre_handler(sceptre_user_data)` which returns the CloudFormation template as a `string`. Sceptre User Data is passed to this function as an argument. If Sceptre User Data is not defined in the Stack Config file, Sceptre passes an empty `dict`.
+Templates with `.j2` extensions are treated as Jinja2 templates. Sceptre User Data is accessible within templates as `sceptre_user_data`. For example `{{ sceptre_user_data.some_variable }}`.
```
jamesroutley: Can we mention that `.j2` files are normal `json`/`yaml` templates with added support for SUD variable injection?

Issue #179, opened 2017-08-08
STS Session Timeouts


TWinsnes:
If your deployment takes longer than the STS session timeout (max 1 hour), sceptre fails with `ClientError: An error occurred (ExpiredToken) when calling the DescribeStacks operation: The security token included in the request is expired`

If `connection_manager.boto_session` caches the sts expiration date and recreates the session if it has expired (or is close to expiring) it will add support for deployment that take a long time.

PR #180, opened 2017-08-08
Recreate STS session if expired

TWinsnes:
STS sessions are limited to maximum 1 hour, if a deploy takes
longer than this, sceptre will fail. This commit fixes this by
caching the expiry time and validating it against current time
when a call is made.

Fixes issue in #179 

------------

* [x] [Squashed related commits together][2].
* [x] Added unit tests.

### DIFF
--- a/sceptre/connection_manager.py
+++ b/sceptre/connection_manager.py
@@ -11,6 +11,8 @@ import functools
 import logging
 import threading
 import time
+from datetime import datetime
+from dateutil import tz
 
 import boto3
 from botocore.exceptions import ClientError
@@ -81,6 +83,7 @@ class ConnectionManager(object):
         self.iam_role = iam_role
         self.profile = profile
         self._boto_session = None
+        self._boto_session_expiration = None
 
         self.clients = {}
 
@@ -110,6 +113,8 @@ class ConnectionManager(object):
         with self._session_lock:
             self.logger.debug("Getting Boto3 session")
 
+            self._clear_session_cache_if_expired()
+
             if self._boto_session is None:
                 self.logger.debug("No Boto3 session found, creating one...")
                 if self.iam_role:
@@ -132,6 +137,7 @@ class ConnectionManager(object):
                         aws_session_token=credentials["SessionToken"],
                         region_name=self.region
                     )
+                    self._boto_session_expiration = credentials["Expiration"]
                     self.logger.debug(
                         "Using temporary credential set: %s",
                         {
@@ -180,12 +186,25 @@ class ConnectionManager(object):
         :rtype: boto3.client.Client
         """
         with self._client_lock:
+            self._clear_session_cache_if_expired()
+
             if self.clients.get(service) is None:
                 self.logger.debug(
                     "No %s client found, creating one...", service
                 )
                 self.clients[service] = self.boto_session.client(service)
             return self.clients[service]
+
+    def _clear_session_cache_if_expired(self):
+        """
+        Removes boto_session and all clients if the existing session has
+        expired.
+        """
+        if self.iam_role and self._boto_session_expiration:
+            if self._boto_session_expiration >= datetime.now(tz.tzutc()):
+                self.logger.debug("Boto session has expired")
+                self.clients = {}
+                self._boto_session = None
 
     @_retry_boto_call
     def call(self, service, command, kwargs=None):
--- a/tests/test_connection_manager.py
+++ b/tests/test_connection_manager.py
@@ -8,6 +8,7 @@ from sceptre.exceptions import RetryLimitExceededError
 from boto3.session import Session
 import botocore
 from botocore.exceptions import ClientError
+from datetime import datetime
 
 
 class TestConnectionManager(object):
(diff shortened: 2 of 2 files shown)


TWinsnes:
Hey guys, would it be possible to get a comment on this change. We need the extended timeout for deploying legacy applications, and it would be nice to be able to install sceptre using pip from an official source for deploying into production.

b-t-g:
Thanks for the pull request! I was talking with @theseanything and we're hesitant to accept it with the addition of a new dependency (pytz). If there is a way to rewrite the functionality without adding a new dependency, we'll take another look. Thanks again!

akdor1154:
looks like boto uses dateutil.tz for the same purpose, would that be an acceptable dependency? http://dateutil.readthedocs.io/en/stable/tz.html
https://github.com/boto/botocore/blob/cd8a96b3fe4e8cb5ef5cdc749a006f14a7112c73/botocore/utils.py#L24

theseanything:
Yup perfect as won't technically increase the number of dependencies, could you update the PR to use dateutil instead?

TWinsnes:
Thanks for the feedback guys, will update it once I get out of work tonight 👍

sceptre/connection_manager.py
```diff
@@ -105,6 +110,9 @@ def boto_session(self):
         with self._session_lock:
             self.logger.debug("Getting Boto3 session")
 
+            if self._is_session_expired():
```
theseanything: Consider this logic if `self.iam_role = None`. I think that with no iam_role, this will make `self._boto_session = None` everytime and recreate the session for every call. (i.e. the expiration will stay "in the past" (the initialised value) when logic flow uses the `else` block (lines 146-163)

sceptre/connection_manager.py
```diff
@@ -78,6 +81,8 @@ def __init__(self, region, iam_role=None):
         self.region = region
         self.iam_role = iam_role
         self._boto_session = None
+        self._boto_session_expiration = \
+            datetime.now(tz.tzutc()) - timedelta(days=9)
```
theseanything: Can we make this equal to None? Then add logic when the expiration is None?

sceptre/connection_manager.py
```diff
@@ -170,13 +179,36 @@ def _get_client(self, service):
...
+
+        :returns: session expired boolean
+        :rtype: boolean
+        """
+
+        if (
+            self.iam_role and
+            self._boto_session_expiration >= datetime.now(tz.tzutc())
+           ):
+            self.logger.debug(
```
theseanything: Does this not fit on one line?

sceptre/connection_manager.py
```diff
@@ -170,13 +179,36 @@ def _get_client(self, service):
...
+        Validates the boto session to make sure it is still valid and returns
+        false if it has expired
+
+        :returns: session expired boolean
+        :rtype: boolean
+        """
+
+        if (
+            self.iam_role and
+            self._boto_session_expiration >= datetime.now(tz.tzutc())
```
theseanything: can we set a variable `expired = self._boto_session_expiration >= datetime.now(tz.tzutc())`, then put if statement on one line ` if self.iam_role and expired:`

sceptre/connection_manager.py
TWinsnes: _is_session_expired() should only return true if iam_role is set, I have added a test that explicitly validates this :)

TWinsnes:
Thanks for the feedback, I've updated the pull request to fix up any issues

nebffa:
Hey @theseanything , just wondering if you've had a chance to review @TWinsnes 's latest changes?

TWinsnes:
hey @theseanything any news on this? Would really appreciate it if we could use an official pip release before we start deploying to prod with sceptre

TWinsnes:
Hey guys, any news on this?

sceptre/connection_manager.py
```diff
@@ -170,13 +177,37 @@ def _get_client(self, service):
...
+                self.clients[service] = None
+
             if self.clients.get(service) is None:
                 self.logger.debug(
                     "No %s client found, creating one...", service
                 )
                 self.clients[service] = self.boto_session.client(service)
             return self.clients[service]
 
+    def _is_session_expired(self):
```
theseanything: Could we refactor this function to something like this:
```python
    def _check_session_expiration(self):
        """
        Removes boto_session and all clients if the existing session has expired.

        """
        if self.iam_role and self._boto_session_expiration:
            if self._boto_session_expiration >= datetime.now(tz.tzutc()):
                self.logger.debug("Boto session has expired")
                self.clients = {}
                self._boto_session = None
```

sceptre/connection_manager.py
theseanything: with refactor suggested below - this if statement isn't needed.

sceptre/connection_manager.py
```diff
@@ -170,13 +177,37 @@ def _get_client(self, service):
         :rtype: boto3.client.Client
         """
         with self._client_lock:
+            if self._is_session_expired():
+                self.clients[service] = None
```
theseanything: replace if statement with `self._check_session_expiration(self)` - with refactor suggested below.

theseanything:
@TWinsnes sorry for the delay. I've made a suggestion to neaten up the code. Could I ask you to also fill out the checklist in the PR description once your one refactoring? (You should be able to squash all the commit into a single commit.) Happy to merge as soon as thats done :)

TWinsnes:
@theseanything  thanks for all the feedback. Updated, rebased of master, and squashed commits together

PR #189, opened 2017-08-29
New CLI init command

theseanything:
This PR introduces:
sceptre init project <project-name>
This will create a project folder containing a templates and config folder. It will also provide an interactive way to generate a base config.yaml

sceptre init env <env-name>
This will create folders for the path containing the env name. It will also provide an interactive way to generate a base config.yaml and not include config that is the same as parent config files..


sceptre/cli.py
```diff
@@ -574,6 +576,170 @@ def get_stack_policy(ctx, environment, stack):
...
+@click.argument('project_name')
+@click.pass_context
+def init_project(ctx, project_name):
+    """
+    Initialises a new project.
+
+    Creates PROJECT_NAME project folder.
+    """
+    cwd = os.getcwd()
+    sceptre_folders = {"config", "templates"}
```
b-t-g: Does sceptre_folders need to be explicitly declared as a variable? It looks like it's only used in one place.

sceptre/cli.py
```diff
@@ -574,6 +576,170 @@ def get_stack_policy(ctx, environment, stack):
...
+    project_folder = os.path.join(cwd, project_name)
+    try:
+        os.mkdir(project_folder)
+    except OSError as e:
```
b-t-g: If you're only checking to see whether the directory exists and re-raising the exception otherwise, would it be just as safe and a bit more concise to just check to see whether the directory already exists?

sceptre/cli.py
```diff
@@ -574,6 +576,170 @@ def get_stack_policy(ctx, environment, stack):
...
+    environment_exists = False
+
+    # Make folders for the anvironment
+    try:
+        os.makedirs(folder_path)
+    except OSError as e:
+        # Check if environment folder already exists
+        if e.errno == errno.EEXIST:
+            environment_exists = True
+        else:
```
b-t-g: Same here as above.

sceptre/cli.py
```diff
@@ -574,6 +576,170 @@ def get_stack_policy(ctx, environment, stack):
...
+            raise
+
+    init_config_msg = 'Do you want initialise config.yaml?'
+
+    if environment_exists:
+        init_config_msg = 'Environment path exists. ' + init_config_msg
+
```
b-t-g: Does it make sense to just put this init_config_msg assignment inside the try/if statement instead?

sceptre/cli.py
```diff
@@ -574,6 +576,170 @@ def get_stack_policy(ctx, environment, stack):
...
+    # Ask for new values
+    for key, value in config.items():
+        config[key] = click.prompt(
+            'Please enter a {0}'.format(key), default=value
+        )
+
+    # Remove nested values that are the same
+    config = {k: v for k, v in config.items() if nested_config.get(k) != v}
+
```
b-t-g: A bit nit-picky, but it seems like the expressions containing nested_config could (and should) be grouped together.

sceptre/exceptions.py
```diff
@@ -7,6 +7,12 @@ class SceptreException(Exception):
     """
 
 
+class ProjectExistsError(SceptreException):
+    """
+    Error raised when Sceptre project already exists.
+    """
+
```
b-t-g: It looks like this is the only error which doesn't call pass, is there a reason for this?

sceptre/cli.py
theseanything: Technically not - I'm being overly verbose just to help the readability of the code. Can take it out if you think its too much.

sceptre/cli.py
theseanything: True, I was following the pythonic "its better to ask for forgiveness than permission" mentality.

sceptre/cli.py
```diff
@@ -574,6 +576,170 @@ def get_stack_policy(ctx, environment, stack):
...
+    """
+    cwd = os.getcwd()
+    sceptre_folders = {"config", "templates"}
+    project_folder = os.path.join(cwd, project_name)
+    try:
+        os.mkdir(project_folder)
+    except OSError as e:
+        # Check if environment folder already exists
+        if e.errno == errno.EEXIST:
+            raise ProjectExistsError(
```
theseanything: Think I'll rename that to ProjectAlreadyExistsError

sceptre/cli.py
```diff
@@ -574,6 +576,170 @@ def get_stack_policy(ctx, environment, stack):
...
+    :param config_dir: The directory path to the top-level config folder.
+    :type config_dir: str
+    :param path: The directory path to the environment folder.
+    :type path: str
+    """
+    # Create full path to environment
+    folder_path = os.path.join(config_dir, new_path)
+    environment_exists = False
+
+    # Make folders for the anvironment
```
theseanything: spelling `environment`

sceptre/cli.py
theseanything: Ah right! instead of using the `environment_exists` variable?

sceptre/cli.py
theseanything: Ah not sure what you mean - could you explain further?

sceptre/exceptions.py
theseanything: You're right it should!

sceptre/cli.py
b-t-g: It may be easier to explain in person (so I can point to code and use words like "this", "that", and "there"). It's not a huge deal either way.

sceptre/cli.py
b-t-g: After thinking about it, it makes the possibility of failure more explicit.

sceptre/cli.py
b-t-g: I'm torn on whether it makes it more readable (since what it's referring to is so simple), I'll defer to your judgement.

sceptre/cli.py
b-t-g: Yep!

b-t-g:
Looks good; I'll merge once the tests get sorted out.

Issue #163, opened 2017-07-10
ValueError: max_workers must be greater than 0


mikkoc:
Just ran into this error when deleting a whole environment.
All the stacks were successfully deleted, the error came up after sceptre was finished.
Not sure it's reproducible...

```
# sceptre --var-file=environments/prod.yaml delete-env base
[...]
[2017-07-10 13:26:07] - sceptre.stack - base/network/vpc - delete complete
Traceback (most recent call last):
  File "/Users/mikkocaldara/tfl/git/tfl-eos-infrastructure/venv/bin/sceptre", line 11, in <module>
    load_entry_point('sceptre==1.1.1', 'console_scripts', 'sceptre')()
  File "/Users/mikkocaldara/tfl/git/tfl-eos-infrastructure/venv/lib/python2.7/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/Users/mikkocaldara/tfl/git/tfl-eos-infrastructure/venv/lib/python2.7/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/Users/mikkocaldara/tfl/git/tfl-eos-infrastructure/venv/lib/python2.7/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/mikkocaldara/tfl/git/tfl-eos-infrastructure/venv/lib/python2.7/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/mikkocaldara/tfl/git/tfl-eos-infrastructure/venv/lib/python2.7/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/Users/mikkocaldara/tfl/git/tfl-eos-infrastructure/venv/lib/python2.7/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/Users/mikkocaldara/tfl/git/tfl-eos-infrastructure/venv/lib/python2.7/site-packages/sceptre/cli.py", line 93, in decorated
    return func(*args, **kwargs)
  File "/Users/mikkocaldara/tfl/git/tfl-eos-infrastructure/venv/lib/python2.7/site-packages/sceptre/cli.py", line 321, in delete_env
    response = env.delete()
[...]

PR #190, opened 2017-08-29
Fix misleading empty environment error

cboss24:
Now when an empty environment (folder) is encountered, an exception will be raises when loading stacks. This exception provides feedback to the developer concerning their invalid environment hierarchy. Prior to this change, various misleading exceptions could be thrown depending on the cli command and placement of the empty environment.

This addresses the issue #163

### DIFF
--- a/sceptre/environment.py
+++ b/sceptre/environment.py
@@ -201,16 +201,21 @@ class Environment(object):
         :param command: The stack command to run. Can be (launch | delete).
         :type command: str
         """
-        num_stacks = len(self.stacks)
-        with ThreadPoolExecutor(max_workers=num_stacks) as executor:
-            futures = [
-                executor.submit(
-                    self._manage_stack_build, stack,
-                    command, threading_events, stack_statuses, dependencies
-                )
-                for stack in self.stacks.values()
-            ]
-            wait(futures)
+        if self.stacks:
+            num_stacks = len(self.stacks)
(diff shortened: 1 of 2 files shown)


theseanything:
@cboss24 thanks for the PR - so was thinking about this more. I agree the current exception and error message is misleading. But not sure that throwing an exception is the correct way to handle this case. For example, if you are launch multiple environments containing an empty environment, then the exception would interrupt this process. I feel like logging a message that an environment is empty is and handling the error would be better.

cboss24:
@theseanything thanks for the feedback. I agree with your comments and have taken a first stab at a new solution. Let me know what you think.

sceptre/environment.py
```diff
@@ -201,6 +201,12 @@ def _build(self, command, threading_events, stack_statuses, dependencies):
         :param command: The stack command to run. Can be (launch | delete).
         :type command: str
         """
+        if not self.stacks:
```
theseanything: Feel this could be neater if we reverse the if statement. i.e.
```python
if self.stacks:
    do_stuff
else:
    self.logger.info(.....)

cboss24:
@theseanything done. Just a procedural question: do you prefer commits be squashed into a single commit once the branch is ready to be merged?

theseanything:
Yeah if this could all be a single commit that’d be great.

cboss24:
Squashed!

PR #192, opened 2017-09-05
Update nested dictionnaries in cascaded configs

afraisse:
Instead of overwriting nested configs, check if the config item is a dict and if so, update it instead.

This allows generic definitions of nested dictionaries, such as parameters or scepte_user_data for instance.
See the updated test for example.
Really enjoying sceptre here at ubisoft, thanks for sharing!

### DIFF
--- a/sceptre/config.py
+++ b/sceptre/config.py
@@ -114,7 +114,8 @@ class Config(dict):
 
         Traverses the environment path, from top to bottom, reading in all
         relevant config files. If config items appear in files lower down the
-        environment tree, they overwrite items from further up. Jinja2 is used
+        environment tree, they overwrite items from further up. If a config item holds a nested
+        dictionnary, update it recursively instead of the config item itself. Jinja2 is used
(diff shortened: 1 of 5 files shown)


afraisse:
Since I created this merge request, changes in the tests have made them to fail. I got no feedback on this PR, but saw that you're deprecating cascading stack config. Should this PR be discarded then ?

theseanything:
Apologies @afraisse! Yes we are deprecating cascading config. I agree we need for shared/common values - however we think cascading config doesn't provide a flexible enough solution. For example config can only be cascaded down into a single file within an environment (due to the file name needing to be same). We currently have support for variables file - which is a partial solution this solution. Further discussion on this problem is needed.

afraisse:
@theseanything Thanks for the feedback !

JuanCanham:
Can we re-look at this for v1, variable files still seam ugly and this still seams like a much cleaner fix when cascading configs are supported (v1)

PR #196, opened 2017-09-14
add a `--confirm` option to update-stack-cs

mithun:
update-stack-cs prompts the user to continue. The `--confirm` option
suppresses this prompt and continues to execute the change-set

### DIFF
--- a/sceptre/cli.py
+++ b/sceptre/cli.py
@@ -461,14 +461,17 @@ def list_change_sets(ctx, environment, stack):
 @cli.command(name="update-stack-cs")
 @stack_options
 @click.option("--verbose", is_flag=True)
+@click.option("--confirm", is_flag=True)
 @click.pass_context
 @catch_exceptions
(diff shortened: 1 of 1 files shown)


theseanything:
Hi @mithun, thanks for your PR. Was wondering what the use case is for this? You are creating a change-set and executing regardless, it would seem that you should just use update-stack?

mithun:
@theseanything I'm using a workflow where all updates are done using change-sets. And sceptre's logs are then archived for audit.

This feature supports the use case where "describe-change-set" output is stored in logs. Which then allows us to review previous execution logs to review change-sets (actually change-set descriptions) that was applied to the stack

b-t-g:
I'm unable to test this now, but would "echo y | <command>" work for this purpose (without needing the --confirm flag)?

mithun:
@b-t-g Good point. I didn't think of trying that. Yes, the `echo y | <command>` works for me.

theseanything:
Closing as redundant functionality.

Issue #198, opened 2017-09-25
Enhancement proposal: support stack creation via change sets


aehlke:
Change sets are typically used for reviewing changes to an existing stack before execution. However, they also support creating new stacks via the `--change-set-type=CREATE` argument (which defaults to `--change-set-type=UPDATE`). Sceptre does not support this parameter and relies on the default.

There may be other compelling use cases for this, but the primary one for our organization is to be able to deploy SAM (Serverless Application Model) templates via Sceptre. SAM templates are a great way to configure Lambda+API Gateway services. For some reason, `create-stack` cannot be used with templates that have a `Transform` directive; the workaround is to create a change set and execute it with `--change-set-type=CREATE` to create the stack, which works with SAM-transformed templates.

awscli's `deploy` command works this way - it creates a change set and executes it, whether it's for a new or existing stack. We use the awscli `deploy` command in lieu of Sceptre to launch SAM stacks.

My proposal is to add the `--change-set-type` optional parameter to Sceptre's `create-change-set` command with the same interface and behavior as [awscli's `create-change-set` command](http://docs.aws.amazon.com/cli/latest/reference/cloudformation/create-change-set.html).

I'm going to start work on this shortly if this sounds good.

aehlke:
What can I do to push this forward? It's unclear how I can trigger a test run for my PR. Thanks.

theseanything:
Hi @aehlke, sorry for the delay! So yeah totally agree we support for this. Thanks for your PR! So had a think about this - could this be better handled by getting create-change-set to see if the stack exists and then use the appropriate "change-set-type"? Which would eliminate the need for setting the flag.

aehlke:
@theseanything I'm open to making that change, just not in tune enough with the design goals of Sceptre to know whether we want to be clever like that. Most of the other commands pretty closely mirror the AWS API 1:1 so I'd wonder if this would lead to "uncanny valley" surprises where it looks very much like the AWS API, but behaves slightly differently in some cases. What're your thoughts on that? I can easily make the change either way.

aehlke:
@theseanything I've put up a PR with tests passing here for that automatic inference option: https://github.com/cloudreach/sceptre/pull/240

ryanpagel:
What is the status of this?  We'd like to be able to use changesets when deploying new stacks to production then let another developer review the changeset and Execute it from the AWS console.  Is it possible to do this with sceptre?

PR #199, opened 2017-09-26
Add change-set-type param to create-change-set

aehlke:
This allows creating stacks via change sets, rather than exclusively updating existing stacks.

(My use case for this is to be able to create stacks from templates which use a `Transform` directive, necessary for Serverless Application Model templates - for some reason, `create-stack` fails with these and the workaround is to create the stack view a change set instead.)

Addresses #198


theseanything:
Closing PR for now - waiting for response on issue #198.

PR #205, opened 2017-09-28
Entry point hooks

theseanything:
This PR changes the way in which resolvers and hooks integrate with Sceptre. With this PR Sceptre takes advantages of entry points to enable plugins. This means custom resolvers and hooks can now be pip installed. Main refactor occurs in the Config class whereby the "hooks" and "resolvers" folder within the local project and the Sceptre installation are searched to find subclasses of either Hook or Resolver. Now the Config class will use classes registered at the `sceptre.hooks` and sceptre.resolvers` entry points to create the Yaml constructors. Unit tests and documentation have been updated. No further integration tests added as should be covered by the tests specific for each resolver or hook.

### DIFF
--- a/docs/docs/hooks.md
+++ b/docs/docs/hooks.md
@@ -102,19 +102,25 @@ Users can define their own custom hooks, allowing users to extend hooks and inte
 
 A hook is a Python class which inherits from abstract base class `Hook` found in the `sceptre.hooks module`.
 
-Hooks are require to implement a `run()` function that takes no parameters and to call the base class initializer on initialisation.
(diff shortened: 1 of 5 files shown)


b-t-g:
I'd say squash these into two or three commits: The one with the feature itself implemented and the documentation changes (and the merge commit since I'd imagine it would be a pain to get of it). Does this have a corresponding issue? If not, I'll just basically copy and paste your PR into an issue and ask to update your commit messages (for aforementioned code spelunking purposes).

theseanything:
@b-t-g Could you give this a final check?

PR #207, opened 2017-09-29
Add describe-env-dependencies command

### LINKED ISSUE #201: Add option to output dependency graph of resources for environment/stack
Motivation: Some stacks and environments can take a long time to fully launch (I'm looking at you, anything that has RDS). If a mistake is made, the feedback loop can be longer than it should be (and sometimes mistakes can remain hidden for a while). Having a comparatively shorter operation to get a fairly nice visualization of a stack/environment would help greatly to quickly write a functional Sceptre project. I'm thinking of this as basically the inverse of CloudFormation Designer.

cboss24:
Fix #201 Add describe-env-dependencies to expose the stack dependencies within an environment

This commit also allows users to output their dependencies using the DOT language, which can be used to graph the dependencies using programs like Graphwiz.

Couple questions:
1. Thoughts on the name of the command? Would something simpler like `graph` be more user friendly?
2. Is it possible to include a picture in the docs? Something like this is good motivation to use a command: dependencies

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -36,6 +36,7 @@ $ sceptre delete-stack
 $ sceptre describe-change-set
 $ sceptre describe-env
 $ sceptre describe-env-resources
+$ sceptre describe-env-dependencies
 $ sceptre describe-stack-outputs
 $ sceptre describe-stack-resources
 $ sceptre diff
@@ -77,4 +78,11 @@ Note that Sceptre prepends the string `SCEPTRE_` to the name of the environment 
 ```shell
 $ env | grep SCEPTRE
 SCEPTRE_<output_name>=<output_value>
+```
+
+## Visualizing Environment Dependencies
+
+The `describe-env-dependencies` command can produce output in the DOT language. DOT can easily be converted to an image by making use of the dot command provided by GraphViz:
+```shell
+$ sceptre describe-env-dependencies ENVIRONMENT --dot | dot -Tpng > dependencies.png
 ```
--- a/sceptre/cli.py
+++ b/sceptre/cli.py
@@ -562,6 +562,52 @@ def describe_env(ctx, environment):
     write(responses, ctx.obj["output_format"], ctx.obj["no_colour"])
 
 
+@cli.command(name="describe-env-dependencies")
+@environment_options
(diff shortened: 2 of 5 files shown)


b-t-g:
Wow, not even 12 hours after I posted my issue! I'll want to play around with this on a few existing environments that I have on top of a standard code review to get a better feel for how it works.

Regarding your first bullet point, I think since describe has a relatively consistent meaning in Sceptre (and the aws cli in general), that desecribe-env-dependencies is not the best name. Off the top of my head, something like output-dependency-graph may be better, but I haven't yet convinced myself that it is the best name either.

I'll look into how to get pictures into the docs, it seems doable in principle, but I don't know how to do it right now.

Thanks for the PR!

theseanything:
@cboss24 Thanks for PR. Could you give some specific use cases for this? I'm reluctant to merge as this seems to add a bit of complexity (API changes etc) for low benefit to Sceptre's core functionality.

cboss24:
Two main uses cases I see:

1. `Dev`: You have a stack in your environment that is slow to CRUD (elasticsearch, RDS...). A graph command would help you determine if the stack you are iterating on will cascade changes to that slow stack. A graph command might even help you simplify your dependency graph to minimize or eliminate cascading changes to that slow stack.

2. `Prod`: You are releasing a new change to PROD. You know which stacks will be directly changed from manual inspection of the commit. However, you would have to manually inspect the code base to determine how those stack changes will propagate to the rest of the environment. A graph command would make this information available in a automated / graphical way.

I disagree that this is not core Sceptre functionality. One of Sceptre killer features is the ability to automatically detect dependencies and orchestrate the stack commands around those dependencies. In my opinion this change is simply enhancing this core feature.

Let me know what you think!

cboss24:
@theseanything @b-t-g any thoughts on this PR?

theseanything:
Sorry for the delay @cboss24. I agree it is an enhancement, but not fundamental to Sceptre functionality. For the sake of project maintainability - I feel this is something we are going to have to leave out. 

We are a looking at maybe introducing the ability to extend sceptre by adding custom cli commands/functionality as plugins which can be pip installable - this may be a good candidate for that.

PR #209, opened 2017-10-03
Refactor validate template

theseanything:
This PR addresses inconsistency with OOP principals within the Stack class. The `validate_template` method now resides in the Template class as `validate` - this is because the method is specific to the template. To enable this the `_get_template_details` method in the Stack class was more aptly renamed to `get_boto_call_parameter` and moved inside the Template class. The API to validate a template is now `Stack().template.validate()` rather than `Stack().validate_template()`. No new functionality has been added.

-----------------

Before submitting the PR make sure the following are checked:

* [x] [Squashed related commits together][2].
* [x] Added unit tests.

### DIFF
--- a/integration-tests/sceptre-project/config/config.yaml
+++ b/integration-tests/sceptre-project/config/config.yaml
@@ -1,1 +1,1 @@
-project_code: sceptre-integration-tests-f9c25af86af611e7a942acbc328d8bb1
+project_code: sceptre-integration-tests-0b4c2dc0a86f11e7bb384a0002fca510
--- a/sceptre/stack.py
+++ b/sceptre/stack.py
@@ -8,7 +8,7 @@ with a particular stack.
 
 """
 
-import datetime
+from datetime import datetime, timedelta
 import logging
 import os
 import time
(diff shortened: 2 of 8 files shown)


integration-tests/sceptre-project/config/config.yaml
```diff
@@ -1 +1 @@
-project_code: sceptre-integration-tests-f9c25af86af611e7a942acbc328d8bb1
+project_code: sceptre-integration-tests-0b4c2dc0a86f11e7bb384a0002fca510
```
theseanything: You can ignore this change - I ran integration test locally which caused this.

sceptre/stack.py
```diff
@@ -157,9 +157,37 @@ def template(self):
...
+                    ])
+
+                s3_details = {
+                     "bucket_name": self.environment_config[
+                        "template_bucket_name"
+                     ],
+                     "bucket_key": template_key,
+                     "region": self.region
+                }
+
```
b-t-g: Could this be delegated to a separate function? It looks a bit long to me and it seems like the only part you really need is the s3_details, i.e., you don't do any weird things with manipulating self here or anything.

sceptre/template.py
```diff
@@ -127,97 +130,70 @@ def _call_sceptre_handler(self):
...
-        :param connection_manager: The connection manager used to make
-            AWS calls.
-        :type connection_manager: sceptre.connection_manager.ConnectionManager
         :returns: The URL of the template object in S3.
         :rtype: str
         :raises: botocore.exceptions.ClientError
 
         """
+        if not self.s3_details:
+            raise Exception()
```
b-t-g: There should be an error message, I think.

sceptre/template.py
```diff
@@ -127,97 +130,70 @@ def _call_sceptre_handler(self):
...
+
         self.logger.debug("%s - Uploading template to S3...", self.name)
 
         with self._boto_s3_lock:
-            if not self._bucket_exists(bucket_name, connection_manager):
-                self._create_bucket(region, bucket_name, connection_manager)
+            if not self._bucket_exists():
+                self._create_bucket()
```
b-t-g: Does there need to be error handling for when a bucket doesn't exist but a chosen name is already taken?

theseanything: Agree! Think however that would be new functionality - will do a separate PR.

b-t-g: Ah yes, that's definitely outside the scope of a "refactor" ticket.

sceptre/template.py
theseanything: Left in from testing - removed.

PR #214, opened 2017-10-10
Allow referencing of stack config & document string/case manipulation by jinja

JuanCanham:
* Add string manipulation examples to docs and to tests - This feature was already there and makes use of sceptre in situations with naming conventions and services that need specific cases easier (e.g s3 buckets)
* Add the ability to reference the stack config, object in the stack_config itself - This is primarily to practice DRY in configs, by allowing the stack_name (and potentially template) to be derived from the config filename - Again primarily useful in situations with naming conventions that do not match the Sceptre provided stack_name format
-----------------

* [ ] [Squashed related commits together][2].
* [x] Added unit tests.

### DIFF
--- a/docs/docs/environment_config.md
+++ b/docs/docs/environment_config.md
@@ -160,6 +160,16 @@ Any templated value can be supplied with a default value with the syntax:
 {% raw %}{{ var.value | default("default_value") }}{% endraw %}
 ```
 
+### String Manipulation
+
+Variables are rendered by Jinja2 and so can have their case change, can be sliced, split or combined with other strings
+
+```jinja2
+{% raw %}
+bucket_name: {{ environment_path.0.lower() }}-{{ environment_path.1.lower() }}-assets
+object_path: {{ environment_path.0.0 }}/{{ environment_path.1[-1] }}/{{ var.value.split("_")[-1] }}/dont/do/this/
+{% endraw %}
+```
 
 ## Examples
 
--- a/sceptre/config.py
+++ b/sceptre/config.py
@@ -178,7 +178,8 @@ class Config(dict):
                         environment_variable=os.environ,
                         var=user_variables,
                         environment_path=self.environment_path.split("/"),
-                        environment_config=environment_config
+                        environment_config=environment_config,
+                        stack_config=self
                     )
 
                     yaml_data = yaml.safe_load(rendered_template)
--- a/tests/fixtures/config/account/environment/region/security_groups.yaml
+++ b/tests/fixtures/config/account/environment/region/security_groups.yaml
@@ -6,3 +6,8 @@ parameters:
   param5: {{ environment_path.2 }}
   param6: {{ environment_config.region }}
   param7: {{ stack_config.name }}
+  param8: {{ stack_config.name.upper() }}
+  param9: {{ stack_config.name.0 }}
+  param10: {{ stack_config.name[-2:] }}
+  param11: {{ stack_config.name.split("_")[-1] }}
+  param12: {{ stack_config.name }}-string
--- a/tests/fixtures/config/account/environment/region/security_groups.yaml
+++ b/tests/fixtures/config/account/environment/region/security_groups.yaml
@@ -5,3 +5,4 @@ parameters:
   param4: {{ environment_path.1 }}
   param5: {{ environment_path.2 }}
   param6: {{ environment_config.region }}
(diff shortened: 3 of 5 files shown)


docs/docs/environment_config.md
```diff
@@ -160,6 +160,16 @@ Any templated value can be supplied with a default value with the syntax:
...
 
+### String Manipulation
+
+Variables are rendered by Jinja2 and so can have their case change, can be sliced, split or combined with other strings
+
+```jinja2
+{% raw %}
+bucket_name: {{ environment_path.0.lower() }}-{{ environment_path.1.lower() }}-assets
+object_path: {{ environment_path.0.0 }}/{{ environment_path.1[-1] }}/{{ var.value.split("_")[-1] }}/dont/do/this/
+{% endraw %}
```
theseanything: Could say this is an example?

sceptre/config.py
```diff
@@ -178,7 +178,8 @@ def get_config(path):
                         environment_variable=os.environ,
                         var=user_variables,
                         environment_path=self.environment_path.split("/"),
-                        environment_config=environment_config
+                        environment_config=environment_config,
+                        stack_config=self
```
theseanything: Whats intended to be made available here? - At this point dictionary values of stack_config is still being read, hence won't be and will be confusing. If its just the `stack_config.name` that we want, then maybe we should look at including it in the environment_path?

docs/docs/environment_config.md
```diff
@@ -160,6 +160,16 @@ Any templated value can be supplied with a default value with the syntax:
 {% raw %}{{ var.value | default("default_value") }}{% endraw %}
 ```
 
+### String Manipulation
```
theseanything: I think this is expanding on Jinja templating functionality - I feel this section should merge with the section above (`Template Defaults`) to be a more general "Jinja Templating" section; with a brief sentence saying what it is, and links to the Jinja docs

tests/fixtures/config/account/environment/region/security_groups.yaml
```diff
@@ -5,3 +5,9 @@ parameters:
   param4: {{ environment_path.1 }}
   param5: {{ environment_path.2 }}
   param6: {{ environment_config.region }}
+  param7: {{ stack_config.name }}
+  param8: {{ stack_config.name.upper() }}
+  param9: {{ stack_config.name.0 }}
+  param10: {{ stack_config.name[-2:] }}
+  param11: {{ stack_config.name.split("_")[-1] }}
+  param12: {{ stack_config.name }}-string
```
theseanything: params 8 to 12 are irrelevant as the are testing jinja's functionality not sceptre's.

JuanCanham: True, but if it's documented functionality, we don't explicitly test for jinja's usage in the config tests. This way if we change rendering engine we don't need to updates the tests, happy to drop the tests if it's easier though

docs/docs/environment_config.md
JuanCanham: Ok will split this out

sceptre/config.py
JuanCanham: The template name isn't part of the environment_path though, it's an attribute of the template/stack.

Currently it contains 
name < what i want
sceptre_dir < not sure why you would need that
environment_path < already exposed
dependencies < could be of interest to hooks

I was just reluctant to add a specific variable for my needs when something generic could work.

tests/fixtures/config/account/environment/region/security_groups.yaml
theseanything: Yup - but I feel its safe to document functionality thats been tested and documented in itself. This is testing specifically that rendering framework conforms the jinja templating specification. The onus of testing that should be within the rendering framework itself - if it claims to render jinja. Also v unlikely we'll change from jinja2.

sceptre/config.py
theseanything: Agree with above - also take back adding to the environment path as would change the API for existing config. Dependencies still wouldn't available at this point - as config hasn't been "read" in. Also in this context `name` is actually just the base_file_name. So could just add `basename=self.name` -> {{ basename }}

theseanything:
@JuanCanham any chance to look at this? Will close soon if no response.

theseanything:
No response, closing for now.

PR #227, opened 2017-10-20
Added hook to changesets

padilo:
This adds the possibility to add hooks points to changeset create and update operation, using this you can parametrize your template

```
hooks:
  before_create:
    - !my_hook "create param"
  before_update:
    - !my_hook "update param"
  before_create_change_set:
    - !my_hook "create_change_set param"
  before_update_change_set:
    - !my_hook "create_change_set param"
```

Before submitting the PR make sure the following are checked:

* [ ] Added unit tests.

### DIFF
--- a/sceptre/stack.py
+++ b/sceptre/stack.py
@@ -514,6 +514,7 @@ class Stack(object):
         )
         return response
 
+    @add_stack_hooks
     def create_change_set(self, change_set_name):
         """
         Creates a change set with the name ``change_set_name``.
@@ -595,6 +596,7 @@ class Stack(object):
             }
         )
 
+    @add_stack_hooks
     def execute_change_set(self, change_set_name):
         """
         Executes the change set ``change_set_name``.


theseanything:
Hi @padilo, could you give me a example use cases for this functionality?

padilo:
Sure.

Currently on our sceptre configuration we have a hook that populates a parameter with the AMI id. The hook basically builds the AMI, in case is not already build and populates a parameter of the sceptre config to be able to launch our cloudformation stack.

This works pretty well when launching or updating the stack, as sceptre has support to do it: `before_update`and `before_create`. But this hooks doesn't launch when you are creating a changeset. I was expecting that.

So to solve my issue I though the best way to do it and still being backward compatible is to have this two additional hooks `before_create_change_set` and `before_update_change_set`.

Here an exemple of our config:
```
stack_name: "zookeeper"
template_path: "cfn/zookeeper/zookeeper-cluster.json"

parameters:
  ClusterMaxSize: "5"
  ClusterMinSize: "2"
  ClusterDesiredSize: "3"

tags:
  Stack: zookeeper-cluster
hooks:
  before_create:
    - !packer "zookeeper ImageId"
  before_update:
    - !packer "zookeeper ImageId"
```

That will launch a packer called `zookeeper` (in case the image is not already build) and will populate `ImageId` as a parameter to launch the cloudformation stack.

This config doesn't work with changesets without this PR.

theseanything:
@padilo Could you not use a resolver? 
e.g. 
```
parameters:
    ami_id: !packer "zookeeper ImageId"
```

padilo:
Sounds very cool. I didn't know that can be done, I'll try it.

padilo:
Thanks @theseanything worked very well for my usecase the use of custom resolvers.

I still think this hooks could be useful for others or at least to keep in mind that create and update hooks aren't launched on changesets.

theseanything:
Awesome! I'm going to decline this PR for now - I'd like to open an issue to discuss this more.

connorwilliamscr:
For SAM stacks, you can only update with update-stack-cs, but there are no hooks that yet exist for this command. This would be super useful for writing a custom hook for running the aws cloudformation package command which needs to be run every time the Lambda code is updated and needs to be re-deployed.

If you try to use update-stack, the following error occurs: An error occurred (ValidationError) when calling the UpdateStack operation: UpdateStack cannot be used with templates containing Transforms.

PR #235, opened 2017-10-25
Generate .gitignore file when creating a new project with init command

seddarj:
When using the `init` command to create a new sceptre project, a `.gitignore` file is generated at the root of the project in order to ignore `.pyc` files.

This avoids committing a bunch of binary files to a repository in the first commit.

-----------------

* [ ] Added unit tests.

### DIFF
--- a/sceptre/cli.py
+++ b/sceptre/cli.py
@@ -630,6 +630,11 @@ def init_project(ctx, project_name):
         folder_path = os.path.join(project_folder, folder)
         os.makedirs(folder_path)
 
+    # Create .gitignore file in project
+    gitignore_path = os.path.join(project_folder, ".gitignore")
+    with open(gitignore_path, "wt") as gitignore_file:
+        gitignore_file.write("*.pyc")
+
     defaults = {
         "project_code": project_name,
         "region": os.environ.get("AWS_DEFAULT_REGION", "")


sceptre/cli.py
```diff
@@ -630,6 +630,11 @@ def init_project(ctx, project_name):
         folder_path = os.path.join(project_folder, folder)
         os.makedirs(folder_path)
 
+    # Create .gitignore file in project
+    gitignore_path = os.path.join(project_folder, ".gitignore")
+    with open(gitignore_path, "wt") as gitignore_file:
```
b-t-g: Why "wt" instead of just "w"?

sceptre/cli.py
```diff
@@ -630,6 +630,11 @@ def init_project(ctx, project_name):
...
+        gitignore_file.write("*.pyc")
+
```
b-t-g: I feel like your changes should be its own function instead of inlined.

b-t-g: This should probably be a flag instead (--gitignore and -g) of the default behavior (not all templates are version controlled, not everyone uses git, etc).

sceptre/cli.py
seddarj: I guess in python text mode is the default behavior and thus "w" would indeed be sufficient.

sceptre/cli.py
seddarj: I wasn't sure about that, but it probably should. Maintaining it would be easier that way.

seddarj: I'm not sure I agree with this.

The spirit of the feature is to avoid forgetting to add a `.gitignore` file, if the flag has to be added then it's just as easy to forget and the feature might be overlooked by most people. Or in other terms, it's almost as easy to add a `.gitignore` file as it is to remember to add the flag.

theseanything:
Hi @seddarj, closing this PR as it only address is fraction of real world use cases. Initialized projects may use other version control systems (hg, svn), non python templates, or be nested with other git repositories. To help, raise a Github issue to discuss future feature requests before PRs. Thanks! :)

PR #240, opened 2017-10-31
Auto create or update stacks with changesets

### LINKED ISSUE #198: Enhancement proposal: support stack creation via change sets

aehlke:
Resolves #198

Change sets are typically used for reviewing changes to an existing
stack before execution. However, they also support creating new stacks
via the --change-set-type=CREATE argument (which defaults to
--change-set-type=UPDATE). Sceptre does not support this parameter and
relies on the default.

There may be other compelling use cases for this, but a necessary one
is to be able to deploy SAM (Serverless Application Model) templates via
Sceptre. SAM templates are a great way to configure Lambda+API Gateway
services. For some reason, create-stack cannot be used with templates
that have a Transform directive; the workaround is to create a change
set and execute it with --change-set-type=CREATE to create the stack,
which works with SAM-transformed templates.

This commit automatically determines which change set type to use when
creating a change set.

The stack_exists method is a fork from awscli's has_stack method, from
https://github.com/aws/aws-cli/blob/bb92a4040900939e8d535923fe8bf3684c574a85/awscli/customizations/cloudformation/deployer.py#L36-L70

-----------------

* [x] Added unit tests.

### DIFF
--- a/sceptre/stack.py
+++ b/sceptre/stack.py
@@ -520,28 +520,57 @@ class Stack(object):
 
         :param change_set_name: The name of the change set.
         :type change_set_name: str
+        :raises: botocore.exceptions.ClientError
         """
-        create_change_set_kwargs = {
-            "StackName": self.external_name,
-            "Parameters": self._format_parameters(self.parameters),
-            "Capabilities": ['CAPABILITY_IAM', 'CAPABILITY_NAMED_IAM'],
-            "ChangeSetName": change_set_name,
-            "NotificationARNs": self.config.get("notifications", []),
-            "Tags": [
-                {"Key": str(k), "Value": str(v)}
-                for k, v in self.config.get("stack_tags", {}).items()
-            ]
-        }
-        create_change_set_kwargs.update(self._get_template_details())
-        create_change_set_kwargs.update(self._get_role_arn())
-        self.logger.debug(
-            "%s - Creating change set '%s'", self.name, change_set_name
-        )
-        self.connection_manager.call(
-            service="cloudformation",
-            command="create_change_set",
-            kwargs=create_change_set_kwargs
-        )
+        for change_set_type in ["UPDATE", "CREATE"]:
+            create_change_set_kwargs = {
+                "StackName": self.external_name,
+                "Parameters": self._format_parameters(self.parameters),
+                "Capabilities": ['CAPABILITY_IAM', 'CAPABILITY_NAMED_IAM'],
+                "ChangeSetName": change_set_name,
+                "NotificationARNs": self.config.get("notifications", []),
+                "ChangeSetType": change_set_type,
+                "Tags": [
+                    {"Key": str(k), "Value": str(v)}
+                    for k, v in self.config.get("stack_tags", {}).items()
+                ]
+            }
+            create_change_set_kwargs.update(self._get_template_details())
+            create_change_set_kwargs.update(self._get_role_arn())
+            self.logger.debug(
+                "%s - Creating %s change set '%s'",
+                self.name, change_set_type, change_set_name
+            )
+            try:
+                self.connection_manager.call(
+                    service="cloudformation",
+                    command="create_change_set",
+                    kwargs=create_change_set_kwargs
+                )
+            except botocore.exceptions.ClientError as e:
+                # If a stack does not exist, describe_stacks will throw an
+                # exception. Unfortunately we don't have a better way than
+                # parsing the exception msg to understand the nature of this
+                # exception.
+                msg = str(e)
+
+                if (
+                        "Stack with id {0} does not exist".format(
+                            self.external_name)
+                        in msg
+                ):
+                    self.logger.debug(
+                        "%s - Stack with id %s does not exist".format(
+                            self.name, self.external_name))
+                    continue
+                else:
+                    # We don't know anything about this exception. Don't handle
+                    self.logger.debug("%s - Unable to get stack details.",
+                                      self.name, exc_info=e)
+                    raise e
+            else:
+                break
+
         # After the call successfully completes, AWS CloudFormation
         # starts creating the change set.
         self.logger.info(
--- a/tests/test_stack.py
+++ b/tests/test_stack.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 
 import pytest
-from mock import patch, sentinel, Mock, MagicMock
+from mock import call, patch, sentinel, DEFAULT, Mock, MagicMock
 
 import datetime
 from dateutil.tz import tzutc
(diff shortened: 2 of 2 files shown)


aehlke:
How can I get my tests ran?

b-t-g:
@aehlke, for unit tests and linting, go to the root directory of sceptre and run `make lint` and `make test` and you'll be good to go. Given that AWS credentials are used for integration tests which are, rightly so, hidden by CircleCI to forks as stated here: https://circleci.com/docs/1.0/fork-pr-builds/, so the integration tests will be run after we review it and, if any problems arise, we can address them then.

aehlke:
@b-t-g sounds reasonable - I did run them locally and they pass (there are a couple linter errors unrelated to my changes as far as I can tell). It would be nice for the contribution docs to mention this, but it's not major. 😃

theseanything:
@aehlke - now added the ability for lint and unit tests to run on PRs. Can you fix the lint issues regardless?

sceptre/stack.py
```diff
@@ -514,19 +514,65 @@ def validate_template(self):
...
+
     def create_change_set(self, change_set_name):
         """
         Creates a change set with the name ``change_set_name``.
 
         :param change_set_name: The name of the change set.
         :type change_set_name: str
+        :raises: botocore.exceptions.ClientError
         """
+        change_set_type = "UPDATE" if self.stack_exists() else "CREATE"
```
theseanything: Could I suggest you mimic the pattern `launch()` - You could go ahead, assume UPDATE change_set_type and make the API call to create_change_set. Boto will throw an error if stack doesn't exist... catch that and retry the API call with CREATE change_set_type. This will also save an API call with an UPDATE change_set succeeds.

aehlke: I'll give that a shot. I mimicked the pattern used in awscli's [`deploy`](http://docs.aws.amazon.com/cli/latest/reference/cloudformation/deploy/index.html) command which first makes this call to decide how to proceed, borrowing that pattern blindly assuming that Amazon had thought it through, but without much of my own justification (asking forgiveness rather than permission is typically nicer, I agree).

aehlke:
@theseanything I've updated it with your suggested approach. No lint errors across the project, tests pass, maintained 100% coverage.

sceptre/stack.py
```diff
@@ -520,28 +520,57 @@ def create_change_set(self, change_set_name):
...
+                # parsing the exception msg to understand the nature of this
+                # exception.
+                msg = str(e)
+
+                if (
+                        "Stack with id {0} does not exist".format(
+                            self.external_name)
+                        in msg
+                ):
+                    self.logger.debug(
```
theseanything: Confusing debug message - not sure the `for` loop is the best way to go about it.... you could just have the `create_change_set` call instead of the `continue`?

sceptre/stack.py
```diff
@@ -520,28 +520,57 @@ def create_change_set(self, change_set_name):
...
+                )
+            except botocore.exceptions.ClientError as e:
+                # If a stack does not exist, describe_stacks will throw an
+                # exception. Unfortunately we don't have a better way than
+                # parsing the exception msg to understand the nature of this
+                # exception.
+                msg = str(e)
+
+                if (
+                        "Stack with id {0} does not exist".format(
```
theseanything: Could be a bit looser with the condition e.g. `msg.endswith("does not exist")` - you can also get the message from the exception via `msg = exp.response["Error"]["Message"]`

sceptre/stack.py
```diff
@@ -520,28 +520,57 @@ def create_change_set(self, change_set_name):
...
+                            self.external_name)
+                        in msg
+                ):
+                    self.logger.debug(
+                        "%s - Stack with id %s does not exist".format(
+                            self.name, self.external_name))
+                    continue
+                else:
+                    # We don't know anything about this exception. Don't handle
+                    self.logger.debug("%s - Unable to get stack details.",
```
theseanything: no need to log - information will be raised in the exception

sceptre/stack.py
aehlke: Since this is forked code from awscli, I minimized the differences beyond style. I typically favor this approach since this is noted in the commit message for these lines, and if we need to update this in the future when these things change, it's useful to be able to draw from updated awscli source.

ngfgrant:
@aehlke This looks like a really good PR and is extremely close to being accepted. Would you be able to take a second to review the remaining comments and make any changes? Would love to include this in our upcoming release!

aehlke:
@ngfgrant yes, the final remaining bits are WIP. I've been strapped for time but still intend to see this through.

(I'll leave this note of feedback as an aside: the extremely whitebox testing of implementation has made this challenging to iterate on and especially with some particular types of changes. With the style of extensively mocking things and testing the internal implementation details rather than more blackbox-style testing, whether via DI or functional style development or otherwise, makes the suite brittle to changes. Refactoring something that preserves behavior but changes implementation causes this style of test to break, when really that test should serve to build confidence that the refactoring was safe. This is a big topic at my workplace lately so I could continue on about this 😅 )

ngfgrant:
Thanks for the response @aehlke - I time is a challenge and appreciate your effort on this so far. Regarding the tests - agreed that testing is always an area that can be looked at for improvement!

ngfgrant:
@theseanything are we going to support this in V2? If not should we look to merge this into v1 branch and close this off?

aehlke:
@ngfgrant @theseanything now that we're in the new year, I have time again to pick this up to finish it off. Please let me know if it's worth doing so or if you have other plans now re: V2. Thanks.

ngfgrant:
@aehlke Thanks - Sean is away on holiday at the moment but I will pick this up with him as soon as he is back.

jakul:
+1 for getting this merged.

I want to migrate to Sceptre from Stacker, but this is a blocker for me

aehlke:
@jakul I'm waiting to hear back about project plans before moving on this

ngfgrant:
Hey @aehlke I would like this to be included in both V2 (master) and if possible V1 (on a separate branch). As it stands this PR is requesting to be merged into master (V2). However, it looks like you started the development on this when master was probably V1. If you could look at the remaining comments and resolve the conflicts I would be also keen to merge this ASAP. 

Thanks again - Niall :)

jnystromdev:
+1 for getting this merged.

Any updates? We want to start using Sceptre and having this functionality is a required for us.

ngfgrant:
@aehlke any chance of having a look at the remaining comments then resolving any conflicts. 🙏

ngfgrant:
Closing as this is has been resolved by https://github.com/cloudreach/sceptre/pull/326.

If there are any improvements that could be made using the code in this PR to the merged solution above please raise a new issue and PR.

PR #241, opened 2017-10-31
Initial CLI Refactor

theseanything:
This PR refactors the CLI into the following:
```
  create      Creates a stack or a change set.
  delete      Deletes a stack or a change set.
  describe    Commands for describing attributes of stacks.
  execute     Executes a change set.
  generate    Prints the template.
  init        Commands for initialising Sceptre projects.
  launch      Launch a stack or environment.
  list        Commands for listing attributes of stacks.
  set-policy  Sets stack policy.
  status      Print status of stack or environment.
  update      Update a stack.
  validate    Validates the template.
```

The CLI has been reorganized in to a folder `cli` splitting up commands into separate files. The `continue-update-rollback` command has been removed due to improvements in CloudFormation and maintain cli consistency. 

-----------------

* [x] Added unit tests.


sceptre/cli/describe.py
```diff
@@ -0,0 +1,151 @@
...
+from helpers import catch_exceptions, get_stack, get_stack_and_env, write
+
+
+@click.group(name="list")
+def list_group():
+    """
+    Commands for listing attributes of stacks.
+
+    """
+    pass
```
b-t-g: Should the reason for this be documented? I'm uncertain.

sceptre/cli/execute.py
```diff
@@ -0,0 +1,17 @@
...
+@click.argument("change-set-name", required=False)
+@click.pass_context
+@catch_exceptions
+def execute(ctx, path, change_set_name):
+    """
+    Executes a change set.
+
+    """
+    stack = get_stack(ctx, path)
+    stack.execute_change_set(change_set_name)
```
b-t-g: Is there another place where this function could be placed?

sceptre/cli/helpers.py
```diff
@@ -0,0 +1,267 @@
...
+
+    if not no_colour:
+        stack_status_colourer = StackStatusColourer()
+        stream = stack_status_colourer.colour(stream)
+
+    click.echo(stream)
+
+
+@contextlib.contextmanager
+def change_set(stack, name):
```
b-t-g: It seems to me like there is a better place for this function than here.

sceptre/cli/helpers.py
```diff
@@ -0,0 +1,267 @@
...
+
+    log_handler = logging.StreamHandler()
+    log_handler.setFormatter(formatter)
+    logger = logging.getLogger("sceptre")
+    logger.addHandler(log_handler)
+    logger.setLevel(sceptre_logging_level)
+    return logger
+
+
+def simplify_change_set_description(response):
```
b-t-g: Is there a good reason for defining this function here and describe.py?

sceptre/cli/describe.py
```diff
@@ -0,0 +1,151 @@
+import click
```
b-t-g: Name of this file is slightly misleading, since it contains all of the "read only" operations, instead of just the describe commands.

sceptre/cli/describe.py
theseanything: Leaving as cli group behaviours are documented with Click documentation

sceptre/cli/describe.py
theseanything: Yeah - so realised I can't split out "list" commands in to a list.py and import list, because it clashes with built-in python keyword "list". Not sure what to rename too?

sceptre/cli/helpers.py
theseanything: Can remove all together

sceptre/cli/helpers.py
theseanything: Yup duplicated function - removed from describe.py and imported from here

sceptre/cli/describe.py
theseanything: Creating extra list.py file

b-t-g: Expanding on this, we discussed offline that we can just do

```python
from list import list_group
```
for the purposes of using that module within Sceptre. It is more verbose, but it makes the structure of the commands more intuitive and, thus, more maintainable.

This may cause confusion if someone wanted to use Sceptre as a Python module and import the Sceptre list.py; however, we don't officially support that and users may sidestep this issue by adding the import statement:

```python
import list as list_commands
```

instead.

theseanything:
Requires #259 before merging.

Issue #234, opened 2017-10-25
Circular dependency cycles with more than two stacks hangs indefinitely


cboss24:
Sceptre will raise an error when two stacks are directly dependent on each other. However, Sceptre will hang indefinitely when a circular dependency with more than 2 nodes is encountered. 

The following setup:
```
# dev/a.yaml
template_path: template/vpc.yaml
parameters:
  B: !stack_output dev/b::B

# dev/b.yaml
template_path: template/vpc.yaml
parameters:
  C: !stack_output dev/c::C

# dev/c.yaml
template_path: template/vpc.yaml
parameters:
  A: !stack_output dev/a::A
```

Will produce the following results:
```bash
$ sceptre --debug launch-env dev
[2017-10-24 20:52:51] - sceptre.config - Reading in 'config.yaml' files...
[2017-10-24 20:52:51] - sceptre.config - Config: {'region': 'us-east-1', 'dependencies': [], 'project_code': 'test'}
[2017-10-24 20:52:51] - sceptre.environment - Initialising stack 'dev/a'
[2017-10-24 20:52:51] - sceptre.environment - Initialising stack 'dev/b'
[2017-10-24 20:52:51] - sceptre.environment - Initialising stack 'dev/c'
[2017-10-24 20:52:51] - sceptre.environment - Launching environment 'dev'
[2017-10-24 20:52:51] - sceptre.stack - dev/a - Loading dependencies...
[2017-10-24 20:52:51] - sceptre.config - Adding resolver yaml constructors
[2017-10-24 20:52:51] - sceptre.config - Added constructor for <class 'project_variables.ProjectVariables'> with node tag !project_variables
[2017-10-24 20:52:51] - sceptre.config - Added constructor for <class 'stack_output.StackOutput'> with node tag !stack_output
[2017-10-24 20:52:51] - sceptre.config - Added constructor for <class 'file_contents.FileContents'> with node tag !file_contents
[2017-10-24 20:52:51] - sceptre.config - Added constructor for <class 'environment_variable.EnvironmentVariable'> with node tag !environment_variable
[2017-10-24 20:52:51] - sceptre.config - Added constructor for <class 'stack_output.StackOutputExternal'> with node tag !stack_output_external
[2017-10-24 20:52:51] - sceptre.config - Adding hook yaml constructors
[...]

PR #251, opened 2017-11-08
Update dependency resolution to detect more general cycles

b-t-g:
As discussed in #234, a cycle of dependencies with more than two nodes fails to be caught by _check_for_circular_dependencies. This PR updates the algorithm to a depth-first search based approach, providing a more robust circular dependency detection.

The tests, I'd imagine will need some work, as I'm new to the terrifying world of Python mocking libraries.

Additionally, I thought integration tests would be helpful, but there were no such tests for detecting circular dependencies; was this a conscious decision?

-----------------

* [x] Added unit tests.
* [?] Added integration tests (if applicable).

### DIFF
--- a/sceptre/helpers.py
+++ b/sceptre/helpers.py
@@ -186,7 +186,7 @@ def _detect_cycles(stack, encountered_stacks):
     :returns: A dictionary containing all of the nodes encountered
     during the depth first search.
     """
-    for dependency in stack._get_launch_dependencies(stack.path):
+    for dependency in stack.dependencies:
         status = encountered_stacks.get(dependency, None)
         if status is "ENCOUNTERED":
             raise CircularDependenciesError(
--- a/sceptre/helpers.py
+++ b/sceptre/helpers.py
@@ -187,7 +187,7 @@ def _detect_cycles(stack, encountered_stacks, available_stacks):
     during the depth first search.
     """
     for dependency_name in stack.dependencies:
-        dependency = available_stacks[dependency_name]
+        dependency = available_stacks[dependency_name.split("/")[-1]]
         status = encountered_stacks.get(dependency, None)
         if status is "ENCOUNTERED":
             raise CircularDependenciesError(
--- a/sceptre/helpers.py
+++ b/sceptre/helpers.py
@@ -189,8 +189,8 @@ def _detect_cycles(stack, encountered_stacks):
         if status is "ENCOUNTERED":
             raise CircularDependenciesError(
                 "Found circular dependency involving "
-                "{0}"
-            ).format(dependency.stack_name)
+                "{0}".format(dependency.stack_name)
+            )
         elif status is None:
             encountered_stacks[dependency] = "ENCOUNTERED"
         else:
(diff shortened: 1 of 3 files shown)


cboss24:
Beat me to this one @b-t-g! Did you consider something like Tarjan's algorithm to detect all the strongly connected components in the graph. That way the error message could be richer (ex: `the following circular dependencies were found: [A, B, C] and [D, E]`

b-t-g:
@cboss24, I completely changed the error reporting based on your comment! I opted for a slightly modified DFS (keeping track of the nodes visited for each "root" and then report the path when a cycle is detected) instead of Tarjan's algorithm purely from an ease of implementation perspective. The downside is that it won't detect all the circular dependencies in one go, but I feel the intuitiveness of the algorithm/expected utility tradeoff is in favor DFS, but feel free to prove me wrong if so inclined!

cboss24:
That seems like a fair tradeoff! I was having a bit of a hard time wrapping my head around Tarjan's algorithm, so I understand wanting to trade a minor functionality decrease for less complexity. Nice work!

tests/test_environment.py
```diff
@@ -403,22 +400,55 @@ def test_get_delete_dependencies(self, mock_get_launch_dependencies):
...
+        stack3.name = "stack3"
+        stacks = {
+            "stack1": stack1,
+            "stack2": stack2,
+            "stack3": stack3
+        }
+        self.environment.stacks = stacks
+        with pytest.raises(CircularDependenciesError) as ex:
+            self.environment._check_for_circular_dependencies()
+        assert "['stack2', 'stack3', 'stack1']" in str(ex)
```
theseanything: Need to allow for different orders

tests/test_environment.py
```diff
@@ -403,22 +400,55 @@ def test_get_delete_dependencies(self, mock_get_launch_dependencies):
...
 
     def test_check_for_circular_dependencies_without_find_dependencies(self):
```
theseanything: Better name

tests/test_environment.py
```diff
@@ -403,22 +400,55 @@ def test_get_delete_dependencies(self, mock_get_launch_dependencies):
...
+        stack2.name = "stack2"
+        stacks = {
+            "stack1": stack1,
+            "stack2": stack2
         }
-
+        self.environment.stacks = stacks
         # Check this runs without throwing an exception
-        self.environment._check_for_circular_dependencies(dependencies)
+        self.environment._check_for_circular_dependencies()
```
theseanything: Test for more structures.

sceptre/helpers.py
```diff
@@ -176,3 +178,35 @@ def get_subclasses(class_type, directory=None):
                         classes[camel_to_snake_case(attr.__name__)] = attr
 
     return classes
+
+
+def _detect_cycles(stack, encountered_stacks, available_stacks, path):
```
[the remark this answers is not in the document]
b-t-g: As discussed earlier, the parameters here are unnecessary unless we want to commit to delete from available stacks as well, which may cause more confusion than we would gain.

Issue #254, opened 2017-11-14
Support YAML extension uploads to S3


luketn:
When the template is uploaded to S3, it now uses the file extension of the template to set the S3 key name extension. 

Creating a pull request now to address this.

theseanything:
I agree thats it confusing to upload a yaml file with .json extension. But whats the need to have the extension? Could we not just use a generic extension like `.template` ? CloudFormation itself doesn't care about the extension.

PR #255, opened 2017-11-14
Support YAML extension uploads to S3

luketn:
Addresses Issue #254
When the template is uploaded to S3, it now uses the file extension of the template to set the S3 key name extension

-----------------

* [ ] Added unit tests.

### DIFF
--- a/sceptre/template.py
+++ b/sceptre/template.py
@@ -127,6 +127,21 @@ class Template(object):
             sys.path.remove(os.path.join(os.getcwd(), directory))
         return body
 
+    def get_template_extension(self):
+        file_extension = os.path.splitext(self.path)[1]
+        if file_extension in {".json", ".yaml"}:
+            return file_extension
+        else:
+            for char in self.body:
+                if (not char.isspace()):
+                    # based on the first non-whitespace char in the rendered template
+                    if (char=="{"):
+                        return ".json"
+                    else:
+                        return ".yaml"
+
+        return ".json"
+
     def upload_to_s3(
             self, region, bucket_name, key_prefix, environment_path,
             stack_name, connection_manager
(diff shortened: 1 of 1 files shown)


theseanything:
We need to make sure we handle `.py` and `.j2` extensions - as they are rendered into `yaml` or `json` before upload. Would it be sensible to use a generic extension `.template`?

luketn:
Is there another property which tells you whether the .py or .j2 extension files will be rendered as YAML / JSON? Maybe we could use that instead of the file extension if so?

JuanCanham:
Python you have no way to know other than to test, j2 is usually used as a second suffix

so `src/template.yaml.j2` or if you hate yourself `src/template.json.j2`, Given you can use J2 to render yaml, i think **most** Python files will be troposphere and .'. json, but short of try json.loads, there is no way to be sure.

theseanything:
I agree thats it confusing to upload a yaml file with .json extension. But whats the need to have the specific extension? Could we not just use a generic extension like .template ? CloudFormation itself doesn't care about the extension. I'd like to minimize complexity of this fix.

b-t-g:
I should also note that the tests that are failed in CircleCI are my flakey tests which have since been fixed.

ngfgrant:
A couple of points:

- I agree with @theseanything that this could be simplified with the use of a generic ```.template``` extension. 

- If we really want to determine the file extension consider using the existing standard python libraries such as `pathlib`. Specifically, see `.suffixes` https://docs.python.org/3.4/library/pathlib.html#pathlib.PurePath.suffixes. I would argue that defining the extension of a file by its first non-whitespace character is not as robust as using `pathlib`. 

A couple of simple fixes:
- The commit messages don't adhere to the CONTRIBUTORS guide. 
- The initial Issue, whilst understandable is a little short on detail. Can I ask that the PR description is fleshed out a little bit more in sentences? Ensuring these small things are followed can help the project maintainers out with context when looking back on issues and commits.

ngfgrant:
This has been dealt with in v1 and currently not supported in v2. I am closing this for now.

PR #281, opened 2017-12-08
[Resolve #272] Enable integration tests in generic AWS accounts

### LINKED ISSUE #272: Integration tests run using Hard Coded AWS account
When I attempt to run the integration tests, after adding one I had to modify all the config files to an AWS account I actually have access to.
I don't want to break the IT automation, so did not submit a PR. 
There are two solutions without changing Sceptre:
1. Create an it-vars.yaml from which all the config files pull, so only one location to change when running locally.
2. Change the use of aws-role to profile, and have each user setup that profile with the account they have access to. (This is my preferred solution).

theseanything:
This PR removes the assume role for integrations tests and randomizes the bucket name. These properties were specific to the account linked to CircleCI and prevented others from running integration tests locally in other AWS accounts.

-----------------

* [x] Added unit tests.

### DIFF
--- a/integration-tests/environment.py
+++ b/integration-tests/environment.py
@@ -10,10 +10,15 @@ def before_all(context):
     context.project_code = "sceptre-integration-tests-{0}".format(
         context.uuid
     )
+    context.bucket_name = "sceptre-integration-tests-templates"
+
+    if not os.environ.get("CIRCLECI"):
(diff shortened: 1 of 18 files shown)


b-t-g:
Perhaps I'm missing something; the integration tests clearly still run but how given that we were previously removed the IAM role? Was it due to a configuration in Circle?

PR #285, opened 2017-12-13
[Resolve #283] Add Support for Multiple --var-file Options

### LINKED ISSUE #283: Add Support for Multiple `--var-file` Options
Currently, If I supply multiple `--var-file` options:

`sceptre --var-file=a.yaml --var-file=b.yaml --var-file=c.yaml launch-env api`

Sceptre will silently ignore `a.yaml` and `b.yaml`, while reading in `c.yaml`. This seems like an unnecessary limitation, especially when you consider that Sceptre already supports specifying multiple `--var` options. 

I believe there are quite a few reasons why you would want to do this. Our specific use case is that we would like one set of var-files to represent AWS account specific configs (dev.yaml, sit.yaml, prod.yaml), while also having a set of var-files which characterize how much the sceptre env will cost (tiny.yaml, stress_test.yaml, full.yaml).

One implementation detail to consider is what to do if the same key exists in multiple files. I think an appropriate solution would use the order in which the `--var-file`s were passed in to determine precedence.

cboss24:
Now able to specify multiple --var-file options when running
sceptre from the CLI. The dictionaries from these files will
be merged and exposed to the jinja templating. This allows
users to break a monolithic variable file into composable
pieces.

**Replace the text above the line with a summary of the changes in your PR.
The more detailed you are, the better. Before submitting the PR, make sure the following are checked.**

-----------------

* [x] Added unit tests.

### DIFF
--- a/sceptre/cli/__init__.py
+++ b/sceptre/cli/__init__.py
@@ -40,18 +40,18 @@ from sceptre.cli.helpers import setup_logging, catch_exceptions
 @click.option(
     "--var", multiple=True, help="A variable to template into config files.")
 @click.option(
-    "--var-file", type=click.File("rb"),
+    "--var-file", multiple=True, type=click.File("rb"),
     help="A YAML file of variables to template into config files.")
 @click.pass_context
 @catch_exceptions
 def cli(
         ctx, debug, directory, no_colour, output, var, var_file
-):  # pragma: no cover
+):
     """
     Sceptre is a tool to manage your cloud native infrastructure deployments.
 
     """
-    setup_logging(debug, no_colour)
+    logger = setup_logging(debug, no_colour)
     colorama.init()
     # Enable deprecation warnings
     warnings.simplefilter("always", DeprecationWarning)
@@ -63,11 +63,27 @@ def cli(
     }
     user_variables = {}
     if var_file:
-        user_variables.update(yaml.safe_load(var_file.read()))
+        for fh in var_file:
+            parsed = yaml.safe_load(fh.read())
+            # intersection
+            overloaded_keys = set(user_variables.keys()) & set(parsed.keys())
+            if overloaded_keys:
+                logger.debug(
+                    "Duplicate variables encountered: {0}. "
+                    "Using values from: {1}."
+                    .format(", ".join(overloaded_keys), fh.name)
+                )
+            user_variables.update(parsed)
     if var:
         # --var options overwrite --var-file options
         for variable in var:
             variable_key, variable_value = variable.split("=")
+            if variable_key in user_variables:
+                logger.debug(
+                    "Duplicate variable encountered: {0}. "
+                    "Using value from --var option."
+                    .format(variable_key)
+                )
             user_variables.update({variable_key: variable_value})
     if user_variables:
         ctx.obj["options"]["user_variables"] = user_variables
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -8,6 +8,7 @@ from uuid import UUID
 from click.testing import CliRunner
 from mock import MagicMock, patch, sentinel
 import pytest
+import click
 
 from sceptre.cli import cli
 from sceptre.config_reader import ConfigReader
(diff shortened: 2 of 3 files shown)


tests/test_cli.py
```diff
@@ -53,6 +55,48 @@ def raises_exception():
...
+                yaml.safe_dump({"a": 1}, foo)
+                yaml.safe_dump({"b": 2}, bar)
+
+            result = self.runner.invoke(
+                cli,
+                ["--var-file", "foo.yaml", "--var-file", "bar.yaml", "noop"]
+            )
+        self.patcher_getcwd.start()
+
+        templating_vars = yaml.safe_load(result.output)
```
cboss24: this is nasty, I am echoing a dict in the noop command, and then parsing the string output back into a dict here. Open to any suggestions on how to test what the cli group is setting in the ctx object.

theseanything:
Sorry for the late input - I put a implementation comment in the issue thread.

cboss24:
okay no longer throwing the error, now just producing a log.

cboss24:
also rebased off master, solution is now using the new CLI format

sceptre/cli/__init__.py
```diff
@@ -63,7 +63,17 @@ def cli(
...
     if var_file:
-        user_variables.update(yaml.safe_load(var_file.read()))
+        for fh in var_file:
+            parsed = yaml.safe_load(fh.read())
+            # intersection
+            overloaded_keys = set(user_variables.keys()) & set(parsed.keys())
+            if overloaded_keys:
+                logger.debug(
+                    "Duplicate variables encountered: {0}. "
+                    "Using values from file with highest precedence: '{1}'."
```
theseanything: Not sure we need `Using values from file with highest precedence: ` as it makes the message overly verbose and is pretty easy to which file is taking precedence.

sceptre/cli/__init__.py
```diff
@@ -63,7 +63,17 @@ def cli(
...
+                    .format(', '.join(overloaded_keys), fh.name)
+                )
+            user_variables.update(parsed)
     if var:
         # --var options overwrite --var-file options
```
theseanything: and we should also probably put a debug with similar effect here as well

tests/test_cli.py
```diff
@@ -55,6 +56,127 @@ def raises_exception():
         raises_exception()
         mock_exit.assert_called_once_with(1)
 
+    def test_single_var(self):
```
theseanything: These test could probably be parametrized (@pytest.mark.parametrize) - there is an example on line 172.

sceptre/cli/__init__.py
```diff
@@ -40,18 +40,18 @@
...
     "--var", multiple=True, help="A variable to template into config files.")
 @click.option(
-    "--var-file", type=click.File("rb"),
+    "--var-file", multiple=True, type=click.File("rb"),
     help="A YAML file of variables to template into config files.")
 @click.pass_context
 @catch_exceptions
 def cli(
         ctx, debug, directory, no_colour, output, var, var_file
-):  # pragma: no cover
```
theseanything: why did you remove `# pragma: no cover `?

sceptre/cli/__init__.py
b-t-g: Is the verbosity that much of an issue? I suspect it may be more helpful when there were unexpectedly repeated variables (the reasoning is similar to why I advocated for erroring out for duplicates) and it informs the user how we're selecting precedence without having to point them to a guide, which is a decently big usability gain. Especially since this is debug logging which is already verbose, anyway, adding all relevant information seems sensible to me.

sceptre/cli/__init__.py
cboss24: The `cli` function now has full coverage. I thought it would be okay to include it in the coverage report now. That way we will be able to detect if untested paths are added or tests are changed.

sceptre/cli/__init__.py
cboss24: Could we convey the same information with a simpler sentence?

`"Duplicate variables encountered: {}. Using values from {}".format(', '.join(overloaded_keys), fh.name)`

sceptre/cli/__init__.py
cboss24: done.

tests/test_cli.py
cboss24: Okay I gave it a go. Although the parametrized solution is less LOC, i'm not sure it's easier to read.

sceptre/cli/__init__.py
theseanything: After this change I'm happy with it!

cboss24: done!

cboss24:
@seanrankine all done and squashed, let me know if there's anything else!

PR #286, opened 2017-12-14
add bash completion script

kifbv:
Bash script providing command completion for Sceptre.
The command options are not handled.

-----------------

* [ ] Added unit tests.
* [ ] The new code doesn't generate flake8 (`make lint`) offenses.


b-t-g:
@kifbv, thanks for your PR! Unfortunately, I think the format of the commands will change rather drastically for 2.0 (which will be cut from the current master branch) so these bash completions would not work for this branch.

I'll keep the PR open to hear @theseanything's feedback. Namely, is this something that makes sense for updating for the 2.0 changes (and if you'd be interested in making those changes). I seem to recall part of the motivation for changing the structure of the commands was to piggy-back on native shell completion, but I don't know for sure whether that would remove the explicit need for sceptre-specific bash completions.

I'm personally in favor of adding bash completions, but I don't recall whether sceptre never had them for a particular reason, or if it's just something that hadn't been gotten to.

kifbv:
@b-t-g thanks for your feedback!
Just one thing, I'm not sure what you mean when you say 'piggy-back on native shell completion'. The stuff I wrote is actually using the facilities already provided by bash (_init_completion, compgen etc) to suggest the main command and then the stack and environment names according to what's in the current directory. I'll happily modify it when the commands in v2 are ready.

b-t-g:
You're right, my wording there was a bit sloppy; I meant to say that the new command format will (as I seem to recall), make more use of things that Bash already knows how to autocomplete (i.e., file/directory names), but there may still be gains to be had from autocompleting the rest of the commands. I'd like to move the discussion to this issue #287 for now, since I feel like issues are a better place to flesh out the ideas and fully discuss the ideas.

theseanything:
Ah yeah thanks @kifbv! Potentially we may just merge this into the v1 branch anyways. So the new CLI was just merged into master - take it with a pinch of salt as it may still change prior to v2 release. Overall, all for having autocomplete - wondering how do we get this to install along with Sceptre?

kifbv:
@b-t-g @theseanything thanks, I've commented in #287

b-t-g:
@ngfgrant, @kifbv, @theseanything, thinking about this a bit more, I'm not entirely sure whether it makes sense to add bash completions to the Sceptre repo given with how Sceptre (as a cli tool) is typically distributed (as a python package), given how bash completions are loaded (running `source <location of completion script>` per the comment by @kifbv on the issue I made in relation to this PR). 

That said, bash completions are indeed immensely useful, but I think that maybe they should live in a stand-alone repo (which can be advertised on the official Sceptre github page).

What are your thoughts, @kifbv?

ngfgrant:
@b-t-g I like the idea of potentially a separate repo. We could also follow the way that git offers bash autocomplete (https://git-scm.com/book/en/v1/Git-Basics-Tips-and-Tricks) and we have the bash autocompletion script in the main Sceptre repo with instructions on how to 'enable' it?

kifbv:
@b-t-g It's up to you really.

Some projects like [i3](https://github.com/i3/i3/tree/next/contrib) or [git](https://github.com/git/git/tree/master/contrib) keep this kind of stuff in a contrib directory, other like [kubernetes](https://github.com/kubernetes/contrib) also have that directory although they recommend creating a separate repo instead.

I personally prefer when projects have a contrib directory because I can then quickly see what 'add-ons' are available without moving to a different project. Also, I'm more likely to keep track of these external contributions when they're part of the original project.

The git auto-completion instructions mentioned by @ngfgrant are indeed well written and can be a good source of inspiration whatever direction you choose :)

If you prefer to have it separately, I'll create a repo where I can maintain both versions of the completions as I see fit and you can just add a pointer in Sceptre's README file.

ngfgrant:
I am leaning more toward doing this in the same way git handle their auto-complete. In contrib directory with instructions on how to enable.

ngfgrant:
@kifbv Decision made 😄 we will approach this as per my last comment. As part of this ticket, lets figure out how this would work and then we can create a new ticket to update the docs.

ngfgrant:
@kifbv In keeping with the way Git repo handles contrib, I have added a contrib directory and some instructions on how to use it. 

The contrib folder can be found in the V1 branch. Could you rebase this branch off of V1 branch and follow the instructions in contrib/README.md? The CLI names in this PR should be consistent with v1. 

Once this script is in contrib and is tested then we can merge this. Maybe once the V2 CLI has settled down a bit we can do the same for master branch (which is now v2). 

Thanks 😃

JohnVonNeumann:
In an attempt to provide a friendly nudge for this PR, I'd like to mention how useful I would find autocompletion, I'm sure this isn't news to anyone but yeah, if this could find it's way in, I'd be grateful.

ngfgrant:
@kifbv Any chance of updating to fix the conflicting files and we will get this into our `contrib` directory

kifbv:
Hey @ngfgrant , I'm not sure what happened there, i raised another PR: #373 
Sorry for the mess :/

ngfgrant:
Closing this PR as it has been addressed in https://github.com/cloudreach/sceptre/pull/373

PR #299, opened 2018-01-11
Add resolver and hook setup function

theseanything:
This PR adds a setup function to Hooks and Resolvers. This is to define a clearer lifecycle for these objects. Not explicit change to current functionality. Removes not so nice use of partials when reading config.

-----------------

* [x] Added unit tests.

### DIFF
--- a/sceptre/helpers.py
+++ b/sceptre/helpers.py
@@ -211,3 +211,30 @@ def _detect_cycles(node, encountered_nodes, available_nodes, path):
             )
             encountered_nodes[dependency] = "DONE"
     return encountered_nodes
+
+
+def _call_func_on_values(func, attr, cls):
+    """
+    Searches through dictionary or list for objects of type `cls` and calls the
+    supplied function `func`. Supports nested dictionaries and lists.
+    Does not detect objects used as keys in dictionaries.
+
+    :param attr: A dictionary or list to search through.
+    :type attr: dict or list
+    :return: The dictionary or list structure.
+    :rtype: dict or list
+    """
+
+    def func_on_instance(key):
+        if isinstance(value, cls):
+            func(attr, key, value)
+        elif isinstance(value, list) or isinstance(value, dict):
+            _call_func_on_values(func, value, cls)
+
+    if isinstance(attr, dict):
+        for key, value in attr.items():
+            func_on_instance(key)
+    elif isinstance(attr, list):
+        for index, value in enumerate(attr):
+            func_on_instance(index)
+    return attr
--- a/sceptre/hooks/__init__.py
+++ b/sceptre/hooks/__init__.py
@@ -2,6 +2,8 @@ import abc
 import logging
 from functools import wraps
 
+from sceptre.helpers import _call_func_on_values
+
 
 class Hook(object):
     """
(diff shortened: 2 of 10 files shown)


b-t-g:
You also mentioned that this work paves the way for work on existing issues; could you add those issue numbers in the commit message as well?

sceptre/helpers.py
```diff
@@ -211,3 +211,30 @@ def _detect_cycles(node, encountered_nodes, available_nodes, path):
...
     return encountered_nodes
+
+
+def _call_func_on_values(func, attr, cls):
+    """
+    Searches through dictionary or list for objects of type `cls` and calls the
+    supplied function `func`. Supports nested dictionaries and lists.
+    Does not detect objects used as keys in dictionaries.
+
+    :param attr: A complex data structure to search through.
```
b-t-g: This description of `attr` seems to obscure what the parameter is. My suggestion is to replace "A complex data structure" with "A dict or list". Otherwise, I take "complex data structure" to mean something much more complicated.

sceptre/helpers.py
```diff
@@ -211,3 +211,30 @@ def _detect_cycles(node, encountered_nodes, available_nodes, path):
...
+            func(attr, key, value)
+        elif isinstance(value, list) or isinstance(value, dict):
+            _call_func_on_values(func, value, cls)
+
+    if isinstance(attr, dict):
+        for key, value in attr.items():
+            func_on_instance(key)
+    elif isinstance(attr, list):
+        for index, value in enumerate(attr):
+            func_on_instance(index)
```
b-t-g: Could the two statements in this if block be made shorter (and just as readable, in my opinion) with list comprehensions?

sceptre/hooks/__init__.py
```diff
@@ -28,6 +39,42 @@ def run(self):
...
+        :return: The attribute stored with the suffix ``name`` in the instance.
+        :rtype: dict or list
+        """
+        return getattr(instance, self.name)
+
+    def __set__(self, instance, value):
+        """
+        Attribute setter which adds a stack reference to any hooks in the
+        data structure `value` and calls the setup method.
+
```
b-t-g: I think documenting the types of instance and value would be useful; even though they can be ascertained by reading the rest of the code, it would make it easier to understand what's going on quickly.

sceptre/resolvers/__init__.py
```diff
@@ -56,34 +66,23 @@ def __get__(self, instance, type):
...
     def __set__(self, instance, value):
-        setattr(instance, self.name, value)
-
-    def resolve_values(self, attr):
         """
-        Searches through dictionary or list for Resolver objects and replaces
-        them with the resolved value. Supports nested dictionaries and lists.
-        Does not detect Resolver objects used as keys in dictionaries.
+        Attribute setter which adds a stack reference to any resolvers in the
+        data structure `value` and calls the setup method.
```
b-t-g: Similar to the above suggestion for documenting the types of instance and value.

sceptre/helpers.py
theseanything: Going to leave as is, as not made much shorter by comprehensions.

PR #300, opened 2018-01-11
Stack_timeout config key for create and update stack

afraisse:
Added a stack_timeout key in stack yaml config files. Allow timeout and auto-rollback of stucked cloudformation deployments.

- For stack creations: timeout is simply passed in the underlying `create_stack` boto3 call using the **TimeoutInMinutes** parameter. Cloudformation solely handles the timeout behavior.
- For stack updates: `wait_for_completion` takes the timeout in as an optional parameter. If elapsed time overreach the timeout, a `cancel_update_stack` call is performed.

Hope this can be found useful by others. Still a lot of love for Sceptre here at Ubisoft, cheers!


sceptre/stack.py
```diff
@@ -50,10 +50,11 @@ class Stack(object):
     notifications = ResolvableProperty("notifications")
 
     def __init__(
-        self, name, project_code, template_path, region, iam_role=None,
-        parameters=None, sceptre_user_data=None, hooks=None, s3_details=None,
-        dependencies=None, role_arn=None, protected=False, tags=None,
-        external_name=None, notifications=None, on_failure=None
+            self, name, project_code, template_path, region, iam_role=None,
```
b-t-g: It looks like an extra level indent was added here.

sceptre/stack.py
```diff
@@ -192,7 +197,22 @@ def update(self):
             "%s - Update stack response: %s", self.name, response
         )
 
-        status = self._wait_for_completion()
+        status = self._wait_for_completion(self.stack_timeout)
+        # Cancel update after timeout
+        if status == StackStatus.IN_PROGRESS:
```
b-t-g: I think I would prefer if line 203 - 215 were consolidated into its own function; just to make things a bit neater.

sceptre/stack.py
```diff
@@ -612,6 +632,21 @@ def _get_role_arn(self):
         else:
             return {}
 
+    def _get_stack_timeout(self):
```
b-t-g: I find this name a bit misleading since it appears to be used elsewhere s a way to wrap the stack timeout for use as a keyword argument.

sceptre/stack.py
```diff
@@ -638,10 +675,20 @@ def _wait_for_completion(self):
...
+            while status == StackStatus.IN_PROGRESS and elapsed < timeout:
+                status = self._get_simplified_status(self.get_status())
+                self._log_new_events()
+                time.sleep(4)
+                elapsed += 4
+        else:
+            while status == StackStatus.IN_PROGRESS:
+                status = self._get_simplified_status(self.get_status())
+                self._log_new_events()
+                time.sleep(4)
```
b-t-g: I'm not a fan of the duplicated code here; my suggestion is to add an inner function called `timed_out(elapsed)` function (which just returns `true` if no timeout is defined and otherwise checks whether elapsed is >= 60*timeout) and then have this block be:
```python
             elapsed = 0
             while status == StackStatus.IN_PROGRESS and not timed_out(elapsed):
                 status = self._get_simplified_status(self.get_status())
                 self._log_new_events()
                 time.sleep(4)
                 elapsed += 4
```
There is some redundant work for when there isn't a timeout, but it's a trivial amount and makes the code clearer.

tests/test_stack.py
```diff
@@ -808,6 +859,11 @@ def test_get_role_arn_without_role(self):
     def test_get_role_arn_with_role(self):
         assert self.stack._get_role_arn() == {"RoleARN": sentinel.role_arn}
 
+    def test_get_stack_timeout(self):
+        assert self.stack._get_stack_timeout() == {
+            "TimeoutInMinutes": sentinel.stack_timeout
+        }
+
```
b-t-g: Although I am a fan of tests in general, I'm not entirely convinced this test is necessary.

b-t-g: I'd also like to get an opinion from @theseanything, especially the tests.

afraisse: It does not hurt though 😄

b-t-g: @afraisse, perhaps, but since that function being tested is obviously correct, I feel the ratio of utility of the test to the upkeep cost is not in its favor (any, even trivial change to that function, will require an update to that test, even if it's still obviously correct).

sceptre/stack.py
afraisse: Strange, I added those because my linter asked me to. I'll revert this.

tests/test_stack.py
afraisse: No argument here. I added it because of 
```python
def test_get_role_arn_with_role(self):
    self.stack._get_role_arn() == {"RoleARN": sentinel.role_arn}
```
I'll gladly remove it.

sceptre/stack.py
b-t-g: Odd. I'll defer to whatever flake8 says; but, since flake8 decided the old version is fine, I'm surprised your linter asked to indent another level.

afraisse: Seems like flake8 finds both acceptable. I used pylint, they may defer on some details I guess.

sceptre/stack.py
afraisse: I'll create a cancel_stack_update method then.
Should the log line:
```python
        self.logger.warning(
            "%s - Update stack time exceeded the specified timeout",
            self.name
        )
```
be kept the same? It assumes the stack cancellation happened because of a timeout exceed - which might not always be the case if the method is reused in the future.

b-t-g: Sounds good to me!

sceptre/stack.py
afraisse: I also follow the "convention" here because of the following method:
```python
def _get_role_arn(self):
```
which does exactly the same thing.

b-t-g: In that case, it's fine as it is (I still think the test case is unnecessary, though).

sceptre/stack.py
afraisse: I like the idea. I assumed you wouldn't want inner functions for some reason. I did as you suggested (although the time_out function should return `False` if there's no timeout, not `True`)

b-t-g: Yes, good call. I was too busy making sure I flipped the `>` that I overlooked the default behavior. 😆

afraisse:
I took into account your remarks. It's great to have that kind of thorough feedback, thank you!
I'll add an integration test and some documentation in an other commit.

afraisse:
I've added some doc and integration tests. Let me know it that works for you!

afraisse:
Thanks for the feedback! I'll wait for Sean's review before squashing commits together then.

docs/docs/stack_config.md
```diff
@@ -122,6 +123,10 @@ The ARN of a [CloudFormation Service Role](http://docs.aws.amazon.com/AWSCloudFo
 
 List of SNS topic ARNs to publish stack related events to. A maximum of 5 ARNs can be specified per stack. This configuration will be used by the `create-stack`, `update-stack`, and `create-change-set` commands. More information about stack notifications can found under the relevant section in the [AWS CloudFormation API documentation](http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_CreateStack.html).
 
+### stack_timeout
+
+A timeout in minutes before considering the stack deployment as failed. After the specified timeout, the stack will be rolled back.
```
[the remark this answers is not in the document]
afraisse: You're right, better to be explicit. I'll update that.

sceptre/stack.py
ngfgrant: I suppose the parameters are within parenthesis so the linter might not care? In any case I would agree with @b-t-g that this should have the additional level of indentation removed.

ngfgrant:
Thanks for the PR @afraisse. One request I would have would be to add a unit test that covers the cancels_stack_update() in the same style as the other parts of the project that are testing a boto3 function.

afraisse:
Hi @ngfgrant, thanks for your feedback. I've added an unit test like you requested.

afraisse:
Any news on this ?

ngfgrant:
Hi @afraisse - apologies been away on holiday. On th whole looks good to me. Agree with @b-t-g that @theseanything will want to take a final look at this.

stolendog:
I am also interested to have stack_timeout in my project. Is this going to be merged to master soon?

afraisse:
Hello guys, sorry for the delay. Squashed and rebased

PR #309, opened 2018-01-29
Allow newer boto versions

akdor1154:
The restriction on Boto < 1.5.0 is starting to get really annoying due to Python's lack of isolated dependencies between versions. We want to use new versions of botocore in other projects as they have support for profile better role assumption (`source_profile` in `~/.aws/config` stuff), but we can't make other projects require new boto versions as their installation then conflicts with Sceptre.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 behave==1.2.5
-boto3>=1.3.0,<1.5.0
+boto3>=1.3.0,<2
 bumpversion==0.5.3
 click==6.6
 colorama==0.3.7
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ with open("HISTORY.rst") as history_file:
     history = history_file.read()
 
 install_requirements = [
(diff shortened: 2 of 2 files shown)


ngfgrant:
Hi @akdor1154 - can you please create an issue for this (https://github.com/cloudreach/sceptre/blob/master/CONTRIBUTING.rst). If agreed upon then you will need to update your commit message. Please adhere to the checklist in your PR message. 

Following the guidelines helps us manage this project more effectively.

theseanything:
Pip checks requirements against those listed in setup.py. So the version needs to be changed [here](https://github.com/cloudreach/sceptre/blob/22b730b05373e0a99bd9d90bd9d0980c61e6f931/setup.py#L14).

akdor1154:
Should be ok? Sorry for the slowness in maintaining this. Also I've gone out on a limb and specified boto3<2 instead of <1.6.0, boto seems to take semver seriously enough. If this is not acceptable I can change it back to speciying the minor version as well.

PR #321, opened 2018-02-14
[Resolve #198] Create changesets for non-existent stacks

### LINKED ISSUE #198: Enhancement proposal: support stack creation via change sets

JuanCanham:
This is a cleaner implementation than #240 and will result in less failed AWS calls

-----------------

* [x] Added unit tests.


ngfgrant:
@JuanCanham hey man could you rebase and if required for version one branch from `v1` 🙏

PR #352, opened 2018-04-20
[Resolves #302] Handled the case in _check_circular_dependencies when there are external dependencies

### LINKED ISSUE #302: Detect Cycles Failing with Sibling Sub-Environments
When I split an environment into two sibling  sub-environments with a dependency between them, the `launch-env` command fails when I try to launch the sub-environment. 

Reproduction steps:
```shell
$ tree
.
├── config
│   ├── bundle
│   │   ├── a
│   │   │   └── x.yaml
│   │   └── b
│   │       └── y.yaml
│   └── config.yaml
└── templates
    └── template.yaml

$ cat config/bundle/a/x.yaml
template_path: template.yaml

parameters:
  SomeInput: "Input"

$ cat config/bundle/b/y.yaml
template_path: template.yaml

parameters:
  SomeInput: !stack_output a/x::SomeOutput

$sceptre launch-env bundle/b
  File "/Users/cbosmann/.local/share/virtualenvs/test-wZAZWvZX/bin/sceptre", line 11, in <module>
    sys.exit(cli())
  File "/Users/cbosmann/.local/share/virtualenvs/test-wZAZWvZX/lib/python3.6/site-packages/click/core.py", line 716, in __call__
  File "/Users/cbosmann/.local/share/virtualenvs/test-wZAZWvZX/lib/python3.6/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/Users/cbosmann/.local/share/virtualenvs/test-wZAZWvZX/lib/python3.6/site-packages/click/core.py", line 1060, in invoke
  File "/Users/cbosmann/.local/share/virtualenvs/test-wZAZWvZX/lib/python3.6/site-packages/click/core.py", line 889, in invoke
  File "/Users/cbosmann/.local/share/virtualenvs/test-wZAZWvZX/lib/python3.6/site-packages/click/core.py", line 534, in invoke
  File "/Users/cbosmann/.local/share/virtualenvs/test-wZAZWvZX/lib/python3.6/site-packages/click/decorators.py", line 17, in new_func
  File "/Users/cbosmann/.local/share/virtualenvs/test-wZAZWvZX/lib/python3.6/site-packages/sceptre/cli.py", line 96, in decorated
    return func(*args, **kwargs)
  File "/Users/cbosmann/.local/share/virtualenvs/test-wZAZWvZX/lib/python3.6/site-packages/sceptre/cli.py", line 315, in launch_env
    response = env.launch()
  File "/Users/cbosmann/.local/share/virtualenvs/test-wZAZWvZX/lib/python3.6/site-packages/sceptre/environment.py", line 127, in launch
    self._check_for_circular_dependencies()
  File "/Users/cbosmann/.local/share/virtualenvs/test-wZAZWvZX/lib/python3.6/site-packages/sceptre/environment.py", line 351, in _check_for_circular_dependencies
    [stack.name]
  File "/Users/cbosmann/.local/share/virtualenvs/test-wZAZWvZX/lib/python3.6/site-packages/sceptre/helpers.py", line 193, in _detect_cycles
    dependency = available_nodes[dependency_name.split("/")[-1]]
KeyError: 'x'
```

Is this expected behavior? If I remove the cycle detection this works fine.

ankitanallana:
When there's an **external** dependency while launching an environment, `_check_circular_dependencies` method in `environment.py` throws up an error when it should not be including that dependency as a part of a circular dependency check _within_ an environment.

This has been handled by adding `top_level_path` of an environment and comparing it against the the dependency. If they are **not** similar, we do not consider it for the circular dependency check.

-----------------

* [x] Added unit tests.

-----------------

P.S. - I am a part of Intuit, Inc and I work with [Omer Azmon](https://github.com/oazmon)

### DIFF
--- a/sceptre/environment.py
+++ b/sceptre/environment.py
@@ -282,12 +282,16 @@ class Environment(object):
         self.logger.debug("Checking for circular dependencies...")
 
         if self.stacks:
+            top_level_path = self.path + '/'  \
+                if self.path and self.path != '.' \
+                else ''
             encountered_stacks = {}
             available_nodes = {stack.name: stack for stack in self.stacks}
             for stack in self.stacks:
                 if encountered_stacks.get(stack, "UNENCOUNTERED") != "DONE":
                     encountered_stacks[stack] = "ENCOUNTERED"
                     encountered_stacks = _detect_cycles(
+                        top_level_path,
                         stack,
                         encountered_stacks,
                         available_nodes,
--- a/sceptre/environment.py
+++ b/sceptre/environment.py
@@ -10,7 +10,7 @@ represent a logical grouping of stacks as an environment.
 
 import logging
 import threading
-
+import os
 import botocore
 
 from concurrent.futures import ThreadPoolExecutor, wait
@@ -282,7 +282,7 @@ class Environment(object):
         self.logger.debug("Checking for circular dependencies...")
 
         if self.stacks:
-            top_level_path = self.path + '/'  \
+            top_level_path = os.path.join(self.path, '')  \
                 if self.path and self.path != '.' \
                 else ''
             encountered_stacks = {}
--- a/sceptre/helpers.py
+++ b/sceptre/helpers.py
@@ -182,7 +182,9 @@ def get_subclasses(class_type, directory=None):
     return classes
 
 
-def _detect_cycles(node, encountered_nodes, available_nodes, path):
+def _detect_cycles(
+        top_level_path, node, encountered_nodes,
+        available_nodes, path):
     """
     Use Depth-first search to detect cycles.
 
@@ -190,6 +192,10 @@ def _detect_cycles(node, encountered_nodes, available_nodes, path):
     during the depth first search.
     """
     for dependency_name in node.dependencies:
+        # if dependency not in available nodes, skip it
+        if not dependency_name.startswith(top_level_path):
+            continue  # skip
+
         dependency = available_nodes[dependency_name]
         status = encountered_nodes.get(dependency)
         if status == "ENCOUNTERED":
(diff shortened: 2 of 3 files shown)


siulcode:
+1
I’ve hit this wall several times before. Looking forward to the pr merge.

ngfgrant:
@ankitanallana thanks for this one. Could you rebase from master to update the branch?

Thanks!

ankitanallana:
Sure! I'll do that!

sceptre/environment.py
```diff
@@ -282,12 +282,16 @@ def _check_for_circular_dependencies(self):
         self.logger.debug("Checking for circular dependencies...")
 
         if self.stacks:
+            top_level_path = self.path + '/'  \
```
ngfgrant: Hi @ankitanallana could I ask that you use the default python path library when working with paths - it helps ensure some level of Windows support. 

In this case something like the following might work (assuming `top_level_path` is correct):

```py
if self.path and self.path != '.':
    top_level_path = os.path.join(self.path, '')
```

ankitanallana: sure! this is a good point! apologize for the delay, i shall incorporate this too!

ankitanallana:
Hello @ngfgrant ! This is done, hope it's alright now!

ngfgrant:
@ankitanallana 

So, first off thank you again for this PR. It definitely highlighted a popular issue with Sceptre and got me to look seriously at the way dependencies are handled. 

Having spent a lot of time recently around this part of the code I don't think this approach is going to work without introducing a bug, mainly that circular dependency checks are skipped if the top level is the same, which doesn't mean there are no circular deps. 

The real fix to this is much much more complicated (as I've found out) and I am now close to a robust solution which has involved ripping out quite a lot of the code and basically rewriting how configs are handled. I will be getting this out as of v2.0.0 and will have to abandon efforts for this in v1 as there is no easy/good fix. 

Thanks again for getting this to this stage and hope you can understand my response!

PR #378, opened 2018-05-24
Adding stack_timeout (same code as PR #300 with integration tests fixed)

jfgiroux:
Added a stack_timeout key in stack yaml config files. Allow timeout and auto-rollback of stucked cloudformation deployments.

For stack creations: timeout is simply passed in the underlying create_stack boto3 call using the TimeoutInMinutes parameter. Cloudformation solely handles the timeout behavior.
For stack updates: wait_for_completion takes the timeout in as an optional parameter. If elapsed time overreach the timeout, a cancel_update_stack call is performed.

Hope this can be found useful by others. Still a lot of love for Sceptre here at Ubisoft, cheers!

* I removed the default profile value because it crashes most of the tests right now.

-----------------

* [x] Wrote [good commit messages].
* [x] [Squashed related commits together]
* [x] Added unit tests.

### DIFF
--- a/docs/docs/stack_config.md
+++ b/docs/docs/stack_config.md
@@ -22,6 +22,7 @@ A stack config file is a yaml object of key-value pairs configuring a particular
 - [role_arn](#role_arn) *(optional)*
 - [notifications](#notifications) *(optional)*
 - [template_path](#template_path) *(required)*
+- [stack_timeout](#stack_timeout) *(optional)*
 
 
 ### dependencies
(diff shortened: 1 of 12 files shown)


jfgiroux:
@ngfgrant everyhting is fixed!

stolendog:
why it's not in seconds?

jfgiroux:
in the aws cli the option is in minute so we used minutes

--timeout-in-minutes (integer)

The amount of time that can pass before the stack status becomes CREATE_FAILED; if DisableRollback is not set or is set to false , the stack will be rolled back.

https://docs.aws.amazon.com/cli/latest/reference/cloudformation/create-stack.html

PR #395, opened 2018-06-11
[Resolves #257] improved Cross Region/Account Support

### LINKED ISSUE #257: Improve cross region/account support
We have limited support for multi region or account deployments. Currently you can have separate "environments" in different regions or accounts - (scoped by your aws cli profile). However, there is no support for stacks within a single "environment" to be in different regions/accounts. The currently implementation also restricts the ability to reference stacks in other account or regions for api operation such as `!stack_output`.

Proposed implementation would be to add a class attribute ConnectionManager to store and keep track of different boto sessions associated with each profile or stack_name. This would allow us to reference sessions used by other stacks and interact with that particular account/region. Also adds stack level granularity to connection_manager object rather than sharing one at an "environment" level.

This addresses #89 and #85

nabeelamjad:
This will deprecate the original functionality of using `iam_role` and instead allow the user to use an AWS Profile (supported by `boto` and `awscli`). The profile itself is usually defined in the `~/.aws/config` file (prefixed with `profile <name>`) but could be placed anywhere with the use of `AWS_CONFIG_FILE` which specifies the path to the configuration file.

This will break any sort of compatibility with the use of `iam_role` and instead a profile has to be created. The profile can be added to the sceptre configuration file in the same manner as the `iam_role` used to. 

The current level of precedence is as follows (and is all according to the default AWS Config/CLI behaviour linked below):
  1. Environment variable credentials take priority over profiles
  2. Region specified in the Sceptre `Stack` configuration will take priority over the one specified in the profile
  3. Profile takes priority last 

You are able to specify a profile which can target a different account and region while providing credentials through environment variables, instance role or else as required. More information on AWS profiles, configuration and credentials can be found [here](https://docs.aws.amazon.com/cli/latest/topic/config-vars.html).

Example stack configuration:

```yaml
profile: development-profile
project_code: cost-code
```

As far as the the YAML / sceptre resolver tag are concerned:
-  `!stack_output` should work as intended with full cross account/region functionality. 
- `!stack_output_external` has been redesigned slightly, it can now take in an optional 2nd arg, which is the profile. Example: `SomeParam: !stack_output_external SomeStack::Output SomeProfile` where `SomeProfile` is the profile defined in your AWS Config file to use for fetching that specific stack output. If the last argument isn't supplied then it is assumed to use whatever default is available.

-----------------

* [x] Added unit tests.

### DIFF
--- a/behave.ini
+++ b/behave.ini
@@ -1,0 +1,3 @@
+[behave]
+stderr_capture=False
+stdout_capture=False
--- a/integration-tests/environment.py
+++ b/integration-tests/environment.py
@@ -26,6 +26,8 @@ def before_all(context):
 
 
 def before_scenario(context, scenario):
+    os.environ.pop('AWS_REGION', None)
+    os.environ.pop('AWS_CONFIG_FILE', None)
     context.error = None
     context.response = None
     context.output = None
--- a/integration-tests/environment.py
+++ b/integration-tests/environment.py
@@ -26,8 +26,8 @@ def before_all(context):
 
 
 def before_scenario(context, scenario):
-    os.environ.pop('AWS_REGION', None)
-    os.environ.pop('AWS_CONFIG_FILE', None)
+    os.environ.pop("AWS_REGION", None)
+    os.environ.pop("AWS_CONFIG_FILE", None)
     context.error = None
     context.response = None
     context.output = None
--- a/integration-tests/features/stack-output-external-resolver.feature
+++ b/integration-tests/features/stack-output-external-resolver.feature
@@ -1,0 +1,9 @@
+Feature: Stack output external resolver
+
+  Scenario: launch a stack referencing the external output of an existing stack
+    Given environment "9" has AWS config "aws-config" set
+    and stack "9/A" exists using "dependencies/independent_template.json"
+    and stack "9/B" does not exist
+    and stack "9/B" has its project code resolved
+    When the user launches stack "9/B"
+    Then stack "9/B" exists in "CREATE_COMPLETE" state
--- a/integration-tests/features/stack-output-resolver.feature
+++ b/integration-tests/features/stack-output-resolver.feature
@@ -20,6 +20,15 @@ Feature: Stack output resolver
     and that stack "6/1/A" was created before "6/1/B"
     and that stack "6/1/B" was created before "6/1/C"
 
+  Scenario: launch a environment where stacks are in different regions
+    Given stack "6/2/A" does not exist in "eu-west-1"
+    and stack "6/2/B" does not exist in "eu-west-2"
+    and stack "6/2/C" does not exist in "eu-west-3"
+    When the user launches environment "6/2"
+    Then stack "6/2/A" in "eu-west-1" exists in "CREATE_COMPLETE" state
+    and stack "6/2/B" in "eu-west-2" exists in "CREATE_COMPLETE" state
+    and stack "6/2/C" in "eu-west-3" exists in "CREATE_COMPLETE" state
+
   Scenario: delete a stack referencing an output of existing stack
     Given stack "6/1/A" exists in "CREATE_COMPLETE" state
     and stack "6/1/B" exists in "CREATE_COMPLETE" state
--- a/integration-tests/sceptre-project/config/1/A.yaml
+++ b/integration-tests/sceptre-project/config/1/A.yaml
@@ -1,1 +1,1 @@
-template_path: templates/malformed_template.json
+template_path: templates/valid_template.json
--- a/integration-tests/sceptre-project/config/6/2/A.yaml
+++ b/integration-tests/sceptre-project/config/6/2/A.yaml
@@ -1,0 +1,2 @@
+template_path: templates/dependencies/independent_template.json
+region: eu-west-1
--- a/integration-tests/sceptre-project/config/6/2/B.yaml
+++ b/integration-tests/sceptre-project/config/6/2/B.yaml
@@ -1,4 +1,4 @@
-template_path: templates/dependencies/dependent_template.json
+template_path: templates/dependencies/dependent_template_local_export.json
 region: eu-west-2
 parameters:
   DependentStackName: !stack_output A::StackName
--- a/integration-tests/sceptre-project/config/6/2/C.yaml
+++ b/integration-tests/sceptre-project/config/6/2/C.yaml
@@ -1,4 +1,4 @@
 template_path: templates/dependencies/dependent_template.json
-region: eu-west-2
+region: eu-west-3
 parameters:
-  DependentStackName: !stack_output A::StackName
+  DependentStackName: !stack_output B::StackName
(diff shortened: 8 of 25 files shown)


JuanCanham:
Does this cover 
> The currently implementation also restricts the ability to reference stacks in other account or regions for api operation such as !stack_output.

looking at the test cases this only seems to cover multiple regions within one environment, but not referencing outputs from environments in other accounts/regions

nabeelamjad:
I've not had a look at that, I'll get that fixed and add some more integration test cases.

nabeelamjad:
I've covered the case for external stack outputs, unfortunately adding an integration test for that wasn't really feasible because:

  - It requires the user to have two AWS accounts to do integration testing
  - Our CircleCI config uses AWS creds for a single account, there needs to be an implementation to support multiple account profiles instead

The above are probably out of scope for this ticket, I'm more than happy to provide the templates used for multi account testing however (provided you have two profiles set up that reference a different account).

basvank:
Hi @nabeelamjad (or anyone else who has played around with this),
I'm playing around with cross-account (and thus cross-profile) `stack_output` resolvers in some stacks, and I was wondering if you have verified that that works as well? If I look closely at your test cases it seems that you are only testing across regions, right?

I ask because I have set up an environment with different accounts, and config files with profiles for the different accounts. It all works when I do updates on stacks that only involve one account, but when I do a `stack_output` resolver to another account it fails. When I look in CloudTrail I see that sceptre is doing a DescribeStacks in the account from where the resolver originates. It thus exits with the message: `An error occurred (ValidationError) when calling the UpdateStack operation: Stack [xxx-stack-name-xxx] does not exist`, which in that account is indeed the case.
The strange thing is that `stack_output_external` to that stack in the other account DOES work, but of course that is not as neat as I have to hardcode the whole stack name (including project_code) and the profile to use.

Let me know if you can help me out with solving this or when you need additional info. Also let me know when you think this should be in a new issue.

Thanks in advance!

nabeelamjad:
@basvank could you paste a very small directory structure/environment config structure?

`!stack_output_external` is only meant if the stack was created externally. `!stack_output` should work, but only if that stack in the same environment I believe (not 100% on this).

Here's an example of what I used to test and had no issues, let me know how you get on.

**Directory Structure:**
```
├── config
│   ├── 1
│   │   ├── A.yaml
│   │   └── B.yaml
│   └── config.yaml
└── templates
    └── dependencies
        ├── dependent_template.json
        └── independent_template.json
```

**config/1/A.yaml**
```
template_path: templates/dependencies/independent_template.json
profile: test
template_bucket_name: some-bucket-on-test

```

**config/1/B.yaml**
```
template_path: templates/dependencies/dependent_template.json
profile: dev
template_bucket_name: some-bucket-on-dev
parameters:
  DependentStackOutput: !stack_output A::Output
  DependentStackName: !stack_output A::StackName

```

**templates/dependencies/independent_template.json**

```
{
  "Resources": {
    "WaitConditionHandle": {
      "Type": "AWS::CloudFormation::WaitConditionHandle",
      "Properties": {}
    }
  },
  "Outputs": {
    "StackName": {
      "Value": {
        "Ref": "AWS::StackName"
      }
    },
    "Output": {
      "Value": "HelloFromStack1",
      "Export": {
        "Name": {
          "Fn::Sub": "${AWS::StackName}-Output"
        }
      }
    }
  }
}
```

**templates/dependencies/dependent_template.json**
```
{
  "Parameters": {
    "DependentStackOutput": {
      "Type": "String"
    },
    "DependentStackName": {
      "Type": "String"
    },
  },
  "Resources": {
    "WaitConditionHandle": {
      "Type": "AWS::CloudFormation::WaitConditionHandle",
      "Properties": {}
    }
  },
  "Outputs": {
    "StackName": {
      "Value": {
        "Ref": "AWS::StackName"
      }
    },
    "Output": {
      "Value": {
        "Fn::Sub": "Output from: ${DependentStackName} -> ${DependentStackOutput}"
      },
      "Export": {
        "Name": {
          "Fn::Sub": "${AWS::StackName}-Output"
        }
      }
    }
  }
}
```

**ONE of the following AWS configs based on your access format (you can override the location using  `AWS_CONFIG_FILE`)**

1. **~/.aws/config**  - AWS Creds format 
```
[default]
region = eu-west-1

[profile test]
region = eu-west-1
aws_access_key_id = ...
aws_secret_access_key = ...

[profile dev]
region = eu-west-1
aws_access_key_id = ...
aws_secret_access_key = ....
```

2. **~/.aws/config**  - AWS MFA format using assume role through another account (you can omit `role_arn` and add the credentials to the respected profile to use without assume role).
```
[default]
output = json
aws_access_key_id = ...
aws_secret_access_key = ...
region = eu-west-1

[profile test]
role_arn = arn:aws:iam::0123456789:role/SomeRole
source_profile = default
region = eu-west-1
mfa_serial = arn:aws:iam::0123456789:mfa/username

[profile dev]
role_arn = arn:aws:iam::0123456789:role/SomeRole
source_profile = default
region = eu-west-1
mfa_serial = arn:aws:iam::0123456789:mfa/username
```

basvank:
@nabeelamjad Thanks for the quick response! I have a very similar setup, so I looked a bit further and I think I have found the 'problem'. The cross account (probably cross profile) resolver does work when running `sceptre deploy` on the whole environment (if it contains both the source and the target of the resolver), but it does not work when running `sceptre deploy` or `sceptre update` on just the stack that contains the resolver. I guess that sort of makes sense. The only downside then is that I can't use change sets (because as far as I can see, sceptre does not support change sets for environments, correct me if I'm wrong).

So I think this is not a problem in the implementation.

nabeelamjad:
I will have a look for you and get back to you in a few days (slightly busy), I believe Sceptre should be able to resolve them even when updating a single stack. The other stack which is being resolved from should have a profile associated which it uses to make the call to fetch stack outputs, again not 100% sure on this as I wasn't involved in the core code but I will have a look.

ngfgrant:
@basvank Good news - this is merged in master and will be part of version 2.

basvank:
@ngfgrant Actually, I am on master and I know the feature has been merged, I'm using it heavily ;)

My problem is as follows, if we take the environment setup from @nabeelamjad as example:

If I have parameter in `config/1/A.yaml` that has a resolver pointing to an output in `config/1/B.yaml`, everything runs fine when I do `sceptre launch .`, even if stacks are in different profiles/accounts, so that is great.

The problem is when I then run `sceptre launch config/1/A.yaml`, and when B is in a different profile/account. I then get the error `KeyError: '1/B'`. So it cannot find the stack in the different profile/account then. It does work when the two stacks are in the same profile/account.

So I expect that when launching a subset of the stack, the profiles aren't used properly when doing the resolver lookup. Maybe the profiles that are outside the subset of the stack that is launched are not loaded properly?

ngfgrant:
Ah ok I see - could you take your previous post/examples and create a new issue for this so it doesn't get lost in this conversation. Sounds like something we should address as a priority.

basvank:
@ngfgrant Done

PR #401, opened 2018-06-13
[Resolve #390] Cast all parameters to strings

### LINKED ISSUE #390: Convert numbers to strings
Even though CFN only accepts strings, it would be nice if Sceptre can support numbers in config files to create the illusion that it does. See similar ansible issue: https://github.com/ansible/ansible/issues/25013

Example config:
```yaml
parameters:
  Port: 3000
  RuleNumber: 500
```

The workaround is to add quotes but it's just a bit awkward. Plus, you don't get to write the numbers with underscores as commas: e.g. `65_535`.

matrinox:
### Problem
Sceptre allows you to pass non-string values as parameters. Boto3 only allows string parameters. Refer to #390 for more details.

Example config:
```yaml
parameters:
  Port: 3000
  RuleNumber: 500
  LoggingEnabled: True
```

### Solution
To solve this, all non-string values are coerced into strings. An exception is made for booleans (True/False becomes "true"/"false"). List values are now recursively formatted. This has the side effect of supporting nested array values for parameters (no tests for this).

### Alternatives

1. Create boto3 PR to support non-string parameters for CloudFormation create/update stack APIs. This is also a possibility but it doesn't hurt to cast the values to strings here as well.
2. Raise exception when passing non-string values. This is slightly better than current – more descriptive error messages – but personally, it isn't as good as the proposed solution. Development time is roughly the same.
3. Do nothing or add this to readme. This is definitely the simplest and least riskiest solution but the UX is weaker than point #2.

### DIFF
--- a/sceptre/plan/actions.py
+++ b/sceptre/plan/actions.py
@@ -639,14 +639,27 @@ class StackActions(object):
         for name, value in parameters.items():
             if value is None:
                 continue
-            if isinstance(value, list):
-                value = ",".join(value)
             formatted_parameters.append({
                 "ParameterKey": name,
-                "ParameterValue": value
+                "ParameterValue": self._format_value(value)
             })
 
         return formatted_parameters
+
+    def _format_value(self, value):
+        """
+        Converts CloudFormation value to the format used by Boto3.
+        """
+        if isinstance(value, list):
+            # recurse
+            new_list = [self._format_value(item) for item in value]
+            return ",".join(new_list)
+        if value is True:
+            return "true"
+        elif value is False:
+            return "false"
+        else:
+            return str(value)
 
     def _get_role_arn(self):
         """
--- a/tests/test_actions.py
+++ b/tests/test_actions.py
@@ -726,12 +726,7 @@ class TestStackActions(object):
             "key2": None,
             "key3": None
         }
-        formatted_parameters = self.actions._format_parameters(parameters)
-        sorted_formatted_parameters = sorted(
-            formatted_parameters,
-            key=lambda x: x["ParameterKey"]
-        )
-        assert sorted_formatted_parameters == []
+        assert self.actions._format_parameters(parameters) == []
 
     def test_format_parameters_with_none_and_string_values(self):
         parameters = {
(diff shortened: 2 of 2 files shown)


sceptre/stack.py
```diff
@@ -618,15 +618,28 @@ def _format_parameters(self, parameters):
...
 
+    def _format_value(self, value):
+        """
+        Converts CloudFormation value to the format used by Boto3.
+        """
+        if isinstance(value, list):
+            # recurse
+            new_list = [self._format_value(item) for item in value]
+            return ",".join(new_list)
+        if value is True:
```
JuanCanham: Why do we want to downcase this?

matrinox: If your parameter is defined like this:
```yaml
Parameters:
  Key:
    Type: String
    AllowedValues:
      - False
      - True
```
Only `"false"`/`"true"` will work, not `"False"`/`"True"`. Same goes for `False`/`True`. So you need to downcase it.

JuanCanham: Makes sense

matrinox:
@ngfgrant Can you look into this PR please?

ngfgrant:
Hey @matrinox - yeh this is definitely on my list. I am aiming to get a v1 release out this week and then will look at this as potentially something for v2 release. It is definitely on my list though ;)

matrinox:
Awesome, thanks for the update!

ngfgrant:
@matrinox I am going to get another couple of PRs in relating to refactoring the way config is handled and then after I am done would it be ok if I ask you to update this from `master`? There is not big changes, mainly that the methods on stack have moved from that object so your code just needs pasted elsewhere 😄

matrinox:
@ngfgrant When you finish doing that, please mention again and I will get to it :).

stig:
Any chance of this being picked up? Numbers / booleans in the stack config file not being coerced to strings and this causing failure is the most frequent cause of `git commit --fixup` at our shop!

alexharv074:
Any chance of this or similar being merged? This has been a big pain point for my team.

matrinox:
@alexharv074 we already migrated to terraform so I'm not interested in updating this branch any longer. Sorry :(

zaro0508:
I have proposed an alternative idea to help with this in #925

PR #405, opened 2018-06-18
[Resolve #368] KMS resolver

### LINKED ISSUE #368: Managing values with KMS
It would be nice to be able to retrieve values, especially secrets, from AWS KMS and be able to reference it as sceptre variables.  I'm not sure of the best way to implement this but it looks like it might be possible by adding a sceptre resolver?

zaro0508:
Add a resolver that will retrieve and decrypt values from KMS.


ngfgrant:
Looks good to me - thanks for following the suggested template for `contrib/`. Could you update the branch? I'll get it merged in today.

zaro0508:
I've rebased onto latest v1 branch.

zaro0508:
Yes i can do that. However can I ask why this and #409 would not be considered core resolvers? I use use KMS and SSM in about 12 infra repos and it would be much nicer to have it as a core resolver than having to duplicately install and maintain these resolvers in every one of my repos.

Raffo:
I just wanted to say thanks for this PR, it is exactly what I was looking for ❤️

zaro0508:
@ngfgrant any thoughts on my last comment?

ngfgrant:
Hi @zaro0508 - thanks for the reminder on this. 

A couple of reasons why I'd prefer this to go into contrib/ at the moment. With the upcoming release of v2, v1 will shortly have a deprecation notice applied and will look into maintaining security fixes only. I am keen to add as few new things to the upcoming release as possible.

We have also created the [contrib/](https://github.com/cloudreach/sceptre/tree/v1/contrib) with these sorts of contributions in mind. Your work will still be accessible through out each of your v1 projects (as it will still be in the repo) and allows others to use and give feedback etc but there is no expectation that it is maintained as part of the core. 

I think there is a bigger piece of work to be done to help people contribute useful functionality to the project without requiring it to be accepted as part of the core. I have had a couple of thoughts about a plugin type system but will likely be something we won't see until well into v2 or even begining v3.

PR #408, opened 2018-06-18
[Resolves #187] Fix launch env dependency path

### LINKED ISSUE #187: Dependencies need to be prefixed with environment name
According to docs and comments from maintainers, if I have project
```yaml
config/
- config.yaml
- my-env/
  - first.yaml
  - second.yaml
      template_path: templates/second.yaml
      dependencies:
        - first
- templates/
  - first.yaml
  - second.yaml
```

Then `first` should be deployed before `second`. However, this (`sceptre launch-env my-env`) does not work (no error message is seen - the stacks launch successfully, but are instantiated in parallel).
EDIT2: the following does work:
```yaml
config/
- config.yaml
- my-env/
  - first.yaml
  - second.yaml
      template_path: templates/second.yaml
      dependencies:
        - my-env/first
- templates/
  - first.yaml
  - second.yaml
```
.

Would be great if this could be fixed, or docs updated at the very least.

ngfgrant:
As highlighted in #187, a user must specify a full path for stack
dependencies, even within the same environment. The documentation
shows the initial intention was to allow a user to reference the
stack name without the full path.

Given the following project structure:

config/
- config.yaml
- dev/
  - vpc.yaml
  - subnet.yaml
- templates/
  - vpc.yaml
  - subnet.yaml

Previously, a user had to specify dependencies in one of two ways:

```
dependencies:
    - dev/vpc
    - {{ environment_path.0 }}/subnet
```

This commit means that users can now specify dependencies as:

```
dependencies:
    - vpc
    - dev/subnet
    - {{ environment_path.0 }}/xyz
```

-----------------

* [x] Added unit tests.

### DIFF
--- a/sceptre/environment.py
+++ b/sceptre/environment.py
@@ -303,14 +303,20 @@ class Environment(object):
         }
 
         # Filter out dependencies which aren't under the top level environmnent
-        launch_dependencies = {
-            stack_name: [
-                dependency
-                for dependency in dependencies
-                if dependency.startswith(top_level_environment_path)
-            ]
-            for stack_name, dependencies in all_dependencies.items()
-        }
+        launch_dependencies = {}
+
+        for stack_name, dependencies in all_dependencies.items():
+            stack_dependencies = []
+            for dependency in dependencies:
+                if dependency.startswith(top_level_environment_path):
+                    stack_dependencies.append(dependency)
+                # If there is no slash in path, head will be empty.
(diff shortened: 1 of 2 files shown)


ufo525:
is this applied to version 2? I received error when trying to remove the environment from dependency

ngfgrant:
No this is v1 only. V2 requires you to have the path.

Issue #145, opened 2017-06-16
Alternative method(s) for resolver/hook registration


sftrabbit:
Sceptre currently looks for resolvers and hooks (and everything else) within specific locations in the directory structure. In some cases, it would be useful to be able to register resolvers and hooks programatically or with some configuration. This will also help with the distribution of common resolvers and hooks for community use - the developer wanting to use the resolver/hook will just add the dependency to their `requirements.txt` as normal and then write some kind of registry file that gets executed before doing any work. Something like:

```
from myhook import MyHook
from sceptre import register_hook

register_hook(MyHook)
# or maybe register_hook('my_hook', MyHook)
```

Alternatively, this could be done with some kind of configuration file:

```
hooks:
  my_hook: myhook.MyHook
```

although listing python imports in a config file feels a bit wrong to me.

jamesroutley:
Yep - this is a good idea, we'll have a think about a user-friendly way of using it.

akdor1154:
`setuptools` has functionality specifically for this - http://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins .
Would then be as simple as `pip install vault-resolver-my-custom-thingy`.

theseanything:
Solved with PR #205 - will be available in version 2.0

cboss24:
@theseanything it looks like that PR removed the ability to have a `hooks` folder at the sceptre root, and then have your hooks registered for `free`. What was the reasoning behind this, could we have best of both worlds where some hooks come from the setuptools entrypoint, while other come from the `hooks` folder?

stolendog:
The same question as @cboss24 wrote, why we don't have a possibility to register hooks from `hooks` folder in v2?

PR #409, opened 2018-06-18
[Resolve #406] SSM resolver

### LINKED ISSUE #406: Resolver for AWS SSM
Similar to issue #368, sceptre should include a resolver to get values from AWS SSM (the parameter store).

zaro0508:
Add a resolver that will retrieve and decrypt values from
the SSM parameter store.

Co-author: @et304383


zaro0508:
Yes i can do that.  However can I ask why this and #405 would not be considered core resolvers?  I use use KMS and SSM in about 12 infra repos and it would be much nicer to have it as a core resolver than having to duplicately install and maintain these resolvers in every one of my repos.

mattvonrocketstein:
I know this is closed, but... I would also benefit from this moving into core; I'm currently maintaining an almost exactly equivalent SSM resolver in `sceptre-resolvers.git` as submodule, checking it out in multiple Sceptre repos.  This functionality seems very core and a clear win, since as far as I know this makes sceptre much more functional out-of-the-box than cloudformation is with respect to secrets [ref](https://forums.aws.amazon.com/thread.jspa?messageID=795109&tstart=0).

@zaro0508 FWIW, my version assigns `self.connection_manager.boto_session.client('ssm')` in `__init__` instead of `resolve()`, which may or may not be slightly better depending on how smart boto_session is with caching.

et304383:
I agree that this should be core functionality.

ngfgrant:
Hi guys,

I wrote a response over on #405 as to why I asked for this to be added to `contrib/` but forgot to copy it across here. The response was:

> Hi @zaro0508 - thanks for the reminder on this.
> 
> A couple of reasons why I'd prefer this to go into contrib/ at the moment. With the upcoming release of v2, v1 will shortly have a deprecation notice applied and will look into maintaining security fixes only. I am keen to add as few new things to the upcoming release as possible.
> 
> We have also created the contrib/ with these sorts of contributions in mind. Your work will still be accessible through out each of your v1 projects (as it will still be in the repo) and allows others to use and give feedback etc but there is no expectation that it is maintained as part of the core.
> 
> I think there is a bigger piece of work to be done to help people contribute useful functionality to the project without requiring it to be accepted as part of the core. I have had a couple of thoughts about a plugin type system but will likely be something we won't see until well into v2 or even begining v3.

Whilst I absolutely agree that this functionality is useful, and for some people it is extremely useful it isn't going to be a core resolver for v1, nor v2. It is quite a specific piece of functionality and ideally contributions for core hooks and resolvers should be more generic such as (#140). As the project goes forward and supports multiple clouds there will be plenty of useful

The issue that has been highlighted is that many people use the same hooks/resolvers as part of different projects and it is a pain to have to copy them across projects and a pain to keep them updated. 

In v2 (master) hooks and resolvers can be pip installable (#145 ) and I think this a good way to quickly add specific hooks/resolvers like this to a project, whilst maintaining version control. 

So, I know exactly where you are coming from but I would much rather provide a better mechanism for users to extend core with this sort of thing, rather than growing core's responsibilities/functionality.

Hope that helps give some context?

Cheers,
Niall

PR #427, opened 2018-07-05
Resolve #426, support for !resolvers in `stack_tags`

### LINKED ISSUE #426: `stack_tags` does not support !resolvers
I have a custom resolver that tags my stacks with repo name/git hash, but using that I end up with cloudformation stack tags like `<resolver object>`.

mattvonrocketstein:
Resolve #426, support for !resolvers in `stack_tags`. This is pretty simple and coded defensively, but there are no tests as of now since I'm not yet familiar with the unit testing/mock setup.  I'm submitting this now anyway to see if there's interest in upsteaming this, to see if Travis starts barking at me, and to receive any guidance on where/how this should be tested.

### DIFF
--- a/sceptre/stack.py
+++ b/sceptre/stack.py
@@ -180,6 +180,20 @@ class Stack(object):
             )
         return self._external_name
 
+    @property
+    def stack_tags(self):
+        """
+        Returns the tags for this stack, rendering them if they use resolvers
+        :returns: The stack's tags
+        :rtype: list
+        """
+        out = []
+        for key, value in self.config.get("stack_tags", {}).items():
+            if hasattr(value, 'resolve'):
+                value = value.resolve()
+            out.append({"Key": str(key), "Value": str(value)})
+        return out
+
     @add_stack_hooks
     def create(self):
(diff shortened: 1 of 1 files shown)


ngfgrant:
@mattvonrocketstein Thanks for the PR! I am pretty full on at the moment but will get round to looking at this in more detail. You can see what impact your code has had to the coverage by running `make coverage`. From there you could see which lines aren't covered and we could work through the testing together if required?

mattvonrocketstein:
I fixed a similar problem in a way that I think works here as well, in a way that's more simple and general.  Does anyone see a problem with adding a `__str__` method to the Resolver base class, where the `__str__` method just calls `self.resolve()` ?

chhibber:
Just wanted to ask if there is an ETA to get this pulled into the 1.X branch?  We are trying to avoid forking or running another wrapper process on top of sceptre.

ngfgrant:
Closing as won't support in v1. v1 EOL as of tomorrow. Will look to support in V2 if not already done so.

PR #429, opened 2018-07-08
[Resolves #438] Add contrib to the sceptre package

### LINKED ISSUE #438: Add contrib to the sceptre package
Contributions are in the same repo as core sceptre therefore it's
versioned with core sceptre. This would mean that it should also
be released with sceptre.

zaro0508:
Contributions are in the same repo as core sceptre therefore it's
 versioned with core sceptre.  This would mean that it should also
 be released with sceptre.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,8 @@ setup(
     packages=[
         "sceptre",
         "sceptre/resolvers",
-        "sceptre/hooks"
+        "sceptre/hooks",
+        "contrib"
     ],
     package_dir={
         "sceptre": "sceptre"


setup.py
```diff
@@ -46,6 +46,7 @@
         "sceptre",
         "sceptre/resolvers",
         "sceptre/hooks"
```
ngfgrant: Hi @zaro0508  - the build is failing. I think there needs to be a comma after `"sceptre/hooks"` before listing the new package.

zaro0508:
thanks @ngfgrant i think it's fixed now.

ngfgrant:
LGTM & integration tests passing locally.

PR #440, opened 2018-07-30
[Resolves #298] Fix explicit dependencies overriding implicit dependencies

seddarj:
Fix explicit dependencies overriding implicit dependencies at least in sceptre v1

-----------------

* [x] Wrote [good commit messages][1].
* [x] [Squashed related commits together][2] after the changes have been approved.
* [x] Commit message starts with `[Resolve #issue-number]` (if a related issue exists).
* [ ] Added unit tests.
* [ ] Added integration tests (if applicable).
* [x] All unit tests (`make test`) are passing.
* [x] Used the same coding conventions as the rest of the project.
* [x] The new code doesn't generate flake8 (`make lint`) offenses.
* [x] The PR relates to *only* one subject with a clear title
  and description in grammatically correct, complete sentences.

[1]: http://chris.beams.io/posts/git-commit/
[2]: https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

PR #444, opened 2018-07-31
[Resolves #298] Fix stack config dependencies merging

### LINKED ISSUE #298: Explicit dependencies not honoured
Sceptre uses implicit dependencies between stacks (ie when a value is read from another stack a dependency on that stack is created).  It also provides a mechanism for explicit declaration of dependencies however the two mechanisms are in conflict resulting in failed stack deployments.

To reproduce:
* create configuration for stacks A and B
* read the output name from stack A as a parameter in stack B
* declare an explicit dependency on stack A
* deploy stacks

DONE WHEN:
* the explicit dependencies override the implicit dependencies

ngfgrant:
Previously, the `dependencies` data structure in stacks were not being merged properly. This resulted in issues, such as #298, being reported. For example, in the sceptre project displayed below, if `bundle/b/y.yaml` had a `!stack_ouput` from `bundle/a/x.yaml` an implicit dependency is created. If we then specified that `bundle/b/y.yaml` had a dependency on `bundle/a/x.yaml` the dependencies would not resolve properly. 

```
.
├── config
│   ├── bundle
│      ├── a
│      │   └── x.yaml
│      ├── b
│      │   └── y.yaml
│      ├── c
│      │   └── z.yaml
│      └── config.yaml
│   
├── templates
    └── template.json
```

The fix submitted with this commit gives us flexibility in dealing with different merge strategies for config items, if required in the future. 

-----------------

* [x] Added unit tests.
* [x] The new code doesn't generate flake8 (`make lint`) offences.

### DIFF
--- a/sceptre/config.py
+++ b/sceptre/config.py
@@ -28,7 +28,7 @@ from . import strategies
 ConfigAttributes = collections.namedtuple("Attributes", "required optional")
 
 CONFIG_MERGE_STRATEGIES = {
-    'dependencies': strategies.child_wins,
+    'dependencies': strategies.list_join,
     'hooks': strategies.child_wins,
     'parameters': strategies.child_wins,
     'protect': strategies.child_wins,
--- a/sceptre/config.py
+++ b/sceptre/config.py
@@ -159,9 +159,11 @@ class Config(dict):
 
         Traverses the environment path, from top to bottom, reading in all
         relevant config files. If config items appear in files lower down the
-        environment tree, they overwrite items from further up. Jinja2 is used
-        to template in variables from user_variables, environment variables,
-        and the segments of the environment path.
+        environment tree, they overwrite items from further up, depending on
+        the config merge strategy used.
+
+        Jinja2 is used to template in variables from user_variables,
+        environment variables, and the segments of the environment path.
 
         :param user_variables: A dict of key value pairs to be supplied to \
         the config file via Jinja2 templating.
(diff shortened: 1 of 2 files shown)


seddarj:
@niallgrantcloudreach Regarding our conversation in #440, I tested the changes from the stack_dependencies_fix branch and it doesn't seem to fix the explicit dependencies overriding the implicit ones.

My best guess is that you need to apply a `list_join` strategy when merging the `dependencies` list from self and from config when doing the `self.update(config)` [here](https://github.com/cloudreach/sceptre/blob/a81087b4b24829c5c53595d3d41c4e1d30e5f0c0/sceptre/config.py#L213)

What do you think?

seddarj:
Maybe something like this:
```
config = get_config(path)
######################################
dependencies = strategies.list_join(
    self.get("dependencies"),
    config.get("dependencies")
)
config.update({"dependencies": dependencies})
######################################
self.update(config)
```

ngfgrant:
Yes, my fault, this should be using the list join strategy - that is why we wrote it. I will just get a commit in for that now.

seddarj:
@ngfgrant Sorry to insist but I think there's still an issue here.

Your last commit addresses the merging of the explicit dependencies from different cascading configs but doesn't address the merging of implicit dependencies located in `self` and explicit dependencies located in `config`.

ngfgrant:
@seddarj Thanks again for testing and pointing out the missing link. That should now be resolved broadly based on your suggestion in your previous comment.

seddarj:
Great! Thanks @ngfgrant

seddarj:
Seems to be working well ;-)

PR #453, opened 2018-08-03
Bump pyyaml version to prevent failures on Python 3.7 enviroments

ror6ax:
Also, build on 3.7


ngfgrant:
Hey thanks for the PR. 

We aren't going to put 3.7 in as supported for now. I have however increased the version allowed for PyYaml so you should be able to use 3.7 if you so wish.

PR #473, opened 2018-09-20
[Resolve #140]: for V1: Nonscalar arguments for hooks and resolvers. 

### LINKED ISSUE #140: Nonscalar resolver and hook arguments

akdor1154:
Rebase of #141 . See original by @sftrabbit for details.

### DIFF
--- a/sceptre/config.py
+++ b/sceptre/config.py
@@ -7,6 +7,7 @@ This module implements a Config class, which stores a stack or environment's
 configuration.
 """
 
+import copy
 import logging
 import os
 import yaml
@@ -282,7 +283,7 @@ class Config(dict):
             :rtype: func
             """
             return lambda loader, node: node_class(
-                loader.construct_scalar(node),
+                loader.construct_object(self.resolve_node_tag(loader, node)),
                 connection_manager,
                 environment_config,
                 self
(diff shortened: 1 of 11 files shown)


akdor1154:
The one hesitation in the original PR was that nested `stack_output`s add dependencies to their owning stacks properly; I've added a test to ensure this functions appropriately (no code changes were needed).

ngfgrant:
@akdor1154 can you get this rebased on v1 and I will get it merged next. Should be in for the next v1 release.

akdor1154:
should be good to go

PR #477, opened 2018-10-01
Support for custom Jinja filters

mattvonrocketstein:
Corresponding issue is [here](https://github.com/cloudreach/sceptre/issues/476).  Syndicating content: My use-case involves adding new custom Jinja filters into projects using the v1 releases of sceptre, which I haven't found a supported way to accomplish.

I think this should work in way similar to hooks/resolvers, i.e. loading this project-specific functionality from "jinja_filters" folder underneath the sceptre root.

This PoC here is rough but working.  Mostly this approach feels bad because I can't see how to access any kind of global sceptre config to get at information like the --dir argument applied by the CLI.  I've had to configure the directory for external jinja filters from an environment variable for now to work around this.

Obviously even if smoothed this pull would needs corresponding docs changes.  Waiting for some guidance about how to move forward and whether/how this could be accepted into 1.x mainline (my org probably can't commit to v2 soon)

### DIFF
--- a/sceptre/template.py
+++ b/sceptre/template.py
@@ -8,6 +8,7 @@ and implements methods for uploading it to S3.
 """
 
 from datetime import datetime
+import glob
 import imp
 import logging
 import os
@@ -37,7 +38,6 @@ class Template(object):
 
     def __init__(self, path, sceptre_user_data):
         self.logger = logging.getLogger(__name__)
-
         self.path = path
         self.sceptre_user_data = sceptre_user_data
         self.name = os.path.basename(path).split(".")[0]
@@ -97,7 +97,7 @@ class Template(object):
         # NB: this is a horrible hack...
         relpath = os.path.relpath(self.path, os.getcwd()).split(os.path.sep)
         relpaths_to_add = [
-            os.path.sep.join(relpath[:i+1])
+            os.path.sep.join(relpath[:i + 1])
             for i in range(len(relpath[:-1]))
         ]
         # Add any directory between the current working directory and where
@@ -270,8 +270,7 @@ class Template(object):
                 }
             )
 
-    @staticmethod
-    def _render_jinja_template(template_dir, filename, jinja_vars):
+    def _render_jinja_template(self, template_dir, filename, jinja_vars):
         """
         Renders a jinja template.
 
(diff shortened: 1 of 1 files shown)


sceptre/template.py
```diff
@@ -291,8 +290,50 @@ def _render_jinja_template(template_dir, filename, jinja_vars):
...
+
+        This (optional) feature relies on the environment variable
+        ${SCEPTRE_JINJA_FILTER_ROOT} because we don't have easy access
+        from here to CLI parsing info or global configuration info.
+        """
+        # give back cached jinja filters if available
+        if hasattr(self, '_jinja_filters'):
+            return self._jinja_filters
+        # load jinja filters if not already cached
+        env_var_name = 'SCEPTRE_JINJA_FILTER_ROOT'
```
nbjcn1: I don't think we should be reaching env_vars directly, can we read this as environment config for consistency

sceptre/template.py
```diff
@@ -291,8 +290,50 @@ def _render_jinja_template(template_dir, filename, jinja_vars):
...
+        else:
+            filter_dir = os.environ[env_var_name]
+            if not os.path.exists(filter_dir):
+                err = '${} is set, but directory does not exist!'
+                raise ValueError(err.format(env_var_name))
+            else:
+                msg = 'loading jinja filters from: {}'
+                self.logger.debug(msg.format(filter_dir))
+                self._jinja_filters = {}
+                for fpath in glob.glob(os.path.join(filter_dir, '*.py')):
```
nbjcn1: Ideally this would be done using entry points, as per the new hooks, resolves code

sceptre/template.py
mattvonrocketstein: How exactly should that be done?  `Template()` seem to only be instantiated with with scepre_user_data, not environment_config.  See also this issue https://github.com/cloudreach/sceptre/issues/355

Does a path from environment_config or stack_config actually help here?  I think what is really needed is the exact value passed from the CLI argument `--dir`, because even given an environment or stack path, if env folders are nested then one couldn't be sure how far to go back up the directory tree to get to the sceptre root.

sceptre/template.py
mattvonrocketstein: Is there an example of that in v1, or would this need backport from v2?

ngfgrant:
Closing as won't support in v1. v1 EOL as of tomorrow.

Would consider support for v2?

PR #479, opened 2018-10-10
[Resolve #445] add Dockerfile and update README

### LINKED ISSUE #445: An official Docker image?
I have just started using `sceptre` _(sweet! no more hand-rolling a bash script for every project's set of stacks)_ and run it as a Docker container. Since there's no official image, either I can use out-dated third-party images from Dockerhub, or roll my own.

It would be much nicer (and quicker) to be able to pull official images, particularly when a new version of `sceptre` comes out.

This seems to do the trick:  https://github.com/cloudreach/sceptre/compare/v1...rgitzel:docker-v1  

A few `docker` commands on each release would be enough to deploy a new image:
```
docker build -t sceptre-local .

docker tag sceptre-local cloudreach/sceptre:1.3.4
docker push cloudreach/sceptre:1.3.4

docker tag sceptre-local cloudreach/sceptre:latest
docker push cloudreach/sceptre:latest
```

rgitzel:
This Dockerfile lets you run Sceptre without installing it. 

Note that the image needs to be published before the examples in the README will work.  

However, if you replace `cloudreach` with `rgitzel` in the Docker image names, the examples should work.

-----------------

* [ x ] Wrote [good commit messages][1].
* [ x] [Squashed related commits together][2] after the changes have been approved.
* [ x] Commit message starts with `[Resolve #issue-number]` (if a related issue exists).
* [n/a] Added unit tests.
* [ n/a] Added integration tests (if applicable).
* [x ] The new code doesn't generate flake8 (`make lint`) offenses.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,42 @@
+FROM python:3.6-alpine
+
+#
+# a simple image for running Sceptre without installing Python
+#
+# to build it:
+#   $ docker build . -t sceptre-local
+#
+# to then deploy it to Dockerhub after version X.Y.Z is released:
+#   $ docker login
+#   $ docker tag sceptre-local rgitzel/sceptre:X.Y.Z
+#   $ docker push rgitzel/sceptre:X.Y.Z
+#
+# if it's the newest version:
+#   $ docker tag sceptre-local rgitzel/sceptre:latest
+#   $ docker push rgitzel/sceptre:latest
+
+
+# we'll copy the repo files to here
+ARG SRC_FOLDER="/src"
+
+# expect the folder of Sceptre files to be mounted here
+ARG SCEPTRE_FOLDER="/sceptre"
+
+# copy the relevant files
+RUN mkdir $SRC_FOLDER
+COPY contrib $SRC_FOLDER/contrib
+COPY HISTORY.rst $SRC_FOLDER
+COPY Makefile $SRC_FOLDER
+COPY README.rst $SRC_FOLDER
+COPY requirements.txt $SRC_FOLDER
+COPY sceptre $SRC_FOLDER/sceptre
+COPY setup.* $SRC_FOLDER/
+
+# install from those files
+RUN pip install -e $SRC_FOLDER
+
+# Sceptre files should be here
+VOLUME ["$SCEPTRE_FOLDER"]
+WORKDIR $SCEPTRE_FOLDER
+
+ENTRYPOINT [ "sceptre" ]
--- a/README.rst
+++ b/README.rst
@@ -83,11 +83,16 @@ Sceptre provides environment-level commands. This one deletes the whole ``dev`` 
 Usage
 -----
 
-Sceptre can be used from the CLI, or imported as a Python package.
+Sceptre can be used directly from the command-line, via Docker, or imported as a Python package.
 
-CLI::
+Command-line
+^^^^^^^^^^^^
+.. code-block:: text
 
+  $ sceptre
   Usage: sceptre [OPTIONS] COMMAND [ARGS]...
+
+    Implements sceptre's CLI.
 
   Options:
   --version             Show the version and exit.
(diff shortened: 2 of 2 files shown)


akdor1154:
Readme docs are in-depth, which is obviously informative, but most of it is just general "how do set up docker so boto-based tools work inside it?" and is not Sceptre-specific at all - would this be better in a separate "How do I use Docker?" doc page?

rgitzel:
That's an excellent point... _really_ nearly all that info should be in the Docker Hub page.

Will give it a go...

rgitzel:
I've removed most of the content from the README in favour of putting it on the Docker Hub page.

That page doesn't exist yet, and isn't under source control.  For now I've put it under my hub:  https://hub.docker.com/r/rgitzel/sceptre/

I also fixed the Dockerfile to simplify the Docker parameters, and now`init` works!

ngfgrant:
@rgitzel Thanks! Can you fix the lint error and rebase off v1 branch 👍

rgitzel:
I think I've fixed the lint issue, but CI is taking awhile.

In the meantime, here's the source for the Dockerhub page I created:


-------------------------

Running Sceptre via Docker is useful in situations where you can't expect to or don't want to install Python and the required libraries.

## Using This Image
The simplest thing of course is to just run it without parameters.
```
$ docker run cloudreach/sceptre
Usage: sceptre [OPTIONS] COMMAND [ARGS]...

```
But in practice you'll want Sceptre to *do* something, so your Sceptre files will need to be accessible from within the container.  In particular, they are expected to be in the `/sceptre` folder, and you need to mount a local directory to that location.

One way of doing this is to run Docker from the local folder containing your files:
```
$ cd /my/sceptre/folder
$ docker run -t --rm \
      -v $(pwd):/sceptre:ro \
      cloudreach/sceptre COMMAND...

```

Another way is to specify the full path to the folder, and then run Docker from wherever you like:
```
$ docker run -t --rm \
      -v /my/sceptre/folder:/sceptre:ro \
      cloudreach/sceptre COMMAND...

```

To make changes in AWS, Sceptre will need credentials. One approach is to mount your local `.aws` folder to be used by the `root` user in the container:
```
$ docker run -t --rm \
      -v ${HOME}/.aws:/root/.aws:ro \
      -v $(pwd):/sceptre:ro \
      cloudreach/sceptre COMMAND...

```
(Note how the folders are mounted read-only.  Sceptre does not need to write anything to the host filesystem for most commands.)

Another approach uses the environment variables which have already somehow been defined in your shell:
```
$ docker run -t --rm \
      -e AWS_ACCESS_KEY_ID \
      -e AWS_SECRET_ACCESS_KEY \
      -e AWS_SESSION_TOKEN \
      -v $(pwd):/sceptre:ro \
      cloudreach/sceptre COMMAND...

```

A tool such as [aws-vault](https://github.com/99designs/aws-vault) can set these variables on the fly each time you run Docker, which is especially useful when you have MFA enabled on your AWS account:
```
$ aws-vault exec sandbox -- docker run -t --rm -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -e AWS_SESSION_TOKEN -v $(pwd):/sceptre:ro cloudreach/sceptre  COMMAND...

```
*(Apologies for the long line,  `aws-vault` commands can't be broken into multiple lines like shell commands.)*


Finally, to use interactive commands like `init` you must add the `-i` option so that Docker can accept input. Then mount your folder as _writable_ by leaving off the `:ro` suffix.  

For example, to create a new project in your current folder:
```
$ docker run -i -t --rm \
      -v $(pwd):/sceptre \
      cloudreach/sceptre init project my-new-project
Please enter a project_code [my-new-project]:
Please enter a region []: us-west-2
$ ls my-new-project/
config		templates

```
If this gives you strange behaviour, probably you forgot the `-i` option.

jamielennox:
Hey @ngfgrant pinging as you've got a review outstanding that I think has been cleared. 

I'd be really keen to see this merged and then auto-publishing to dockerhub. We drive sceptre through our CI/CD process and it'd be great to have an image that is usable and updated when a new version is released.

ngfgrant:
@rgitzel thanks for this PR. I am going to close in favour of #711  which will handled publishing to DockerHub automatically as part of build. The dockerfile in this pr is not too far off what #711 has in any case.

PR #481, opened 2018-10-16
Update cli.md: Remove the `diff` command.

gomlgs:
Docs: Remove the `diff` command.


1oglop1:
@niallgrantcloudreach there is no need for PRs like this since v2 will have cli generated from code directly

nabeelamjad:
Good to know, wasn't aware the documentation process was automated in v2.

1oglop1:
Yeah it's not very transparent yet as docs for V2 is a last task before we release, cuz they are dependent on all changes and it's only semi-automatic process to convert `.md` to `.rst`. 
Sphinx can go through the Click cli, which is effortless and all docs are pulled from code.

PR #482, opened 2018-10-17
Use Directed Acyclic Graph for Stack Configs

ngfgrant:
This PR changes the underlying data structure that holds stack 
dependencies for launching and delete a stack or stack group.

Previously, a dictionary was utilised with this commit the behaviour 
remains the same but a Directed Acyclic Graph is used to store 
dependencies instead.

* [x] Added unit tests.


sceptre/config/graph.py
```diff
@@ -0,0 +1,85 @@
...
+from sceptre.exceptions import CircularDependenciesError
+
+
+class StackDependencyGraph(object):
+    """
+    A Directed Graph representing the relationship between Stack config
+    dependencies. Responsible for initalizing the graph object based on
+    a given inital stack config path.
+
+    :param inital_config_path: The inital stack config you want to generate
```
dwhiteley-cr: This parameter documentation is out of date, it no longer takes a string `initial_config_path`

sceptre/config/graph.py
```diff
@@ -0,0 +1,85 @@
...
+    the graph from
+    :type inital_config_path: str
+    """
+
+    def __init__(self, dependency_map=None):
+        self.logger = logging.getLogger(__name__)
+        self.graph = nx.DiGraph()
+        if dependency_map is None:
+            dependency_map = {}
```
ewencochran-cr: Why not set `dependency_map={}` in the function definition rather than `dependency_map=None`? Then these two lines wouldn't be needed.

sceptre/config/reader.py
```diff
@@ -163,14 +180,22 @@ def read(self, rel_path, base_config=None):
...
             raise ConfigFileNotFoundError(
                 "Config file \"{0}\" not found.".format(rel_path)
             )
 
         # Parse and read in the config files.
-        config.update(self._recursive_read(directory_path, filename))
+        this_config = self._recursive_read(directory_path, filename)
+
+        if "dependencies" in config or "dependencies" in this_config:
+            this_config['dependencies'] = strategies.list_join(
```
ewencochran-cr: Should this use the strategy for `dependencies` that is defined in `CONFIG_MERGE_STRATEGIES` rather than always doing `list_join`?

sceptre/config/strategies.py
```diff
@@ -0,0 +1,33 @@
...
+
+    if b is not None:
+        return a + b
+
+    return a
+
+
+def dict_merge(a, b):
+    if a and not isinstance(a, dict):
+        raise TypeError('{} is not a list'.format(a))
```
ewencochran-cr: Should be `{} is not a dict`

sceptre/config/strategies.py
```diff
@@ -0,0 +1,33 @@
...
+    if b and not isinstance(b, dict):
+        raise TypeError('{} is not a list'.format(b))
```
ewencochran-cr: Should be `{} is not a dict`

sceptre/helpers.py
```diff
@@ -41,14 +45,14 @@ def recurse_into_sub_stack_groups(func):
     @wraps(func)
     def decorated(self, *args, **kwargs):
         function_name = func.__name__
-        responses = {}
+        responses = factory()
         num_stack_groups = len(self.sub_stack_groups)
-
+        print("sub stack groups: " + str(num_stack_groups))
```
ewencochran-cr: Wild print statement? Is it a debug one or should it be logged?

sceptre/config/graph.py
dwhiteley-cr: This is the whole https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments thingy

sceptre/helpers.py
ngfgrant: Correct, completely wild. 😟 must have made it through an earlier fixup I did. Will get it removed.

sceptre/config/graph.py
ewencochran-cr: Aha, makes sense.

sceptre/config/reader.py
ngfgrant: yes good spot

ngfgrant:
@ewencochran-cr and @dwhiteley-cr  Comments addressed.

PR #504, opened 2018-11-19
Resolve #503, Update setup.py and unlock version of dependencies

### LINKED ISSUE #503: Sceptre v2-rc missin internal sub packages during installation.
In Sceptre v2-rc `setup.py` packages are updated manually and does not support newly created/refactored sub-packages.

https://github.com/cloudreach/sceptre/blob/v2-rc/setup.py#L45

For example sub-package `sceptre.config` is not installed now.

1oglop1:
Resolve #503
* use find_packages instead of manual specification
* Update click to 7.0 < 8.0
* Update packaging to 18.0 < 19.0
* Unlock colorama <1.0.0
* Unlock six <2.0.0

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 
 from sceptre import __version__
-from setuptools import setup
+from setuptools import setup, find_packages
 
 with open("README.rst") as readme_file:
     readme = readme_file.read()
@@ -12,7 +12,7 @@ with open("HISTORY.rst") as history_file:
 
 install_requirements = [
     "boto3>=1.3,<2.0",
-    "click==6.7",
+    "click>=6.7,<7.0",
     "PyYaml==3.13",
     "Jinja2>=2.8,<3",
     "packaging==16.8",
@@ -43,18 +43,7 @@ setup(
     author_email="sceptre@cloudreach.com",
     license='Apache2',
     url="https://github.com/cloudreach/sceptre",
-    packages=[
-        "sceptre",
-        "sceptre/resolvers",
-        "sceptre/hooks",
-        "sceptre/cli",
-        "sceptre/config",
-        "sceptre/stack_policies",
-        "sceptre/plan"
-    ],
-    package_dir={
-        "sceptre": "sceptre"
-    },
+    packages=find_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests"]),
     py_modules=["sceptre"],
     entry_points={
         "console_scripts": [


setup.py
```diff
@@ -12,12 +12,13 @@
...
 install_requirements = [
     "boto3>=1.3,<2.0",
-    "click==6.7",
+    "click>=7.0<8.0",
     "PyYaml==3.13",
     "Jinja2>=2.8,<3",
-    "packaging==16.8",
-    "colorama==0.3.7",
-    "six==1.11.0"
+    "packaging>=18.0<19.0",
```
ngfgrant: I am not sure what the consequences of upgrading this package by 2 major version releases will be. Prefer to keep `>=16.0,<17.0` for now.

setup.py
```diff
@@ -12,12 +12,13 @@
 
 install_requirements = [
     "boto3>=1.3,<2.0",
-    "click==6.7",
+    "click>=7.0<8.0",
```
ngfgrant: I am not sure what the consequences of upgrading this package by 2 major version releases will be. Prefer to keep `>=6.7,<7.0`. Until we can do further test

setup.py
```diff
@@ -12,12 +12,13 @@
...
     "PyYaml==3.13",
     "Jinja2>=2.8,<3",
-    "packaging==16.8",
-    "colorama==0.3.7",
-    "six==1.11.0"
+    "packaging>=18.0<19.0",
+    "colorama>=0.3.7<1.0.0",
```
ngfgrant: Pre-stable packaging. Would definitely like to get this dependency to a stable release i.e. `1.0`. But for now a lot could have changed between `0.3.7` and `1.0` so I want to keep this pinned to `0.3.7` until we get v2 out and feedback from that has settled.

setup.py
1oglop1: then it will still conflict with awscli and other libs

setup.py
1oglop1: Okay this is fair point. 
So could be updated to latest 0.4.0 ?

ngfgrant: Again, it is a pre-stable release so there are no guarantees about API compatibility between versions. Indeed looks like they have dropped Python 2.6 support in 0.4.0 which is something we still support. Happy for this to be pinned to `0.3.9` until we drop python 2 support.

setup.py
ngfgrant: `17.0` drops support for python 2.6 which we still support (according to our classifiers)

PR #508, opened 2018-11-20
Update project docs

ngfgrant:
Updates the project docs and converts most of the project does to Markdown. 

Amends setup.py and requirements.txt to take account of PyPI added support of Markdown.


README.md
```diff
@@ -0,0 +1,219 @@
...
+- LogicalResourceId: Subnet
+  PhysicalResourceId: subnet-445e6e32
+  dev/vpc:
+- LogicalResourceId: VirtualPrivateCloud
+  PhysicalResourceId: vpc-c4715da0
+```
+
+Sceptre provides Stack Group level commands. This one deletes the whole `dev`
+Stack Group. The subnet exists within the vpc, so it must be deleted first.
+Sceptre handles this automatically::
```
nabeelamjad: I think we should add additional information over the following behavior:

```bash
$ tree
.
├── config
│   └── dev
│        ├── config.yaml
│        ├── subnets.yaml
│        └── vpc.yaml
└── templates
    ├── subnets.py
    └── vpc.py
```

A user can delete the Subnets stack using `sceptre delete dev/subnets.yaml`. An important thing to note is that doing so will delete **all** stacks the Subnet is dependent on, this is recursive until there are no more dependent stacks.

Example: The Subnet stack is dependent on the VPC stack, which is dependent on an infrastructure stack, which is finally dependent on an environment stack.

Deleting the Subnet stack will cause the VPC, infrastructure and environment stacks to be deleted.

ngfgrant: Agreed - I will wait until we have a definitive delete behaviour and then add a note about deletion behaviour.

requirements.txt
```diff
@@ -16,11 +16,12 @@ pygments==2.2.0
 pytest-runner>=3.0.0,<3.1.0
 pytest>=3.2.0,<3.3.0
 PyYaml==3.13
-readme-renderer==17.2
-setuptools==37.0.0
+readme-renderer>=24.0
```
connec: We should maybe stick with `==` (or add an upper bound) here to avoid pulling in an incompatible dependency in the future.

requirements.txt
```diff
@@ -16,11 +16,12 @@ pygments==2.2.0
...
+setuptools>=40.6.2
```
connec: As above.

requirements.txt
```diff
@@ -16,11 +16,12 @@ pygments==2.2.0
...
 six==1.11.0
 Sphinx==1.6.5
-tox==2.9.1
+tox>=2.9.1,<3.0.0
 troposphere>=2.0.0,<2.1.0
+twine>=1.12.1
```
connec: As above.

PR #509, opened 2018-11-21
V2 Release Candidate

ngfgrant:
This release candidate brings with it greater stability and reliability of stack execution where there are complex dependencies. The underlying codebase has also been improved making the project easier to work with. There are now clearer boundaries and concepts that exist such as SceptreContext and SceptrePlan. 

A number of changes to to the codebase including:

# Removed

## Environment

An Environment is now referred to as a StackGroup. You can still structure your project as you have in v1 but under the hood Sceptre doesn't actually create or store a StackGroup but it still respects cascading config etc. Instead a StackGraph is used to store Stacks and their relationships.

# Changed 

## Stack

A Stack used to hold both the data and the functionality within the same class. In v2 a Stack is now a simple data class and the actions, such as launch(), delete(), update() have been moved into StackActions

## CLI 
The CLI commands have changed slightly to avoid having to specify whether the path you give is either a environment (now removed) or a stack. You can now state your intention and provide either a StackGroup or a individual Stack for example, `sceptre launch prod/us/database` will launch the `prod/us/database` StackGroup and `sceptre launch prod/us/database/master.yaml` will launch the `master` stack in `prod/us/database` StackGroup.

# Added

## Sceptre Context

SceptreContext is a place where we hold information that is relevant to the project, including references to the project paths such as the path to your Sceptre project, templates path, config path, and the default names for your configuration files.

If you are adding functionality to the CLI or you want to use Sceptre as a module you will need to create a SceptreContext and pass it an absolute path to your Sceptre project directory and the "command_path" - the path you are going to execute commands on.

## Sceptre Plan

A SceptrePlan takes a SceptreContext and constructs everything you will need to execute a command against your project. After creating a SceptrePlan you will be able to call all available methods for your stacks and access the results from each stack. 

## SceptrePlanExecutor

You won't be able to interact with the SceptrePlanExecutor directly but this part of the code is responsible for taking a SceptrePlan and ensuring all commands on every stack, are executed in the correct order, concurrently. The executor algorithm focuses on correctness over maximal concurrency. It know what to execute and when based on a StackGraph which is created when a SceptrePlan is created.

## StackGraph

A StackGraph is a Directed Acyclic Graph where the nodes hold Stack objects and the edges represent a "depends on" relationship. The graph is created when a SceptrePlan is instantiated and is then used during the execution phase to ensure that stack dependency relationships are complete and correct. Previously, in v1 stack dependencies were available between Environments and even within Environments they could be temperamental, this concept resolves this issue.

## StackActions

The StackActions class takes a Stack object and uses the data held on the Stack when calling AWS. StackActions is used by the SceptrePlanExecutor. To add new functionality you can add a method to StackActions and it will become available to all Stacks in Sceptre.


sceptre/config/graph.py
```diff
@@ -2,87 +2,113 @@
...
+        all adjecent edges that represent a 'depends on' relationship with
+        other Stacks.
         """
-        for stack, dependencies in self.dependency_map.items():
-            self._generate_edges(stack, dependencies)
+        return self.graph.remove_node(stack)
 
-    def _is_acyclic(self):
-        return nx.is_directed_acyclic_graph(self.graph)
+    def write(self):
```
ngfgrant: remove `write()`. Was useful for debugging but not suitable for release in it's current form. Will also break as `pydot` is not listed in requirements.

sceptre/config/graph.py
```diff
@@ -2,87 +2,113 @@
...
             self.logger.debug("Added edge: {}".format(edge))
 
-        if not dependency_paths:
-            self.graph.add_node(stack_path)
-
-        self.graph.remove_edges_from(nx.selfloop_edges(self.graph))
+        if not dependencies:
+            self.graph.add_node(stack)
 
     def reverse_graph(self):
```
ngfgrant: I think we can remove `reverse_graph()` as it does not appear to be called anywhere?

sceptre/config/reader.py
```diff
@@ -154,10 +147,64 @@ def class_constructor(loader, node):
...
+        and a final set of Stacks is returned.
+
+        :returns: A set of Stacks.
+        :rtype: set
+        """
+        stack_map = {}
+
+        root = self.context.full_command_path()
+        if path.isfile(root):
+            todo = {root}
```
ngfgrant: were we going to think of a better variable name for `todo`?

sceptre/config/reader.py
```diff
@@ -154,10 +147,64 @@ def class_constructor(loader, node):
...
+            todo = set()
+            for directory_name, sub_directories, files in walk(root):
+                for filename in fnmatch.filter(files, '*.yaml'):
+                    if filename.startswith('config'):
+                        continue
+
+                    todo.add(path.join(directory_name, filename))
+
+        stack_group_configs = {}
+        done = set()
```
ngfgrant: Were we going to think of a better variable name than done? `stacks_visited`?

PR #511, opened 2018-11-21
Delete launch order

connec:
When deleting a stack group we want to delete any stacks in the group,
but also any stacks that depend on those stacks. In order to identify
these stacks properly the configuration for all stacks is now loaded
into the StackGraph, and the launch order is computed by finding all the
stacks reachable from the stacks targetted by the given command path.

For deletion, this search is performed on the reversed graph, meaning
anything that depends on the stacks to be deleted will also be deleted,
as desired.

-----------------

* [ ] Added unit tests.
* [ ] The new code doesn't generate flake8 (`make lint`) offenses.

### DIFF
--- a/integration-tests/steps/change_sets.py
+++ b/integration-tests/steps/change_sets.py
@@ -139,14 +139,14 @@ def step_impl(context, change_set_name, stack_name):
     allowed_errors = {'ValidationError', 'ChangeSetNotFound'}
 
     try:
-        sceptre_plan.describe_change_set(change_set_name)
+        responses = sceptre_plan.describe_change_set(change_set_name)
     except ClientError as e:
         if e.response['Error']['Code'] in allowed_errors:
             context.error = e
             return
         else:
             raise e
-    context.output = sceptre_plan.responses
+    context.output = responses
 
 
 @then('stack "{stack_name}" has change set "{change_set_name}" in "{state}" state')
@@ -196,10 +196,10 @@ def step_impl(context, change_set_name, stack_name):
     )
 
     del response["ResponseMetadata"]
-    for output in context.output:
+    for stack, output in context.output.items():
         del output["ResponseMetadata"]
 
-    for output in context.output:
+    for stack, output in context.output.items():
         assert response == output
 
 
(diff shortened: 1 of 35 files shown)


sceptre/config/reader.py
```diff
@@ -158,48 +158,40 @@ def construct_stacks(self):
...
-                for filename in fnmatch.filter(files, '*.yaml'):
-                    if filename.startswith('config'):
-                        continue
+        root = self.context.full_config_path()
+        command_path = self.context.full_command_path()
 
-                    todo.add(path.join(directory_name, filename))
+        for directory_name, sub_directories, files in walk(root):
+            for filename in fnmatch.filter(files, '*.yaml'):
+                if filename.startswith('config.'):
```
ngfgrant: is this a erroneous `.` after `config?

connec: No, I figured I'd make sure it was `config.<ext>` rather than potentially accidentally catching `configblah.yaml`, which could be a valid stack name 😄 

It's not much better really, we probably should filter for `config.ya?ml` or something like that but I didn't want to go that far.

integration-tests/steps/change_sets.py
```diff
@@ -196,10 +196,10 @@ def step_impl(context, change_set_name, stack_name):
     )
 
     del response["ResponseMetadata"]
-    for output in context.output:
+    for stack, output in context.output.items():
```
ngfgrant: will `.items()` have issues in Python2 in this case?

connec: It will work, but be slightly less performant than in Python 3 (for Py2 we should really use `iteritems` for that behaviour). Functionally it is compatible, though, and we use `items` a lot now.

nabeelamjad:
👍

integration-tests/steps/change_sets.py
connec: I suspect that any performance difference here will be completely dwarfed by the network calls to AWS, anyway.

nabeelamjad: Yeah, we use `items` a lot, and I guess I could've thrown away the response for `stack` and used `.values()` instead.

sceptre/config/reader.py
ngfgrant: 👍

PR #534, opened 2018-12-03
Fix accessing stack group

ngfgrant:
In v1 `environment_path` was accessible as a list in templates,
split by the path separator.

In v2 the `environment_path` is known as the `command_path`. This
commit adds support for accessing the `command_path` back in to
Sceptre. For example, the user can now access parts of the
`command_path` in templates like {{ command_path.0 }}.

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/config/reader.py
+++ b/sceptre/config/reader.py
@@ -330,7 +330,9 @@ class ConfigReader(object):
             template = jinja_env.get_template(basename)
             self.templating_vars.update(stack_group_config)
             rendered_template = template.render(
-                self.templating_vars, environment_variable=environ
+                self.templating_vars,
+                command_path=self.context.command_path.split("/"),
+                environment_variable=environ
             )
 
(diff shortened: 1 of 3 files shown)


sceptre/config/reader.py
```diff
@@ -330,7 +330,9 @@ def _render(self, directory_path, basename, stack_group_config):
             template = jinja_env.get_template(basename)
             self.templating_vars.update(stack_group_config)
             rendered_template = template.render(
-                self.templating_vars, environment_variable=environ
+                self.templating_vars,
+                stack_group_path=self.context.command_path.split(sep),
```
connec: Should we just call this `command_path`? It could be slightly misleading otherwise since the path could be a 'stack group' or a 'stack'.

ngfgrant: Yes, good idea. Will make the change.

PR #535, opened 2018-12-03
[Resolves #531] Normalise os paths

### LINKED ISSUE #531: Normalise OS paths v2
Running sceptre generate in v2.0 is giving the following error: `ValueRrror: max() arg is an empty sequence`.  Any ideas on this?  I'm just doing a barebones with one template.
image

ngfgrant:
The intention is to avoid unnecessary path manipulation between the end user and executing a command. In v1 [`_validate_path`](https://github.com/cloudreach/sceptre/blob/v1/sceptre/environment.py#L75) was responsible for this. Users of any operating system should now be able pass paths with native separators and sceptre work as intended.

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/helpers.py
+++ b/sceptre/helpers.py
@@ -1,4 +1,7 @@
 # -*- coding: utf-8 -*-
+from os import sep
+
+from sceptre.exceptions import PathConversionError
 
 
 def get_external_stack_name(project_code, stack_name):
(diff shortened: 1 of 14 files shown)


sceptre/helpers.py
```diff
@@ -62,3 +64,23 @@ def func_on_instance(key):
         for index, value in enumerate(attr):
             func_on_instance(index)
     return attr
+
+
+def normalise_path(path):
+    """
+    Converts backslashes to forward slashes.
+    Raises an PathConversionError if the path has a leading or
```
connec: This won't error for leading slashes.

sceptre/helpers.py
```diff
@@ -62,3 +64,23 @@ def func_on_instance(key):
...
+    trailing slash.
+    :param path: A directory path
+    :type path: str
+    :raises: sceptre.exceptions.PathConversionError
+    :returns: A normalised path with forward slashes.
+    :returns: string
+    """
+    path = path.replace('\\', '/')
+    if path.endswith("/"):
```
connec: Just checking - why do we want to throw for trailing slashes? I think this could be quite annoying for people using `--dir` and tab completion to fill in the `project_path`.

tests/test_helpers.py
```diff
@@ -1,10 +1,46 @@
...
 
 class TestHelpers(object):
 
     def test_get_external_stack_name(self):
         result = get_external_stack_name("prj", "dev/ew1/jump-host")
         assert result == "prj-dev-ew1-jump-host"
+
+    def test_normalise_path_with_valid_path(self):
+        path = normalise_path("valid/path")
+        assert path == "valid/path"
```
connec: These assertions will fail on Windows (I assume - I don't have a Windows environment to test in). It would be OK I think to assert on `path.join`:

```py
assert path == join("valid", "path")
```

sceptre/helpers.py
ngfgrant: My thinking here is that we use `path.split()` through out the codebase and if there is a trailing slash I think split would return an empty string as the "head" rather than the last part of the path. E.g. 

```python
from os.path import split

with_slash = "this/is/a/path/" 
without_slash = "this/is/a/path"

split(with_slash) # returns ('this/is/a/path', '')
split(without_slash) # returns ('this/is/a', 'path')
```

I am not sure that sceptre has ever supported a trailing slash in stack/stackgroup names.

PR #585, opened 2019-01-07
add missing output_format support to generate

pha6d:
prior this fix generate action output_format default to "str" in v2.


ngfgrant:
Hey @pha6d,

Thank you so much for submitting the PR. It turns out that great minds think alike and I've also just fixed this as part of a slightly bigger PR 🤦‍♂️ - would normally accept but in this case I am going to close this in favour of https://github.com/cloudreach/sceptre/pull/587

Apologies and thanks again,
Niall

PR #626, opened 2019-02-06
[Resolve #558] Fix RecursionError

### LINKED ISSUE #558: Fix sceptre_user_data with !stack_output
In v2.0 When referencing `sceptre_user_data` with a `stack_output` resolver an error occurs where the resolvable property is read Sceptre enters into a recursion and eventually returns a `RecursionError: maximum recursion depth exceeded`.

zaro0508:
Getting the connection manager from template object results in
a RecursionError.  Getting it from the stack object does not.


zaro0508:
Sorry, I forgot to run behave tests.  This naive fix doesn't work when launching a config group (multiple stacks).

PR #629, opened 2019-02-09
[Resolve #623] Catch missing config attributes

### LINKED ISSUE #623: Improve exception handling for KeyErrors
# CONTEXT

If a required key is missing from a stack or config file and Sceptre attempts to load the config or stack the user is presented with a KeyError.

```python
  File "/Users/niallgrant/.pyenv/versions/sceptre-3.6.2/bin/sceptre", line 11, in <module>
    load_entry_point('sceptre', 'console_scripts', 'sceptre')()
  File "/Users/niallgrant/.pyenv/versions/3.6.2/envs/sceptre-3.6.2/lib/python3.6/site-packages/click/core.py", line 722, in __call__
  File "/Users/niallgrant/.pyenv/versions/3.6.2/envs/sceptre-3.6.2/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Users/niallgrant/.pyenv/versions/3.6.2/envs/sceptre-3.6.2/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
  File "/Users/niallgrant/.pyenv/versions/3.6.2/envs/sceptre-3.6.2/lib/python3.6/site-packages/click/core.py", line 895, in invoke
  File "/Users/niallgrant/.pyenv/versions/3.6.2/envs/sceptre-3.6.2/lib/python3.6/site-packages/click/core.py", line 535, in invoke
  File "/Users/niallgrant/.pyenv/versions/3.6.2/envs/sceptre-3.6.2/lib/python3.6/site-packages/click/decorators.py", line 17, in new_func
  File "/Users/niallgrant/Codebase/sceptre/sceptre/sceptre/cli/helpers.py", line 37, in decorated
    return func(*args, **kwargs)
  File "/Users/niallgrant/Codebase/sceptre/sceptre/sceptre/cli/template.py", line 66, in generate_command
    plan = SceptrePlan(context)
  File "/Users/niallgrant/Codebase/sceptre/sceptre/sceptre/plan/plan.py", line 31, in __init__
  File "/Users/niallgrant/Codebase/sceptre/sceptre/sceptre/config/reader.py", line 203, in construct_stacks
    stack = self._construct_stack(rel_path, stack_group_config)
  File "/Users/niallgrant/Codebase/sceptre/sceptre/sceptre/config/reader.py", line 429, in _construct_stack
    config["template_path"]
KeyError: 'template_path'
```

Required keys are detailed in the dicts: https://github.com/cloudreach/sceptre/blob/master/sceptre/config/reader.py#L54

and 

https://github.com/cloudreach/sceptre/blob/master/sceptre/config/reader.py#L66

# DONE WHEN

Instead of a KeyError the user should see a nice error message saying they need to include the required field. Bonus points for saying which stack failed/which file they need to amend.

devenney:
Signed-off-by: Brendan Devenney <brendan.devenney@cloudreach.com>

There are some unanswered questions to consider when reviewing this PR:

* Should the code be differentiating between Stack Config and Stack Group Config for the user's benefit? I'm a little thrown off by the fact you can successfully deploy a Stack without any files in the tree named `config.yaml`, despite the implication that `config.yaml` is mandatory in the [StackGroup Config](https://sceptre.cloudreach.com/latest/docs/stack_group_config.html) docs. Effectively, you can store all of your config in `vpc.yaml` and Sceptre won't complain.
* Do we care enough about invalid keys to add these checks? My opinion is yes, as a typo in a stack's config could be masked due to the attribute cascading down from a parent directory.

Also, for what it's worth, I have no idea if this is the most efficient way to flatten lists of lists into a single flat list. Some rough testing suggests it's the fasted way I know but please do consider if there's a better/cleaner way. 👍 

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/config/reader.py
+++ b/sceptre/config/reader.py
@@ -20,6 +20,7 @@ from packaging.specifiers import SpecifierSet
 from packaging.version import Version
 
 from sceptre import __version__
+from sceptre.exceptions import InvalidConfigFileError
 from sceptre.exceptions import InvalidSceptreDirectoryError
 from sceptre.exceptions import VersionIncompatibleError
 from sceptre.exceptions import ConfigFileNotFoundError
@@ -79,6 +80,19 @@ STACK_CONFIG_ATTRIBUTES = ConfigAttributes(
         "stack_tags",
         "stack_timeout"
     }
+)
+
+INTERNAL_CONFIG_ATTRIBUTES = ConfigAttributes(
+    {
+        "project_path",
+        "stack_group_path",
+    },
+    {
+    }
+)
+
+REQUIRED_KEYS = STACK_GROUP_CONFIG_ATTRIBUTES.required.union(
+    STACK_CONFIG_ATTRIBUTES.required
 )
 
 
@@ -162,7 +176,7 @@ class ConfigReader(object):
         """
         Traverses the files under the command path.
         For each file encountered, a Stack is constructed
-        using the correct config. Depependencies are traversed
+        using the correct config. Dependencies are traversed
         and a final set of Stacks is returned.
 
         :returns: A set of Stacks.
@@ -424,6 +438,16 @@ class ConfigReader(object):
         parsed_stack_group_config = self._parsed_stack_group_config(stack_group_config)
         config = self.read(rel_path, stack_group_config)
         stack_name = path.splitext(rel_path)[0]
+
+        # Check for missing mandatory attributes
+        for required_key in REQUIRED_KEYS:
+            if required_key not in config:
+                raise InvalidConfigFileError(
+                    "Required attribute '{0}' not found in configuration of '{1}'.".format(
+                        required_key, stack_name
+                    )
+                )
+
         abs_template_path = path.join(
             self.context.project_path, self.context.templates_path,
             config["template_path"]
--- a/.gitignore
+++ b/.gitignore
@@ -117,3 +117,22 @@ _site/
 /docs/_api/sceptre.rst
 /docs/_api/sceptre.hooks.rst
 /docs/_api/sceptre.resolvers.rst
+
+### Vim ###
+# Swap
+[._]*.s[a-v][a-z]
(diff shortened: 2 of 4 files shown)


sceptre/config/reader.py
```diff
@@ -424,6 +439,29 @@ def _construct_stack(self, rel_path, stack_group_config=None):
...
+                raise InvalidConfigFileError(
+                    "Required key '{0}' not found in Stack Group Config of '{1}'.".format(
+                        required_key, stack_name
+                    )
+                )
+
+        for required_key in STACK_CONFIG_ATTRIBUTES.required:
+            if required_key not in config:
+                raise InvalidConfigFileError(
+                    "Required key '{0}' not found in Stack Config ({1})".format(
```
nabeelamjad: Probably want to keep the error message format consistent with the one above.

If we don't care about the _exact_ error message (Stack Config / Stack Group Config) then we should probably use a single loop to check both required config attributes.

I believe you mentioned that the **stack group config** can be missing with everything in the **stack config** instead, which means that the error message above (Stack Config / Stack Group Config) won't be correct anyhow. We could fix the underlying issue to force the correct attributes to exist in the `config.yaml` but that would definitely be a breaking change at this point, so I'd rather suggest we change the error message instead.

devenney:
Thanks for the review @nabeelamjad. tl;dr of my latest push:

* I have removed the invalid key testing. Clearly this issue isn't as clear-cut as I had assumed and I feel it warrants its own issue rather than flying under the radar with this fix. I have left the `INTERNAL_CONFIG_ATTRIBUTES` code though as this was key to me understanding the codebase when determining valid keys.
* The check no longer differentiates between Stack Config and StackGroup Config as discussed.
* I've added a test for each mandatory key. I can take a more holistic approach with this to ensure it automatically picks up added/removed required keys if you want but it feels overkill.
* I removed the vim swap file and added vim regexes to `.gitignore`.

PR #654, opened 2019-03-01
Add Python 3.7 Support

ngfgrant:
The previous PR was not triggering a build. This PR should. It is failing on some dependency things so need to look at circleci docker image and probably tox setup.


devenney:
I can't push to the source branch but I have debugged as discussed. Seems trivial - an issue in moto rather than an issue in boto. PR raised (#657) which targets this branch and should resolve the CI issues. 👍

Issue #152, opened 2017-06-24
Add project level configuration file


jamesroutley:
For tool-level config. To allow specification of the tools behaviour at per-machine level.
Config may be stored at ~/.sceptre/config.yaml
Example config could include:
- Default output for Sceptre - i.e. colour, no-colour, verbose
- Disable requirement -y flag for commands
- Default AWS profile to use
- Specify default location to search for external templates.

zaro0508:
we already have a place for project level configs, it's in `config/config.yaml`

PR #673, opened 2019-04-09
Fixes #672 - Set output format at config.yaml Level 

### LINKED ISSUE #672: Set output format at config.yaml Level
Coming from using Sceptre v1 a lot to 2.0 I have noticed the move towards favouring text output as the default output_format rather than JSON as it was before. 

It took me a little while to figure out that a lot of my commands looked broken as I was not setting an `--ouput` flag on my commands to JSON (or YAML) and since getting used to it, It has occurred to me that at a project level and even at a Stack level, it would be really useful to be able to set this once and not have to worry about setting a flag each time.

I would like to be able to set this using something like `output` or `output_format` at my `config.yaml` files and then utilise this as my preferred option over setting --output.

Cascading config should still work and maybe if --ouput is set, it overrides any value in the config.

nmyster:
Coming from using Sceptre v1 a lot to 2.0 I have noticed the move towards favouring text output as the default output_format rather than JSON as it was before.

It took me a little while to figure out that a lot of my commands looked broken as I was not setting an --ouput flag on my commands to JSON (or YAML) and since getting used to it, It has occurred to me that at a project level and even at a Stack level, it would be really useful to be able to set this once and not have to worry about setting a flag each time.

I would like to be able to set this using something like output or output_format at my `config.yam` files and then utilise this as my preferred option over setting `--output`.

### Example
```yaml
region: eu-west-1
project_code: some-project
template_bucket: some-bucket
template_prefix: some-prefix
output: yaml 
or
output: json
```

Cascading config should still work and maybe if `--ouput` is set, it overrides any value in the config.
## PR Checklist

## Other Information


nmyster:
Looking for some help to add appropriate testing

nmyster:
Commit currently utilises similar code to `s3_details` to get the output value and overrides the context.output_format if it is provided

nmyster:
@ngfgrant  looking for some input on this PR and some assistance getting appropriate testing in place. 
tl;dr - I'd like to set the ouput (json or yaml) in the config.yaml - this PR allows that but it may be a bit hacky.

Have some ideas in general to make the config.yaml more capable so future values may be good to parse - is this a good time to think about how to refactor parts of the code to make this easier in the future

Example future values I'd like to add: profile mappings so that when a profile is provided to different parts of the stacks, it can read the profile mapping to understand what role to assume when run in a CICD environment (where profiles don't exist in the same fashion as running locally)

ngfgrant:
Thanks @nmyster will try get a look soon. Main focus for me is to get 2.1.1 out.

ngfgrant:
Hey @nmyster 

Based on your feedback here, from #709 and in v2.1.4 onward the default output on the CLI will be JSON.

Had a look at this and think it could probably be achieved in a more straightforward way. If we still wanted to set this in the config, which I am not 100% sure we want to do. I think probably this better belongs in a project-wide config file e.g. `sceptre.yaml`. Basically anything that is in `config.yaml` should probably relate to the values of the stack being deployed to CloudFormation.

If we close this one out we could look at project wide config which is definitely something that would be worth doing?

ngfgrant:
Knew we had something regarding project-wide settings https://github.com/cloudreach/sceptre/issues/152

ngfgrant:
Going to close this out in favour of #152 but will take this functionality into any future solution. Thanks again @nmyster

Issue #683, opened 2019-04-16
Enhancement: Show diff (and change set details) before updating a stack


jfhamlin:
# Current Behavior

Today, when a user runs a command to update a stack, they see a prompt like this:

```
$ sceptre update prod/vpc
Do you want to update 'prod/vpc' [y/N]:
```

While the user's goal might be to get `prod/vpc` into some new state, this isn't really enough information for her to confidently answer in the affirmative. There are many reasons the answer could be "N," even when trying to make a change:

1. The user hasn't checked out the latest version of the repo containing the sceptre project, and updating will actually revert older changes rather than apply new ones.
2. The user has the latest changes but forgot to add some resource or parameter.
3. The update is destructive or will cause the replacement of some resource, and the user wants to announce or further plan around such a change.
4. ...

# Desired Behavior

For every stack being updated (either a single stack or each stack in a stack group), show the diff of the template, the stack parameters, and if change sets are used, the change set details, before showing a stack update prompt.

For example:
```
$ sceptre update prod/vpc
Changes for 'prod/vpc/vpc.yaml':
Template:
  @@ -27,6 +27,8 @@
         Tags:
         - Key: Name
           Value: !Ref Name
  +      - Key: Test
  +        Value: '123'
     InternetGateway:
       Type: AWS::EC2::InternetGateway
       Properties:

Parameters:
  @@ -2,5 +2,5 @@
   Name: VPC
  -Tenancy: default
  +Tenancy: dedicated

Change Set:
Action Resource                   Replacement?
==============================================
Modify AttachGateway              False
Modify BastionSecurityGroup       Conditional
Modify NATSecurityGroup           Conditional
Modify PrivateSubnetSecurityGroup Conditional
Modify PublicRoute                Conditional
Modify PublicRouteTable           Conditional
Modify VPC                        Conditional

Do you want to update 'prod/vpc' with these changes? [y/N]:
```

If the general proposal sounds reasonable, we can discuss implementation details and any edge/corner cases we should consider.

jfhamlin:
@ngfgrant, let me know your thoughts on this! I'm turning some of the points in https://github.com/cloudreach/sceptre/issues/520#issuecomment-480911084 into enhancement proposals.

ngfgrant:
Hey @jfhamlin

I think this is currently possible with:

`sceptre create`
`sceptre describe` and 
`sceptre execute`

I.e. 

```shell 
$ sceptre create bundle/a/x a-change-set

[...]

PR #685, opened 2019-04-18
Resolve #658: Generate non-trivial json and yaml

### LINKED ISSUE #658: Generate Command
Tried to run:
```
sceptre --output yaml generate ap-southeast-1/waf.yml
```

Fails to write the generated template (no error message).

Without the output statement it writes a strange array of resources which are escaped.

worr:
In cases where Cloudformation-specific tags were used (like !GetAtt, for
example), sceptre would fail to generate yaml and json output when `sceptre
generate` was called.

Here we use the loaders provided by Amazon's cfn-lint package to properly read
the cloudformation yaml after substitution, so that we can translate it to yaml
and json correctly without error.

A potential issue with this approach is that I used cfn-lint as a library to provide functionality for resolving all of the custom tags in cloudformation. There aren't any guarantees that I'm aware of for the stability of this API, but it will work at the pinned version. In service of this particular problem, I opened up an issue [here](https://github.com/aws-cloudformation/cfn-python-lint/issues/831) about this, since having a cfn encoder/decoder as a library would be super useful for more than just sceptre.

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/cli/helpers.py
+++ b/sceptre/cli/helpers.py
@@ -1,9 +1,10 @@
 import logging
 import sys
-from functools import wraps
+from functools import partial, wraps
 
 import json
 import click
+import six
 import yaml
 
 from boto3.exceptions import Boto3Error
@@ -92,13 +93,13 @@ def _generate_json(stream):
                 if isinstance(item, dict):
                     items.append(item)
                 else:
-                    items.append(json.loads(item))
+                    items.append(yaml.load(item, Loader=CfnYamlLoader))
             except Exception:
                 print("An error occured writing the JSON object.")
         return encoder.encode(items)
     else:
         try:
-            return encoder.encode(json.loads(stream))
+            return encoder.encode(yaml.load(stream, Loader=CfnYamlLoader))
         except Exception:
             return encoder.encode(stream)
 
@@ -115,14 +116,14 @@ def _generate_yaml(stream):
                 else:
                     items.append(
                         yaml.safe_dump(
-                            yaml.load(item, Loader=yaml.FullLoader),
+                            yaml.load(item, Loader=CfnYamlLoader),
                             default_flow_style=False, explicit_start=True
                         )
                     )
             except Exception:
                 print("An error occured whilst writing the YAML object.")
         return yaml.safe_dump(
-            [yaml.load(item, Loader=yaml.FullLoader) for item in items],
+            [yaml.load(item, Loader=CfnYamlLoader) for item in items],
             default_flow_style=False, explicit_start=True
         )
     else:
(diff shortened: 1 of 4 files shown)


ngfgrant:
I have two lines of thinking on this at the moment.

First, we _could_ utilise `SafeLoader` in pyyaml and implement the required custom yaml constructors to parse `!...` which would generate the template even with things like `!Ref`. This would resolve the generate issue and not introduce a new dependency. We'd also be able to quickly add any missing yaml constructors as appropriate. It is a naive way of doing things but would generate a template.

An example of what I mean is described here: https://stackoverflow.com/questions/41089065/loading-cloudformation-yaml-using-python and there is a project that looks to use this approach: 
https://github.com/awslabs/aws-cfn-template-flip.

The `sceptre generate` command just calls `sceptre.template.body()` Therefore, the functionality for this PR should probably reside in `sceptre/template.py` maybe as a private method something like `_render_template_as_valid_json()` that `template.body()` can utilise i.e.

```python

# https://github.com/cloudreach/sceptre/blob/master/sceptre/template.py#L67

   @property
    def body(self):
        """
        Represents body of the CloudFormation template.
        :returns: The body of the CloudFormation template.
        :rtype: str
        """
        if self._body is None:
            file_extension = os.path.splitext(self.path)[1]
...

        return self._render_template_as_valid_json(self._body)
```

My second line of thinking is that if we do choose to introduce `cfn_lint` as a dependency then I think it would be better to fully utilise the linter i.e. fail to deploy if linter fails; we could trigger the linter on every command. I have a few concerns with this:

- sceptre as a deployment tool starts to become a development tool i.e. scope increase see: #683 
- the task of linting cfn properly is pretty complicated and the package makes no guarantee it is correct. This could end up frustrating our users. i.e what is the point of providing a linter unless it is 100% accurate.
- we will be playing catch up/relying on `cfn-lint` being up to date.
- thinking ahead to Azure and GCP support we would need to offer matching functionality to keep a consistent API/feature parity.

If we did decide to use the linter fully then we could do something like

```python
# sceptre/template.py

@staticmethod
def lint():
     ... implement linting logic

```

On balance, I am not convinced that using the package suggested is necessarily a good idea, even though it might work. Probably a custom SafeLoader would do the trick.

craighurley:
The very first line in the readme is:

> Sceptre is a tool to drive CloudFormation

... I think including a linter goes too far for sceptre.

We already use `cfn-lint` separately and use custom ignore filters for particular templates that have some oddities.

worr:
I don't particular mean to expand the scope of what sceptre does – that's definitely not within my purview anyway. I just selected that package since it's maintained by Amazon, and provides a loader that I could reuse.

That said, I'm happy to revisit this and implement this with a custom loader if that's what y'all are looking for.

worr:
Hey @ngfgrant!

So I finally got time to revisit this. As recommended, I implemented this without pulling in extra deps, just writing a custom loader that can handle and translate the appropriate tags. I also added an integration test as had been pointed out earlier.

I did opt to keep this in `sceptre.cli.helpers`, since this code is only called from `write`, which is in the same file and the entrypoint for writing out these templates. I didn't think it made too much sense to move this into templates, so far from where it's called. If there's a better place to put the constants for the tag names, I'm happy to move them.

ngfgrant:
Amazing thanks will review and get this in 2.1.4.

sceptre/cli/helpers.py
```diff
@@ -268,3 +269,69 @@ def default(self, item):
...
+    'Transform',
+]
+
+TAGS = [
+    'Condition',
+    'Ref'
+]
+
+
+def getatt_constructor(loader, node):
```
ngfgrant: We could probably make this a private function?

sceptre/cli/helpers.py
```diff
@@ -268,3 +269,69 @@ def default(self, item):
         :rtype: str
         """
         return str(item)
+
+
+FNS = [
```
ngfgrant: Could you rename to something like `CFN_FUNCTIONS`

sceptre/cli/helpers.py
```diff
@@ -268,3 +269,69 @@ def default(self, item):
...
+    'Condition',
+    'Ref'
+]
+
+
+def getatt_constructor(loader, node):
+    if isinstance(node.value, six.text_type):
+        return node.value.split('.', 1)
+    elif isinstance(node.value, list):
+        ret = loader.construct_sequence(node)
```
ngfgrant: What does ret stand for?

sceptre/cli/helpers.py
```diff
@@ -268,3 +269,69 @@ def default(self, item):
...
+        for item in ret:
+            if not isinstance(item, six.text_type):
+                raise ValueError("Fn::GetAtt does not support complex datastructures")
+        return ret
+    else:
+        raise ValueError("Fn::GetAtt only supports string or list values")
+
+
+def tag_constructor(loader, tag_suffix, node):
```
ngfgrant: Could also be a private function (`_tag_constructor`)

sceptre/cli/helpers.py
```diff
@@ -268,3 +269,69 @@ def default(self, item):
...
+    'ImportValue',
+    'Join',
+    'Not',
+    'Or',
+    'Select',
+    'Sub',
+    'Transform',
+]
+
+TAGS = [
```
ngfgrant: Could you rename to something like `CFN_TAGS`

sceptre/cli/helpers.py
```diff
@@ -268,3 +269,69 @@ def default(self, item):
...
+            if not isinstance(item, six.text_type):
+                raise ValueError("Fn::GetAtt does not support complex datastructures")
+        return ret
+    else:
+        raise ValueError("Fn::GetAtt only supports string or list values")
+
+
+def tag_constructor(loader, tag_suffix, node):
+    if tag_suffix not in FNS and tag_suffix not in TAGS:
+        raise ValueError("Bad tag: !{tag_suffix}".format(tag_suffix=tag_suffix))
```
ngfgrant: We could go a step further and show the list of allowed tags in the error message which might be more helpful?

worr:
@ngfgrant I've made allt he requested changes! Let me know if there's anything else you need before merge!

PR #698, opened 2019-05-08
[Resolve #605] Fix docs for protected stacks

### LINKED ISSUE #605: Stack protection feature documentation discrepancy
The sceptre docs refers to stack protection feature with key `protected`[1] however it only seems to work when using the `protect` keyword (i.e. `protect: true`).  a quick grep shows that many tests use `protected` keyword.  Not sure which is the correct keyword to use.

[1] https://sceptre.cloudreach.com/latest/docs/stack_config.html#protected

(py36-sceptredev) ~/w/sceptre ❯❯❯ git grep protected                                                                                                                                               docs/docs/stack_config.md:* [protected](#protected) _(optional)_
docs/docs/stack_config.md:### protected
docs/docs/stack_config.md:If a user tries to run one of these commands on a protected Stack, Sceptre will
sceptre/config/reader.py:            protected=config.get("protect", False),
sceptre/plan/actions.py:        if self.stack.protected:
sceptre/stack.py:    :param protected: Stack protection against execution.
sceptre/stack.py:    :type protected: bool
sceptre/stack.py:        dependencies=None, role_arn=None, protected=False, tags=None,
sceptre/stack.py:        self.protected = protected
sceptre/stack.py:            "protected={protected}, "
sceptre/stack.py:                protected=self.protected,
sceptre/stack.py:            self.protected == stack.protected and
tests/test_actions.py:            role_arn=sentinel.role_arn, protected=False,
tests/test_actions.py:        self.actions.stack.protected = True
tests/test_config_reader.py:            protected=False,
tests/test_plan.py:            role_arn=sentinel.role_arn, protected=False,
tests/test_stack.py:            role_arn=sentinel.role_arn, protected=False,
tests/test_stack.py:        assert stack.protected is False
tests/test_stack.py:            "protected=False, "\

jmourelos:
The documentation is now aligned with the implementation and specifies
that the word "protect" should be used to protect stacks (instead of
"protected").

## PR Checklist

## Other Information

### DIFF
--- a/docs/_source/docs/stack_config.rst
+++ b/docs/_source/docs/stack_config.rst
@@ -18,7 +18,7 @@ particular Stack. The available keys are listed below.
 -  `notifications`_ *(optional)*
 -  `on_failure`_ *(optional)*
 -  `parameters`_ *(optional)*
--  `protected`_ *(optional)*
+-  `protect`_ *(optional)*
 -  `role_arn`_ *(optional)*
 -  `sceptre_user_data`_ *(optional)*
 -  `stack_name`_ *(optional)*
@@ -118,8 +118,8 @@ Example:
        - !stack_output security-groups::BaseSecurityGroupId
        - !file_contents /file/with/security_group_id.txt
 
-protected
-~~~~~~~~~
+protect
+~~~~~~~
 
 Stack protection against execution of the following commands:
 
@@ -131,6 +131,12 @@ Stack protection against execution of the following commands:
 
 If a user tries to run one of these commands on a protected Stack, Sceptre will
 throw an error.
+
+Example:
+
+.. code-block:: yaml
+
+  protect: true
 
 role_arn
 ~~~~~~~~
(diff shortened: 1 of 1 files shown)


ngfgrant:
@jmourelos I think protected is inline with the api? https://github.com/cloudreach/sceptre/blob/master/sceptre/stack.py#L132

jmourelos:
> @jmourelos I think protected is inline with the api? https://github.com/cloudreach/sceptre/blob/master/sceptre/stack.py#L132

@ngfgrant I assumed that this was the expected fix because of the tag "type:docs" in the issue. Otherwise, I could instead change the implementation to check for both keywords "protect" and "protected" in the stack config (to avoid breaking backwards-compatibility). The change I mean would be here: https://github.com/cloudreach/sceptre/blob/2ae24bab9a80a9eb20b79fcc3e7f25abd81f826d/sceptre/config/reader.py#L478

I think it would be specially dangerous to break backwards-compatibility because of the nature of the "protected" feature.

ngfgrant:
In this case I think it is ok to change https://github.com/cloudreach/sceptre/blob/2ae24bab9a80a9eb20b79fcc3e7f25abd81f826d/sceptre/config/reader.py#L478 to `protected`, this is clearly a typo, the docs and API as well as existing usage elsewhere would be `protected`. In other words changing `protect`-> `protected` won't break the public facing API.

jmourelos:
Fair enough, I will close the pull request and create a new one then.

PR #711, opened 2019-05-22
Add Dockerized Sceptre Image

### LINKED ISSUE #445: An official Docker image?

ngfgrant:
A Dockerfile for a sceptre image and updated CircleCI config for deployment to Docker Hub.

Potentially worth having discussion around `ENTRYPOINT` vs `CMD`?

[Resolves #445]

## PR Checklist

## Other Information


hreeder:
It would be worth adding a `.dockerignore` file alongside the `Dockerfile` to limit what files are transferred to the Docker context, especially given the limited number of files being added to the image.

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3.6.8-alpine3.9
+RUN apk add --no-cache bash
+WORKDIR /app
+COPY setup.cfg setup.py README.md CHANGELOG.md ./
+COPY sceptre/ ./sceptre
+RUN python setup.py install
+WORKDIR /project
+CMD ["sceptre"]
```
jamielennox: For containers like this I would suggest we go with ENTRYPOINT here. It means that executing the container looks like:

```
$ docker run --rm -v `pwd`:/project:ro -it cloudreach/sceptre:latest launch prod
```

rather than redundantly specifying sceptre as the cmd. This has always seemed the pattern for a single application container such as this.

Also people will quickly come to think of this as a stable interface so it's best to decide now rather than change it later.

ngfgrant:
@hreeder You had some thoughts on using CMD over ENTRYPOINT from a CICD point of view?

craighurley:
How about pointing `ENTRYPOINT` at a script to provide flexibility, for example.

`Dockerfile`:
```
ENTRYPOINT [ "docker-entrypoint.sh" ]
```

`docker-entrypoint.sh`:
```
#!/bin/sh

set -e

if [ "$1" = "sh" ] ; then
    shift
    exec sh "$@"
else
    exec sceptre "$@"
fi
```

Now you can run sceptre like normal: `docker run --rm -it --v $(pwd):/project:ro cloudreach/sceptre:latest [ARGUMENTS]`

... alternatively you can drop into a shell like this: `docker run --rm -it --v $(pwd):/project:ro cloudreach/sceptre:latest sh`

https://github.com/craighurley/docker-sceptre/blob/master/Dockerfile#L16
https://github.com/craighurley/docker-sceptre/blob/master/docker-entrypoint.sh

jamielennox:
In hindsight whilst the entrypoint is good for direct execution when used in a CI like circle or Bitbucket pipelines you do normally end up with a shell in the image. 

I don't know if entrypoint affects those scenarios (though I'd assume not). Either works for me because it'll probably need overriding in the ci pipeline anyway.

hreeder:
Apologies for the delayed response on this.

My gut feeling is it depends what you expect the primary usage of the docker image to be - from experience, if an entrypoint is set, I've had to unset it for some CI systems to function. I think my biggest bugbear with this at the time was it wasn't a documented dependency in the CI systems (which is their own shortcoming, not a problem with the other images I was using), so I discovered the need to unset the entrypoint myself through investigation rather than from docs.

That said, if the primary usage is going to be by developers on the command line then an entrypoint makes sense (because you'd just be doing `docker run cloudreach/sceptre [args]` rather than the additional `sceptre` prior to `[args]` as previously pointed out.

ngfgrant:
Ok how about we just use ENTRYPOINT:

```Dockerfile
FROM python:3.6.8-alpine3.9
RUN apk add --no-cache bash
WORKDIR /app
COPY setup.cfg setup.py README.md CHANGELOG.md ./
COPY sceptre/ ./sceptre
RUN python setup.py install
WORKDIR /project
ENTRYPOINT ["sceptre"]
```

This would allow us to do something like:

```shell
docker run -v $(pwd):/project -v /Users/me/.aws/:/root/.aws/:ro cloudreach/sceptre:latest --output json generate bundle
```

but would also allow you to overwrite the `ENTRYPOINT` with your own script using `--entrypoint=...` so if you wanted to enter into the container shell you could do so with:

`docker run -ti --entrypoint='' cloudreach:test sh`

Seem reasonable?

PR #719, opened 2019-06-03
Hotfix/windows stack output

lukeplausin:
This PR is to fix an issue whereby the stack_output resolver does not work on windows systems.
https://github.com/cloudreach/sceptre/issues/716

In the stack_output resolver code, a function call is made to the `sceptreise_path` function so that windows path separators are transformed to unix style separators in the template name before comparison.

### DIFF
--- a/sceptre/resolvers/stack_output.py
+++ b/sceptre/resolvers/stack_output.py
@@ -7,7 +7,7 @@ import shlex
 
 from botocore.exceptions import ClientError
 
-from sceptre.helpers import normalise_path
+from sceptre.helpers import normalise_path, sceptreise_path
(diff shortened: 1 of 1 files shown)


ngfgrant:
@lukeplausin Thanks for PR I'll pull down locally and test - will get this out in the next release. It maybe the case we just need to call `sceptreise_path` rather than both `normalise..` and `sceptreise`

lukeplausin:
I think it would be good to add test cases for windows paths in case this gets broken again in the future. Unfortunately my knowledge of mock and pytest is limited...

Arlington1985:
Hi Lukas, when version 2.1.4 will be available in pip?

PR #801, opened 2019-08-07
Allow stack tags to use Resolvers by making them a ResolvableProperty

jlongman:
Simple change to make resolvers work for tags. 

My apologies, I know you work hard and I do appreciate the work you do on this.  I'll admit this is a pretty bad PR but it will be quick to reject at least. ;-).  I'm not familiar enough with your testing structure to get them working much less modify them.

## Notes

Adds `tags` as a `ResolvableProperty` in `sceptre/stacks.py`.

This may be related to #426, #427 but (maybe because I'm new to sceptre) these look like they were filed against v1 (v1.3?) and not v2.

## Explanation

given `./config/dev/config.yaml`:
```
profile: {{ var.profile }}
stack_tags:
  "mv:echo": !echo echo_this
```

and a resolver that looks like:
```
class EchoResolver(Resolver):
    def __init__(self, *args, **kwargs):
        super(EchoResolver, self).__init__(*args, **kwargs)

    def resolve(self):
        return self.argument
```
(note: properly installed ie:
```
from setuptools import setup
setup(
    name='echo',
    py_modules=['echo'],
    entry_points={
        'sceptre.resolvers': [
            'echo = echo:EchoResolver',
        ],
    }
)
```

will generate a tag that looks like:
`mv:echo` | `<tmv.EchoResolver object at 0x10bb52160>`

This is clearly wrong, either the resolver should fail or it should successfully resolve.

The correct output should be (if it should resolve):
`mv:echo` | `echo_this`

I can create a bug if necessary.

## PR Checklist

## Other Information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ Categories: Added, Removed, Changed, Fixed, Nonfunctional, Deprecated
 
 ## Unreleased
 
+## 2.2.1 (2019.08.19)
+
+### Fixed
+
+- `typing` install dependency for Python version < 3.5
+- Race condition in stacks causing RecurrsiveGet exception
+
 ## 2.2.0 (2019.08.16)
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Categories: Added, Removed, Changed, Fixed, Nonfunctional, Deprecated
 
 ## Unreleased
 
+- Fix recursion in sceptre_user_data becoming infinite
+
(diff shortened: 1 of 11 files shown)


ngfgrant:
Thanks for the PR @jlongman - I will try get to this over the next week or two.

ngfgrant:
@jlongman 

Thanks for this PR - I am going to run with the similar implementation that @zaro0508 has done in https://github.com/Sceptre/sceptre/pull/847. 

Thanks again and please post anything you think would be of value to the project!

Issue #791, opened 2019-07-12
No useful error message in templating errors


lukeplausin:
When templating with Jinja2, if there is an error in the template sceptre currently produces some rather unhelpful output. Example:

```
[2019-07-12 17:16:59] - default/examples/testapp - Creating Stack
'dict object' has no attribute 'ec2_user_data'
```

It would be nice if the error handling included some useful output from the exception, such as file name and line number. Example:

```
[2019-07-12 17:16:59] - default/examples/testapp - Creating Stack
Jinja template error in top-level template code - file scripts\userdata_ubuntu.sh.j2, line 27:

{% for logfile in sceptre_user_data.ec2_user_data.cw_log_files | default([]) %}

'dict object' has no attribute 'ec2_user_data'
```

pointbreak1992:
ModuleNotFoundError: No module named 'resolver.file'

PR #792, opened 2019-07-12
Resolve #791 - Improve exception output when Jinja templating errors occur

lukeplausin:
[Your PR description here]

https://github.com/Sceptre/sceptre/issues/791

This PR will improve sceptre output when templating errors occur, by showing the filename, line number and line content of all Jinja2 files in the stack trace to the operator. That will allow them to quickly find and correct errors in their templates and take the guesswork out of template debugging.

## PR Checklist

- [x] Wrote a good commit message & description [see guide below].
- [ ] Commit message starts with `[Resolve #issue-number]`.
- [ ] Added/Updated unit tests.
- [ ] Added/Updated integration tests (if applicable).
- [ ] All unit tests (`make test`) are passing.
- [ ] Used the same coding conventions as the rest of the project.
- [ ] The new code passes flake8 (`make lint`) checks.
- [ ] The PR relates to _only_ one subject with a clear title.
      and description in grammatically correct, complete sentences.

## Approver/Reviewer Checklist

- [ ] Before merge squash related commits.

## Other Information

[Guide to writing a good commit](http://chris.beams.io/posts/git-commit/)

Issue #710, opened 2019-05-16
Sceptre generate command fails with valid yaml


zaro0508:
I have a valid yaml file but when i run sceptre generate it fails with error "An error occured whilst writing the YAML object".  It seems like the problem is the special character `!` in my CFN yaml.  The generate command works if I change `!Ref` to `Ref:`. 

Steps to repo:
```
(py36-default) ❯❯❯ sceptre --version
Sceptre, version 2.1.2

(py36-default) ❯❯❯ cat templates/template.yaml
AWSTemplateFormatVersion: 2010-09-09
Resources:
  MyBucket:
    Type: "AWS::S3::Bucket"
  MyBucketPolicy:
    Type: "AWS::S3::BucketPolicy"
    Properties:
      # sceptre generate fails for this
      Bucket: !Ref MyBucket
      # sceptre generate works for this
      # Bucket:
      #   Ref: MyBucket
      PolicyDocument:
        Version: "2012-10-17"
        Statement:
          - Sid: "AWSCloudTrailAclCheck"
            Effect: "Allow"
            Principal:
              Service: "cloudtrail.amazonaws.com"
            Action: "s3:GetBucketAcl"

(py36-default) ❯❯❯ cat config/prod/test-stack.yaml
template_path: template.yaml
stack_name: test-stack

(py36-default) ❯❯❯ aws cloudformation validate-template --template-body file://templates/template.yaml
{
    "Parameters": []
}

(py36-default) ❯❯❯ echo $?
0

(py36-default) ❯❯❯ sceptre validate  prod/test-stack.yaml
Template prod/test-stack is valid. Template details:
{'Parameters': []}

(py36-default) ❯❯❯ sceptre --output yaml generate prod/test-stack.yaml
An error occured whilst writing the YAML object.
--- []
```

zaro0508:
Verified that PR https://github.com/cloudreach/sceptre/pull/685 fixes this issue.

craighurley:
I might be worth adding that I have similar issues generating json.  Generating text and yaml works fine.

I have **not** verififed if #685 fixes this issue.

Let me know if I should create a separate issue for this?

## Generate commands

```
$ sceptre --debug --output json generate dev
[2019-05-19 11:46:20] - Adding yaml constructors for the entry point groups ['sceptre.hooks', 'sceptre.resolvers']
[...]

PR #832, opened 2019-09-18
Fix/formatting

lukeplausin:
This PR is to improve output when using `sceptre generate` in the case of badly formatted templates.
- Switched default output format from `json` to `text`
- Changed the formatter so that it will append output from multiple stacks rather than display in python serialized format

I'm not sure my commits from #791 are showing in the diff, I thought they were merged?


craighurley:
@lukeplausin in this commit, @ngfgrant changed the default output from text to json: https://github.com/Sceptre/sceptre/commit/65142f9b696b5ff98febff32c4e71396a2698225

I think defaulting to json was a better option than defaulting to text.

I'm not yet doing it, but I plan on using `sceptre generate` on projects that include jinja and troposphere based templates, so I would end up with an array of valid json templates that I could run through `cfn-lint` and `cfn-nag` as part of automated tests.

In general, I (and my colleagues) use yaml for templates, but using `--output yaml` creates an output, that at this point for me, seems harder to process than a json array.  See https://github.com/Sceptre/sceptre/issues/710.  I need to either work out how to process yaml in that format or raise a ticket as requested in issue 710.

lukeplausin:
Personally I only ever use `sceptre generate` to investigate broken templates. When my templates are producing broken cloudformation, neither the json nor yaml outputs work, because of the statement on line 99 attempt to parse the output as yaml.

https://github.com/Sceptre/sceptre/blob/57b4eec5b052dfb81585b598214c575db4bd1bcb/sceptre/cli/helpers.py#L99

To me the most reasonable default behaviour would be that sceptre prints the output exactly as it is being rendered  by my templates and does not try to parse it or do anything fancy.

The current default behaviour of sceptre just produces exceptions, and even the `--text` option does not produce anything useful either since the output is printed in python serialised format which is impossible to read.

craighurley:
> The current default behaviour of sceptre just produces exceptions, and even the --text option does not produce anything useful either

Considering that, I'm curious to know the reason for changing the default to text?

lukeplausin:
The reason is that when the templates are producing broken (unparsable) output, both `json` and `yaml` outputs fail, and the text option is quite difficult to understand. Here is what I see when I use all three:

```bash
$ sceptre --output yaml generate my/broken/template.yaml

An error occured whilst writing the YAML object.

$ sceptre --output json generate my/broken/template.yaml

An error occured writing the JSON object.

$ sceptre --output text generate my/broken/template.yaml

['---\n# Clouformation template for some networking stuff.\n\n\n\nParameters:\n  VpcCidrBlock:\n
   Type: String\n      AllowedPattern: \'^([0-9]{1,3}\\.){3}[0-9]{1,3}(\\/([0-9]|[1-2][0-9]|3[0-2]))?$\'\n    ConstraintDescription: Must be a valid IPv4 range in CIDR notation (e.g. 10.8.0.0/16 )\n    Default: 10.0.0.0/20\n\n  Team:\n    Description: Name of the team\n    Type: String\n\n  Application:\n    Description: Name of the application\n    Type: String\n\n  SubnetCidrBits:\n    Description: The number of bits in the CIDR block of the subnets. This is the inverse of the bitmask length.\n    ConstraintDescription: The subnet CIDR must be between 4 and 16 bits\n    Type: Number\n    Default: 8\n    MinValue: 4\n    MaxValue: 16\n\n\n\n\n  EipNataAllocId:\n    Description: Elastic IP address allocation for the nat-a gateway in subnet pub-a\n    Type: String\n    AllowedPattern: \'^eipalloc\\-([0-9a-f]{8,18})$\'\n    ConstraintDescription: Must be a EIP allocation ID (e.g "eip-abcdef012345678")\n\n\n\n  EipNatbAllocId:\n    Description:
Elastic IP address allocation for the nat-b gateway in subnet pub-b\n    Type: String\n    AllowedPattern: \'^eipalloc\\-([0-9a-f]{8,18})$\'\n    ConstraintDescription: Must be a EIP allocation ID (e.g "eip-abcdef012345678")\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n

(and so on...)

```

The ideal behaviour is for sceptre to reproduce the output exactly as it is being rendered by my templates, so that I can identify and fix syntax errors:

```bash
$ sceptre --output text generate my/broken/template.yaml

# Clouformation template for some networking stuff.



Parameters:
  VpcCidrBlock:
    Type: String
      AllowedPattern: '^([0-9]{1,3}\.){3}[0-9]{1,3}(\/([0-9]|[1-2][0-9]|3[0-2]))?$'
    ConstraintDescription: Must be a valid IPv4 range in CIDR notation (e.g. 10.8.0.0/16 )
    Default: 10.0.0.0/20

  Team:

(and so on...)
```

craighurley:
Makes sense and seems useful to me now, thanks.

lukeplausin:
Glad we're on the same page.

The PR should just concern `helpers.py`, I don't know why my commits from a previous pull request are showing in the diff. I thought these were merged into the core code already:
https://github.com/Sceptre/sceptre/pull/792

Any ideas about that?

ngfgrant:
Hi Luke thanks for the PR - I will have a proper look soon but if you want to pull master to your fork then rebase your branch it should update accordingly?

lukeplausin:
I'll give that a try

lukeplausin:
It was because I didn't update my fork.. All good now!

ngfgrant:
The default behaviour in V1 was `text`. It was a misstep on part to change it to json.

Maybe there is a way to accommodate the flavour of this PR and keep the default as text?

PR #845, opened 2019-09-26
Allow to keep specific versions of docs

1oglop1:
This update prevents github-pages script to delete specific versions.
Which is needed for docs of `1.5.0`.

Merge this before https://github.com/Sceptre/sceptre.github.io/pull/2

### DIFF
--- a/.circleci/documentation-versions.py
+++ b/.circleci/documentation-versions.py
@@ -1,16 +1,43 @@
 """
-Version helper.
+Version helper updates file version-helper.js
+which takes care of injecting versions into rtd template.
+After version-helper.js is updated,
+script prints paths of old versions to be removed.
 
-Checks the directory 'build_dir' and return list of last 7 versions.
-[dev, tag, tag-1, tag-2,..., tag-5]
+EXAMPLE
 
-Then updates file version-helper.js, which takes care of injecting versions
-into rtd template.
-Return paths, for old versions to be removed.
+docs directory (sceptre.github.io) contains following directories
+
+    ├── 1.5.0
+    ├── 2.1.3
+    ├── 2.1.4
+    ├── 2.1.5
+    ├── 2.2.0
+    ├── 2.2.1
+    ├── dev
+    └── latest -> 2.2.1 (symlink)
+
+Script settings:
+KEEP_SPECIFIC_VERSIONS = ["1.5.0"]
+KEEP_ACTIVE_VERSIONS = 3
+
+The content of version-helper.js will be:
+    ['latest', 'dev', '2.2.1', '2.2.0', '2.1.5', '1.5.0']
+    (where latest points to 2.2.1)
+Output of versions to remove will be:
+    <path>/2.1.4
+    <path>/2.1.3
+
 """
 import os
 from operator import attrgetter
 import sys
+import re
+
+VERSION_RE = re.compile(r"\d+.\d+.\d+")
+KEEP_SPECIFIC_VERSIONS = ["1.5.0"]  # these versions won't be removed
+# number of active versions to keep, eg. len([2.2.1, 2.2.0, 2.1.9]) == 3
+KEEP_ACTIVE_VERSIONS = 3
 
 
 def main():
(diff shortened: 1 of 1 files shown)


.circleci/documentation-versions.py
```diff
@@ -19,20 +19,22 @@ def main():
...
+    keep_versions = ['1.5.0']
+    ignored_dirs = {".git", *keep_versions}
     dirs = [
         item
         for item in os.scandir(build_dir)
         if item.is_dir() and item.name not in ignored_dirs
     ]
     sorted_dirs = sorted(dirs, reverse=True, key=attrgetter("name"))
-    active_versions = [item.name for item in sorted_dirs[:7]]
+    active_versions = [item.name for item in sorted_dirs[:7]] + keep_versions
```
ngfgrant: what is the importance of `:7`? maybe we could put that into a constant?

ngfgrant:
also need to fix linting

.circleci/documentation-versions.py
1oglop1: Sounds like a good idea, it makes sure that last `n-1` versions are displayed and old versions removed, as we discussed before and chosen number of 6.

then it builds a `version-helper.js` which fills in this 
image

+ outputs the versions to be removed (which reminds me that I need to update the code in PR as it's not exactly right)

.circleci/documentation-versions.py
```diff
@@ -19,20 +19,22 @@ def main():
...
     versions_to_remove = (item.path for item in sorted_dirs[7:])
     with open(build_dir + "/version-helper.js", "w+") as outf:
         # select 7 latest versions
         outf.write("let versions = {};".format(active_versions))
     # print out old versions to be removed
-    print(",".join(versions_to_remove))
+    print("remove")
```
1oglop1: Remove this

.circleci/documentation-versions.py
```diff
@@ -19,20 +19,22 @@ def main():
...
+    print("\n".join(versions_to_remove))
```
1oglop1: revert to comma separated print

.python-version
```diff
@@ -1,3 +0,0 @@
-2.7-dev
```
ngfgrant: Don’t remove this file pls

.circleci/documentation-versions.py
```diff
@@ -1,16 +1,43 @@
...
-Then updates file version-helper.js, which takes care of injecting versions
-into rtd template.
-Return paths, for old versions to be removed.
 """
 import os
 from operator import attrgetter
 import sys
+import re
+
+VERSION_RE = re.compile(r"\d+.\d+.\d+")
```
ngfgrant: May as well make this `VERSION_REGEX`

.circleci/documentation-versions.py
```diff
@@ -1,16 +1,43 @@
...
+KEEP_SPECIFIC_VERSIONS = ["1.5.0"]  # these versions won't be removed
```
ngfgrant: `KEEP_VERSIONS`

.circleci/documentation-versions.py
```diff
@@ -1,16 +1,43 @@
...
+# number of active versions to keep, eg. len([2.2.1, 2.2.0, 2.1.9]) == 3
+KEEP_ACTIVE_VERSIONS = 3
```
ngfgrant: If you rename to something like `NUMBER_OF_VERSIONS_TO_KEEP` you could lose the comment above. 

Also just checking this should be 7?

.circleci/documentation-versions.py
```diff
@@ -19,15 +46,24 @@ def main():
     except IndexError:
         sys.exit("Missing build dir path: python /path/docs")
 
-    ignored_dirs = {".git"}
+    # directories which contain docs
     dirs = [
```
ngfgrant: You could rename this to something like `documentation_directories` and lose the comment above

.circleci/documentation-versions.py
```diff
@@ -19,15 +46,24 @@ def main():
...
         item
         for item in os.scandir(build_dir)
-        if item.is_dir() and item.name not in ignored_dirs
+        if item.is_dir() and VERSION_RE.match(item.name)
     ]
+
     sorted_dirs = sorted(dirs, reverse=True, key=attrgetter("name"))
```
ngfgrant: If you use `sorted()` on dirs list comp on line 50 you can get rid of this line.

.circleci/documentation-versions.py
```diff
@@ -19,15 +46,24 @@ def main():
...
+        + [item.name for item in sorted_dirs[:KEEP_ACTIVE_VERSIONS]]
+        + KEEP_SPECIFIC_VERSIONS
+    )
+    versions_to_remove = (
+        item.path
+        for item in sorted_dirs[KEEP_ACTIVE_VERSIONS:]
+        if item.name not in KEEP_SPECIFIC_VERSIONS
+    )
     with open(build_dir + "/version-helper.js", "w+") as outf:
         # select 7 latest versions
```
ngfgrant: Nuke this comment

.circleci/documentation-versions.py
1oglop1: The number does not need to be 7 anymore.
Reason for 7 was that the whole list contained 7 items `len(['latest', 'dev', v2.1,..vN..])==7`.
We do not count `dev`, `latest` and `len(KEEP_SPECIFIC_VERSIONS)`  towards the total number of active versions.

.circleci/documentation-versions.py
1oglop1: sure no problem

.circleci/documentation-versions.py
1oglop1: the output of `versions_to_remove` is used in 
https://github.com/Sceptre/sceptre/blob/master/.circleci/github-pages.sh#L72

I wanted to avoid deleting files in python since it was part of a shell script anyway.

.circleci/documentation-versions.py
```diff
@@ -19,17 +45,27 @@ def main():
...
+    )
+    versions_to_remove = (
+        item.path
+        for item in documentation_directories[NUMBER_OF_VERSIONS_TO_KEEP:]
+        if item.name not in KEEP_VERSIONS
+    )
     with open(build_dir + "/version-helper.js", "w+") as outf:
-        # select 7 latest versions
         outf.write("let versions = {};".format(active_versions))
     # print out old versions to be removed
```
ngfgrant: maybe make the comment "print versions to remove to stout for deletion by bash script"

PR #868, opened 2019-11-29
Adding Jinja2 example

nunogrl:
Documentation: Adding Jinja2 example

Provided example showing how to use `sceptre_user_data` and jinja2.


docs/_source/docs/templates.rst
```diff
@@ -31,6 +31,121 @@ Template. Sceptre User Data is accessible within Templates as
...
+          Type: A
+          HostedZoneId: !Ref Zone
+          AliasTarget:
+            DNSName: "{{ rule.dnsnamealias }}"
+            HostedZoneId: "{{ rule.hostedzoneid }}"
+      {% endfor %}{% endif %}
+    Outputs:
+      {% if sceptre_user_data.CNAMErecords is defined %}{% for rule in sceptre_user_data.CNAMErecords %}add{{ rule.record |    replace("-","d")|replace("_","s")|replace('.',"p")}}cnamerecord:
+        Value: !Ref 'add{{ rule.record |replace("-","d")|replace("_","s")|replace('.',"p")}}cnamerecord'
+        # Description: 'add{{ rule.record |replace("-","d")|replace("_","s")|replace('.',"p")}}cnamerecord'
```
ngfgrant: is this deliberately commented out?

docs/_source/docs/templates.rst
```diff
@@ -31,6 +31,121 @@ Template. Sceptre User Data is accessible within Templates as
...
+        Description: '{{ rule.address }}'
+      {% endfor %}{% endif %}
+      {% if sceptre_user_data.ALIASrecords is defined %}{% for rule in sceptre_user_data.ALIASrecords %}add{{ rule.record |    replace("-","d")|replace("_","s")|replace('.',"p")}}aliasrecord:
+        Value: !Ref 'add{{ rule.record |replace("-","d")|replace("_","s")|replace('.',"p")}}aliasrecord'
+        # Description: 'add{{ rule.record |replace("-","d")|replace("_","s")|replace('.',"p")}}aliasrecord'
```
ngfgrant: is this deliberately commented out?

docs/_source/docs/templates.rst
nunogrl: no, this was a mistake. Thanks for noticing :smile:

docs/_source/docs/templates.rst
nunogrl: fixed! thanks

docs/_source/docs/templates.rst
```diff
@@ -31,6 +31,119 @@ Template. Sceptre User Data is accessible within Templates as
...
+
+Stack:
+
+.. code-block:: yaml
+
+    template_path: dns-extras.j2
+    dependencies:
+    - prod/route53/domain-zone.yaml
+    parameters:
+      DomainName: "example.com"!stack_output prod/route53/example-com-zone.yaml::FullDomainName
```
zaro0508: there's a dependency to `prod/route53/domain-zone.yaml` but it is not provided as part of this example?  should we just remove it?  if this is an example of creating a jinja2 template then i would suggest reducing the scope.

docs/_source/docs/templates.rst
```diff
@@ -31,6 +31,119 @@ Template. Sceptre User Data is accessible within Templates as
...
+      Zone: !stack_output prod/route53/example-com-zone.yaml::HostedZoneID
+    sceptre_user_data:
+      CNAMErecords:
+        - record: "example01"
+          address: "example01.otherdomain.com."
+          ttl: 600
```
zaro0508: I think sceptre requires passing in all sceptre numbers in quotes (as strings)?

nunogrl: it's working as is :) but I agree that the documentation should present the best practice.

docs/_source/docs/templates.rst
nunogrl: this contribution is a working stack :) I agree with you. I shall replace the dependency with the values to make it easier to the reader

PR #922, opened 2020-07-26
[Resolves #921] Drop python version 2.7 support

### LINKED ISSUE #921: Circle CI build fails with "ERROR: py27: commands failed"
I believe CircleCI has dropped python version 2.7 support.  The [latest CI build](https://circleci.com/gh/Sceptre/sceptre/4904) failed with `ERROR:   py27: commands failed
` and ... 
```
py27 installed: DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020.
Please upgrade your Python as Python 2.7 won't be maintained after that date. A future
version of pip will drop support for Python 2.7.
```

zaro0508:
Python 2.7 is deprecated and it is no longer availble on
Circle CI build VMs. This PR removes python ver 2.7 support
from Sceptre.


zaro0508:
could you please take a look @ngfgrant ?

ngfgrant:
We are going to need to support Python 2.7 for sceptre 2. 

This issue however might be a good opportunity to move to GitHub Actions for our CI/CD. That would also solve the issue of builds failing for contributors that are outside of the org.

zaro0508:
Wondering how moving to github actions is a solution for #921 @ngfgrant ?  Github actions has also removed python 2.7 support: https://github.blog/changelog/2020-02-27-github-actions-breaking-change-python-2-being-removed-from-all-virtual-environments/

RobReus:
@ngfgrant Why is Python 2.7 support a requirement for Sceptre 2? Python 2.7 is EOL and having to support something that is EOL does not sound like a good idea to me.

ngfgrant:
It maybe that we are forced to make this change as other project dependencies have dropped support for 2.7

PR #928, opened 2020-08-10
[Resolves #919] Add merge_keys option

### LINKED ISSUE #919: Add feature allowing merging of dictionaries
As far as I can tell, there is no feature in Sceptre that would allow a dictionary to defined in a common YAML var file and then merge that with more specific dictionaries in other var files. Is this a feature that Sceptre might implement or accept a PR for?

alexharv074:
Adds CLI option --merge-keys. Dictionaries in multiple var files are
merged instead of overwritten. Test cases added.
    
I have added flake8 noqa in `sceptre/cli/__init__.py` to avoid C901 code
complexity in the cli function.

## PR Checklist

- [X] Added/Updated integration tests (if applicable).
- [X] The new code passes flake8 (`make lint`) checks.

## Other Information

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,9 +209,10 @@ Options:
   --dir TEXT                 Specify sceptre directory.
   --output [text|yaml|json]  The formatting style for command output.
   --no-colour                Turn off output colouring.
-  --var TEXT                 A variable to template into config files.
-  --var-file FILENAME        A YAML file of variables to template into config
-                             files.
+  --var TEXT                 A variable to replace the value of an item in
+                             config file.
+  --var-file FILENAME        A YAML file of variables to replace the values
+                             of items in config files.
   --ignore-dependencies      Ignore dependencies when executing command.
   --merge-vars               Merge dict keys in vars and var files.
   --help                     Show this message and exit.
--- a/README.md
+++ b/README.md
@@ -214,7 +214,8 @@ Options:
   --var-file FILENAME        A YAML file of variables to replace the values
                              of items in config files.
   --ignore-dependencies      Ignore dependencies when executing command.
-  --merge-vars               Merge dict keys in vars and var files.
+  --merge-vars               Merge variables from successive --vars and var
+                             files.
   --help                     Show this message and exit.
 
 Commands:
--- a/README.md
+++ b/README.md
@@ -213,7 +213,7 @@ Options:
   --var-file FILENAME        A YAML file of variables to template into config
                              files.
   --ignore-dependencies      Ignore dependencies when executing command.
-  --merge-keys               Merge dict keys in vars and var files not overwrite.
+  --merge-vars               Merge dict keys in vars and var files.
   --help                     Show this message and exit.
 
 Commands:
--- a/README.md
+++ b/README.md
@@ -198,23 +198,23 @@ Sceptre can be used from the CLI, or imported as a Python package.
 
 ## CLI
 
-```sh
+```text
 Usage: sceptre [OPTIONS] COMMAND [ARGS]...
 
   Sceptre is a tool to manage your cloud native infrastructure deployments.
 
 Options:
-  --version              Show the version and exit.
-  --debug                Turn on debug logging.
-  --dir TEXT             Specify sceptre directory.
-  --output [yaml|json]   The formatting style for command output.
-  --no-colour            Turn off output colouring.
-  --var TEXT             A variable to template into config files.
-  --var-file FILENAME    A YAML file of variables to template into config
-                         files.
-  --ignore-dependencies  Ignore dependencies when executing command.
-  --merge-keys           Merge keys in successive var files not overwrite.
-  --help                 Show this message and exit.
+  --version                  Show the version and exit.
+  --debug                    Turn on debug logging.
+  --dir TEXT                 Specify sceptre directory.
+  --output [text|yaml|json]  The formatting style for command output.
+  --no-colour                Turn off output colouring.
+  --var TEXT                 A variable to template into config files.
+  --var-file FILENAME        A YAML file of variables to template into config
+                             files.
+  --ignore-dependencies      Ignore dependencies when executing command.
+  --merge-keys               Merge keys in successive var files not overwrite.
+  --help                     Show this message and exit.
 
 Commands:
   create         Creates a stack or a change set.
--- a/README.md
+++ b/README.md
@@ -213,6 +213,7 @@ Options:
   --var-file FILENAME    A YAML file of variables to template into config
                          files.
   --ignore-dependencies  Ignore dependencies when executing command.
+  --merge-keys           Merge keys in successive var files not overwrite.
   --help                 Show this message and exit.
 
 Commands:
--- a/README.md
+++ b/README.md
@@ -213,7 +213,7 @@ Options:
   --var-file FILENAME        A YAML file of variables to template into config
                              files.
   --ignore-dependencies      Ignore dependencies when executing command.
-  --merge-keys               Merge keys in successive var files not overwrite.
+  --merge-keys               Merge dict keys in vars and var files not overwrite.
   --help                     Show this message and exit.
 
 Commands:
--- a/docs/_source/docs/cli.rst
+++ b/docs/_source/docs/cli.rst
@@ -75,9 +75,9 @@ we could overwrite ``nested: world`` to ``nested: hi`` using:
   requirement.
 
 Alternatively, it is possible to have keys merged according to a deep merge
-algorith, by specifying ``--merge-keys``. For example:
+algorithm, by specifying ``--merge-keys``. For example:
 
-``sceptre --var CommonTags.Version=1 --var-file vars.yaml launch stack``
+``sceptre --merge-keys --var CommonTags.Version=1 --var-file other_tags.yaml launch stack``
 
 Command reference
 -----------------
--- a/docs/_source/docs/cli.rst
+++ b/docs/_source/docs/cli.rst
@@ -74,10 +74,34 @@ we could overwrite ``nested: world`` to ``nested: hi`` using:
   a dependency. Using a --var-file with all variables set can help meet this
   requirement.
 
-Alternatively, it is possible to have keys merged according to a deep merge
-algorithm, by specifying ``--merge-keys``. For example:
+It is also possible to have keys merged according to a deep merge
+algorithm from successive var files, by specifying ``--merge-vars``. So, if we
+had a second variable file "vars2.yaml":
 
-``sceptre --merge-keys --var CommonTags.Version=1 --var-file other_tags.yaml launch stack``
+.. code-block:: yaml
+
+  # other_vars.yaml
+  ---
+  top:
+    middle3:
+      nested: more world
+
+
+We could merge all of this together using:
+
+``sceptre --merge-vars --var-file vars.yaml --var-file other_vars.yaml launch stack``
+
+The ``top`` dictionary would then be expected to contain:
+
+.. code-block:: python
+
+  {
+    "top": {
+      "middle": {"nested": "hello"},
+      "middle2": {"nested": "world"},
+      "middle3": {"nested": "more world"}
+    }
+  }
 
 Command reference
 -----------------
--- a/docs/_source/docs/cli.rst
+++ b/docs/_source/docs/cli.rst
@@ -74,6 +74,11 @@ we could overwrite ``nested: world`` to ``nested: hi`` using:
   a dependency. Using a --var-file with all variables set can help meet this
   requirement.
 
+Alternatively, it is possible to have keys merged according to a deep merge
+algorith, by specifying ``--merge-keys``. For example:
+
+``sceptre --var CommonTags.Version=1 --var-file vars.yaml launch stack``
+
 Command reference
 -----------------
 
--- a/sceptre/cli/__init__.py
+++ b/sceptre/cli/__init__.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+# flake8: noqa
+
 """
 sceptre.cli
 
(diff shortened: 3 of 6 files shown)


alexharv074:
I would appreciate feedback on the concept, general approach, and how to get the unit tests to pass. (I don't understand how to get my new CLI option available in the cli function under test.)

alexharv074:
This build is failing although the failure does not appear to be related to my changes.

ngfgrant:
Yeh apologies this is a temporary thing. Resolving now. 🙏

sceptre/cli/__init__.py
```diff
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+# flake8: noqa
+
```
alexharv074, line 4: I have disabled flake8 in this file, because my change has tipped the cli function foul of the code complexity check, whereas I feel it would be the job of the maintainers to refactor this (if they wish) to satisfy flake8's code complexity opinion. Also, there does not appear to be any way in flake8 to say to just disable that one check for either the function or the whole file.

alexharv074:
@ngfgrant It looks like the tests still fail. Did you fix the issue you were looking at?

alexharv074:
@ngfgrant We are wondering if we should fork this project or what would the expected turnaround be to get this PR and the one by my colleague merged and released?

zaro0508:
@alexharv074 the networkx dependency issue, #931, has not been fixed yet.

alexharv074:
@zaro0508 I see that networkx 2.4 was reverted however my build is still red. It says "unauthorized" for me. Have I done something wrong?

README.md
```diff
@@ -228,6 +228,7 @@ Options:
   --var-file FILENAME    A YAML file of variables to template into config
                          files.
   --ignore-dependencies  Ignore dependencies when executing command.
+  --merge-keys           Merge keys in successive var files not overwrite.
```
zaro0508, line 231: Could you please add this to the documentation (in sceptre/docs folder) and provide a few examples how to do this and when you might want to do this?  i don't see of any of that info in issue #919 either.

zaro0508 (changes requested):

ngfgrant:
@alexharv074 

Thanks for the PR.

I've had a look through and think the changes made would be better suited in the API rather than the CLI. We really want to keep the very vast majority of logic out of the CLI if possible.

The flag would remain as a CLI option but the merging should be done in https://github.com/Sceptre/sceptre/blob/master/sceptre/config/reader.py

We have an existing strategies folder and it might be a case of adding or using a strategy from the during the variable handling if that makes sense?

Happy to chat through a bit more if you have questions.

alexharv074:
@ngfgrant 

> The flag would remain as a CLI option but the merging should be done in https://github.com/Sceptre/sceptre/blob/master/sceptre/config/reader.py

I don't suppose you can be more specific about where in that class the feature would go?

alexharv074:
> The flag would remain as a CLI option but the merging should be done in https://github.com/Sceptre/sceptre/blob/master/sceptre/config/reader.py
> 
> We have an existing strategies folder and it might be a case of adding or using a strategy from the during the variable handling if that makes sense?
> 
> Happy to chat through a bit more if you have questions.

Yeah, I could definitely use a clue as to precisely where in the code the entry point is for me!

alexharv074:
@ngfgrant I had a bit more of a look. I am a bit confused because it really appears that the merging of YAML files does occur in the CLI and not in the API. Maybe you mean you wish it was in the API? I am not sure I am the right person to refactor all of this, although I am happy to have a go if I have a little bit more guidance. The code base is a bit of a learning curve.

However, the way I see it is my change could probably go in in its current form before or after the above refactoring without really adding to the complexity of it?

alexharv074:
@ngfgrant any thoughts?

jerry153fish:
LGTM.

sceptre/cli/__init__.py
ngfgrant: @alexharv074 you can do `#noqa` in line with the function def and it will skip that specific function. 

Generally speaking - if flake8 is complaining about complexity then it is probably a good idea to refactor to help the maintainers understand the PR more easily.

alexharv074:
@zaro0508 thoughts?

README.md
alexharv074: I just updated the docs to provide an example. There are more examples in the tests. Is that sufficient?

zaro0508: tests are not docs nor are they examples.  users wanting to use this feature will most likely never try to understand your tests.   Also your example does not explain the purpose of this new flag.  It was not at all obvious to me why and when I would want to use this flag.

Also the description containing `..not overwrite` isn't really true, correct?  Your tests show that vars do get overrriden when using this flag. 

Another issue that may require further discussion is possibly making this functionality the default instead of requiring the user to specify `--merge-keys` flag.  I'm not sure I see any issue with that but others may disagree.  I would suggest you get on [#sceptre slack channel](https://sceptre.cloudreach.com/2.6.2/index.html?highlight=slack#communication) to drive this discussion.

sceptre/cli/__init__.py
zaro0508: i disagree with this statement `..whereas I feel it would be the job of the maintainers to refactor this..`.  I agree with @ngfgrant here, please follow his advice.

sceptre/cli/__init__.py
```diff
@@ -39,21 +41,46 @@
...
 @catch_exceptions
 def cli(
-        ctx, debug, directory, output, no_colour, var, var_file, ignore_dependencies
+        ctx, debug, directory, output, no_colour, var, var_file, ignore_dependencies, merge_keys
 ):
     """
     Sceptre is a tool to manage your cloud native infrastructure deployments.
-
     """
+    def deep_merge(source, destination):
```
zaro0508, line 63: I agree with @ngfgrant's suggestion, `I've had a look through and think the changes made would be better suited in the API rather than the CLI. We really want to keep the very vast majority of logic out of the CLI if possible.`?  Have you made an attempt to move this functionality into the API yet?

zaro0508 (changes requested):

sceptre/cli/__init__.py
alexharv074: @zaro0508 I am happy to move the functionality, but I would need some guidance as I have noted above. When I looked at it last time, I could not figure out what had to move and where it had to move to. I can have another look perhaps if I get some time. But if it's too difficult, we would probably just use our fork and not submit the feature at all.

zaro0508: The idea is that the CLI should only be responsible for gathering the user variables then it should pass those variables along to config/reader.py for processing.  You should do your processing in reader.py which is the API.

I think i would suggest you leave this code as-is because it already gathers all the user vars and passes it along to reader.py.  You can access the all the[ user vars in reader.py](https://github.com/Sceptre/sceptre/blob/master/sceptre/config/reader.py#L132) and update them there.  Also it looks like [config/strategies.py already has a dict_merge](https://github.com/Sceptre/sceptre/blob/master/sceptre/config/strategies.py#L36) method that you can use to merge the keys.

sceptre/cli/__init__.py
alexharv074: @ngfgrant fwiw, placing `#noqa` on the line with the function def does not seem to disable the linter for flake8, e.g.

```diff
diff --git a/sceptre/cli/__init__.py b/sceptre/cli/__init__.py
index 8382063..7d35dae 100644
--- a/sceptre/cli/__init__.py
+++ b/sceptre/cli/__init__.py
@@ -54,7 +54,7 @@ from sceptre.cli.helpers import setup_logging, catch_exceptions
     help="Deep merge keys in successive var dicts.")
 @click.pass_context
 @catch_exceptions
-def cli(
+def cli(  # noqa
         ctx,
         debug,
         directory,
@@ -65,7 +65,7 @@ def cli(
         ignore_dependencies,
         j2_extensions,
         merge_keys
-):
+):  # noqa
     """
     Sceptre is a tool to manage your cloud native infrastructure deployments.
     """
```

```
make lint
...
sceptre/cli/__init__.py:34:1: C901 'cli' is too complex (15)
```

sceptre/cli/__init__.py
alexharv074: @zaro0508 

> I think i would suggest you leave this code as-is because it already gathers all the user vars and passes it along to reader.py.

Unfortunately, that is not quite true. As you can see [here](https://github.com/Sceptre/sceptre/blob/master/sceptre/cli/__init__.py#L76-L82) on lines 76-82, duplicate dictionary keys in the first `var_file` are deleted in favour of those appearing in the second `var_file`. This behaviour is further illustrated in the tests [here](https://github.com/Sceptre/sceptre/blob/master/tests/test_cli.py#L113-L125).

Moreover, the context object does not allow us to pass in the raw dictionaries from the `var_files`, but has only a single field [`user_variables`](https://github.com/Sceptre/sceptre/blob/master/sceptre/cli/__init__.py#L62), where it is expected that these dictionaries are already merged.

The `ConfigReader` class then initialises the `user_variables` [here](https://github.com/Sceptre/sceptre/blob/master/sceptre/config/reader.py#L132.

This is why I did not think my change here should refactor this code. 

I suppose, what we could do is remove the `user_variables` input to `ConfigReader`, and replace it with a list of `--var-files` and a list of `--vars`. Then, the code in the `cli` function could be moved to a private method of the `ConfigReader` class.

But would this change have other consequences? I am not sure.

alexharv074: Actually, this would require quite extensive changes all over the place. It is not only the `ConfigReader` class that expects a merged `user_variables` as an input, but the `SceptreContext` class expects it too. 

This tells me that, without a major redesign of the code base, the logic probably belongs in `sceptre.cli.helpers`.

alexharv074: @zaro0508 for what it's worth, this is what it looks like after moving the functionality to `sceptre.cli.helpers`: https://github.com/Sceptre/sceptre/pull/928/commits/64dbd6f5e67e4e26c52ef11f8bab8b24b51ae4e7

README.md
alexharv074: @zaro0508 I have updated the docs. I can't think of a better wording than 

```
--merge-keys           Merge keys in successive var files not overwrite.
```

The existing behaviour is for keys found in the last `var_file` to overwrite any that were defined in earlier ones. The new behaviour when `--merge-keys` is specified is to merge these keys instead of overwriting them. It is true that in some scenarios some subkeys still end up being overwritten according to the deep merge algorithm but I can't think of any easy way to express that in a few words. I am open to suggestions.

docs/_source/docs/cli.rst
```diff
@@ -74,6 +74,11 @@ we could overwrite ``nested: world`` to ``nested: hi`` using:
   a dependency. Using a --var-file with all variables set can help meet this
   requirement.
 
+Alternatively, it is possible to have keys merged according to a deep merge
+algorith, by specifying ``--merge-keys``. For example:
+
+``sceptre --var CommonTags.Version=1 --var-file vars.yaml launch stack``
```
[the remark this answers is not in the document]
alexharv074: fixed.

README.md
```diff
@@ -198,22 +198,23 @@ Sceptre can be used from the CLI, or imported as a Python package.
...
+  --version                  Show the version and exit.
+  --debug                    Turn on debug logging.
+  --dir TEXT                 Specify sceptre directory.
+  --output [text|yaml|json]  The formatting style for command output.
+  --no-colour                Turn off output colouring.
+  --var TEXT                 A variable to template into config files.
+  --var-file FILENAME        A YAML file of variables to template into config
+                             files.
+  --ignore-dependencies      Ignore dependencies when executing command.
+  --merge-keys               Merge dict keys in vars and var files not overwrite.
```
zaro0508, line 216: again, i have a problem with `no overwrite` in this description. 

what do you think about changing this to `--merge-vars`?

zaro0508 (changes requested):

docs/_source/docs/cli.rst
```diff
@@ -74,6 +74,11 @@ we could overwrite ``nested: world`` to ``nested: hi`` using:
   a dependency. Using a --var-file with all variables set can help meet this
   requirement.
 
+Alternatively, it is possible to have keys merged according to a deep merge
```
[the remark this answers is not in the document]
alexharv074: @zaro0508 How does it read now?

sceptre/cli/__init__.py
```diff
@@ -39,63 +38,36 @@
...
     Sceptre is a tool to manage your cloud native infrastructure deployments.
-
     """
     logger = setup_logging(debug, no_colour)
     colorama.init()
     # Enable deprecation warnings
     warnings.simplefilter("always", DeprecationWarning)
     ctx.obj = {
-        "user_variables": {},
+        "user_variables": setup_vars(var_file, var, merge_keys, logger),
```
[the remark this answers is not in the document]
alexharv074: ok done.

tests/test_cli.py
```diff
@@ -131,6 +131,111 @@ def raises_exception():
...
+                        "Envlist": ["sandbox", "dev"]
+                    }
+                },
+                "test.yaml": {"CommonTags": {"Envlist": ["test"]}},
+            },
+            {
+                "CommonTags": {
+                    "Organization": "Parts Unlimited",
+                    "Department": "IT Operations",
+                    "Envlist": ["test"]  # May not be what we ultimately want!
```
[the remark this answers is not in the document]
alexharv074: Oh, my thinking is that perhaps someone might expect the lists to be merged too. At the moment I'm happy with it as-is, if you are?

tests/test_cli.py
```diff
@@ -131,6 +131,111 @@ def raises_exception():
...
+            },
+            {
+                "CommonTags": {
+                    "Organization": "Parts Unlimited",
+                    "Department": "IT Operations",
+                    "Envlist": ["sandbox", "dev"],
+                    "Version": "1.0.0"
+                }
+            }
+        ),
```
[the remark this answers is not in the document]
alexharv074: ok, done.

README.md
alexharv074: agreed that that's better, done.

sceptre/cli/__init__.py
```diff
@@ -39,63 +38,35 @@
...
     "--var-file", multiple=True, type=click.File("rb"),
-    help="A YAML file of variables to template into config files.")
+    help="A YAML file of variables to replace the values of items in config files.")
 @click.option(
-    "--ignore-dependencies", is_flag=True, help="Ignore dependencies when executing command.")
+    "--ignore-dependencies", is_flag=True,
+    help="Ignore dependencies when executing command.")
+@click.option(
+    "--merge-vars", is_flag=True, default=False,
+    help="Merge dict keys in successive vars and var files.")
```
zaro0508, line 51: How about this instead?  `Merge variables from successive --vars and var files` or `Aggregate variables  from successive --vars and var files`

docs/_source/docs/cli.rst
```diff
@@ -74,6 +74,35 @@ we could overwrite ``nested: world`` to ``nested: hi`` using:
...
+.. code-block:: python
+
+  {
+    "top": {
+      "middle": {"nested": "hello"},
+      "middle2": {"nested": "world"},
+      "middle3": {"nested": "more world"}
+    }
+  }
+
```
zaro0508, line 105: 👍

tests/test_cli.py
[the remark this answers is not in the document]
zaro0508: Please either remove this comment or provide a more thorough description of your thought.  It makes reading the code more confusing if left in as-is

zaro0508 (approved):

sceptre/cli/__init__.py
alexharv074: ok done.

PR #958, opened 2020-12-18
Add documentation for AWS CDK

zaro0508:
Sceptre can generate cloudformation templates from AWS CDK code
without much effort. Adding documention on how to use sceptre
to drive AWS CDK python files.

### DIFF
--- a/docs/_source/docs/templates.rst
+++ b/docs/_source/docs/templates.rst
@@ -156,6 +156,9 @@ Config file, Sceptre passes an empty ``dict``.
 Example
 ~~~~~~~
 
+Troposphere
+^^^^^^^^^^^
+
 This example is using `troposphere`_
 to generate CloudFormation Template as a `json` string.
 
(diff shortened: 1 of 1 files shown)


docs/_source/docs/templates.rst
```diff
@@ -177,3 +180,57 @@ to generate CloudFormation Template as a `json` string.
...
+    def __init__(self, scope: cdk.Stack, construct_id: str, sceptre_user_data, **kwargs) -> None:
+      super().__init__(scope, construct_id, **kwargs)
+
+      bucket_name = sceptre_user_data['BucketName']
+      bucket = s3.Bucket(self, "S3Bucket",
+                         bucket_name=bucket_name)
+
+  def sceptre_handler(sceptre_user_data):
+      S3CdkStack(cdk.App(), "S3CdkStack", sceptre_user_data)
+      template = app.synth().get_stack("S3CdkStack").template
```
dlahn, line 217: Where is `app` defined?

dlahn:
I think there is some issue here. It works fine to generate an individual stack, however, trying to run multiple stacks using the same template file blows up with all sorts of random errors, like:

```
'CloudAssembly' object has no attribute 'template'
'CloudAssembly' object has no attribute 'template'
'CloudAssembly' object has no attribute 'template'
'CloudFormationStackArtifact' object has no attribute 'get_stack'
'CloudFormationStackArtifact' object has no attribute 'get_stack'
'CloudFormationStackArtifact' object has no attribute 'get_stack'
"Parameter validation failed:\nInvalid type for parameter TemplateBody, value: None, type: <class 'NoneType'>, valid types: <class 'str'>"
```

docs/_source/docs/templates.rst
zaro0508: fixed.  thanks for pointing out that error.  Please let us know if this works for you.

zaro0508:
did my fix make any difference for you @dlahn ?

dlahn:
@zaro0508 I had already made the change you proposed on our side. We are still seeing those errors with the change in place. It works fine for an individual stack, but not when we run a folder for example, containing a number of stacks that use the same template file.

zaro0508:
hmm, it seems to work for me.  Can you please provide pip/version info and see if your environment is much different than mine? here is details of my environment and execution: http://paste.openstack.org/show/801572/

dlahn:
@zaro0508 It's pretty similar. I'm running Python 3.8.1. I was on Sceptre 2.4, but 2.3 exhibits the same problem. Upgrading CDK to what you have gives a few new errors:

```
Cannot reference across apps. Consuming and producing stacks must be defined within the same CDK app.
'CloudAssembly' object has no attribute 'template'
Cannot reference across apps. Consuming and producing stacks must be defined within the same CDK app.
'CloudFormationStackArtifact' object has no attribute 'get_stack'
```

These are definitely defined in the same apps. I use the same approach you do above. 

As far as your run example, it doesn't look like (though I could be wrong) that you are running a number of the same template in parallel. In my case, I have about 6 stacks defined that insert routes into a route table. They all use `route.py`. If I run them individually, all is good. If I run the whole folder, it all blows up/hangs. Perhaps there is some sort of collision between these being run at the same time by Sceptre?

lloydpick:
Might be related to https://github.com/aws/aws-cdk/issues/1479

> As an additional note about something that I missed last time around: App is designed to only be instantiated once. It's represents the "CDK App", not "your app".

zaro0508:
Thanks for the reference @lloydpick.    That makes sense because my test run consisted of one CDK file with a bunch of other non-CDK files.  The issue that @dlahn is seeing seems like a CDK problem not a sceptre problem.

PR #971, opened 2021-02-08
[Resolves #663] Support changeset creation for non-existent stacks

### LINKED ISSUE #663: Add ability to create a new stack with change sets
I would like to be able to use change sets on new stacks as well as existing stacks.

AWS/boto supports creating a new stack with change sets using the `ChangeSetType` keyword [see here for details](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation.html#CloudFormation.Client.create_change_set).  To the untrained eye, it appears that updating the cli `create` and `update` to pass an argument of `change_set_type` to the `create_change_set` function, and updating that function to accept the new argument may get me the functionality I'm after. I'm up for a PR if this makes sense to implement.

Outputs may be a bit tricky to deal with, but I haven't gotten that far down the path yet.

It looks like there was an attempt to address this with #198 , but that appears to have been closed, and inactive for awhile (thus the new issue).

acaire:
This commit modifies the behaviour of changeset creation
so that when a non-existent stack error occurs, instead
of failing, the call will be reattempted with an amended
ChangeSetType param of CREATE instead of UPDATE

I feel that update before create behaviour is likely to cover the majority of create/update attempts and makes the most of Python's ask forgiveness philosophy.

### DIFF
--- a/sceptre/plan/actions.py
+++ b/sceptre/plan/actions.py
@@ -28,6 +28,8 @@ from sceptre.exceptions import UnknownStackChangeSetStatusError
 from sceptre.exceptions import StackDoesNotExistError
 from sceptre.exceptions import ProtectedStackError
 
+from botocore.exceptions import ClientError
+
 
 class StackActions(object):
     """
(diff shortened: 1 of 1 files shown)


acaire:
Unfortunately even with this PR, execution still fails for dependant stacks, so I've had to resort to dropping changeset execution via CI completely, and just blindly use launch instead and hope for the best 😭

PR #980, opened 2021-03-17
Resolve #982 add examples of on_failure

### LINKED ISSUE #982: Documentation should provide examples for on_failure option
Documentation provides a link to AWS documentation, but it should include a few examples on how to use `on_failure`. This makes it easier to search on a search engine.

#980  addresses this.

felipe1982:
Resolve #982 

this makes it easy to search for "rollback" on google.

add three valid examples of values for the `on_failure` key. This makes it easier to search in google for **sceptre rollback**.

## PR Checklist

## Other Information


zxiiro (approved):
Good idea. Saves having to click another link just to see available options.

zaro0508:
I'm not sure I agree @zxiiro.  I think this would cause a maintenance burden on sceptre.  IMO it's better to reference the source than to be the source.

zxiiro:
> I'm not sure I agree @zxiiro. I think this would cause a maintenance burden on sceptre. IMO it's better to reference the source than to be the source.

Upstream will always be the primary source of truth but I also don't think it's harmful to list the the valid options here on a best effort basis. I also believe the `OnFailure` parameter is unlikely to change very often if at all.

zaro0508:
What are your thought @craighurley ?

craighurley:
For me, in this case the wording makes it clear that they are examples and the source is the AWS docs, and I think those example values are unlikley to be changed by AWS in the future.

zaro0508 (approved):

PR #983, opened 2021-03-28
Detail how to use Sceptre hooks to deploy lambda

zxiiro:
In order to deploy Lambda functions using either the AWS CLI
or SAM CLI we need to generate the packaged version of the
Lambda functions first. This patch details how to add the "package"
command using Sceptre hooks in a way that reuses existing Sceptre
config.yaml file information to populate the command details.

## PR Checklist

## Other Information

### DIFF
--- a/docs/_source/docs/faq.rst
+++ b/docs/_source/docs/faq.rst
@@ -95,6 +95,38 @@ be found in these projects:
 - `Sage-Bionetworks Sceptre lambda github template`_
 - `Versant SAM Sceptre skeleton example`_
 
+Another approach is to use the Sceptre `before_launch` hook to run either the
+SAM or CFN "package" command to generate the packaged version of the template.
+
+CloudFormation Example:
+
+```yaml
+template_path: generated/lambda-packaged.json
+stack_name: cfn-lambda-example
+hooks:
+  before_launch:
+    - !cmd >
+        aws --profile {{ stack_group_config.profile }} cloudformation package
+        --template-file templates/cfn-lambda.yaml
+        --s3-bucket {{ stack_group_config.template_bucket_name }}
+        --output json
+        --output-template-file templates/generated/lambda-packaged.json
+```
+
+SAM Example:
+
+```yaml
+template_path: generated/sam-packaged.yaml
+stack_name: sam-example
+hooks:
+  before_launch:
+    - !cmd >
+        sam package --profile {{ stack_group_config.profile }}
+        --s3-bucket {{ stack_group_config.template_bucket_name }}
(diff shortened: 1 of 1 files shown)


docs/_source/docs/faq.rst
```diff
@@ -95,6 +95,52 @@ be found in these projects:
...
+run either the SAM or CFN "package" command to generate the packaged version of
+the template.
+
+CloudFormation Example:
+
+```yaml
+template_path: generated/lambda-packaged.json
+stack_name: cfn-lambda-example
+hooks:
+  before_create:
```
zaro0508, line 108: i always use launch command which does both create and update.  seems like you can just do this with just one `before_launch` hook ?

zxiiro: No the problem with `before_launch` is it executes for every stack. So if you have 10 stacks it will run 10 times. I've found `before_create` and `before_update` is the only way to get it to run once even when you run `sceptre launch`.

I saw in the dev branch there's a `before_generate` which I'd be curious if it will run only once per stack.

zaro0508: i'm not sure i understand.  if i run `launch --ignore-dependencies prod/my_stack.yaml` the hook will only execute once for the stack defined in my_stack.yaml

zxiiro: Do you usually run with `sceptre launch --ignore-depencencies prod/my_stack.yaml` to deploy each one of your stacks?

Normally I run like this `sceptre launch -y prod` to deploy everything under prod. If you do this using the `before_launch` hook, that hook will execute x times where x is the number of stacks that exist in the prod environment.

It seems the `before_launch` hook runs against every stack and not just the one stack when you are deploying multiple stacks.

zxiiro: Just retested and it seems like `before_launch` is working as you describe. I'm not sure why I was having problems with it earlier but it's working for me now too. I'll update the PR.

docs/_source/docs/faq.rst
```diff
@@ -95,6 +95,39 @@ be found in these projects:
 - `Sage-Bionetworks Sceptre lambda github template`_
 - `Versant SAM Sceptre skeleton example`_
 
+Another approach is to use Sceptre `before_create` and `before_update` hooks to
```
zaro0508, line 98: change to `before_launch`?

zaro0508 (changes requested):

docs/_source/docs/faq.rst
zxiiro: done

zaro0508 (approved):

PR #987, opened 2021-03-30
update container image for builds

zaro0508:
Update container image with fix from PR https://github.com/Sceptre/sceptre-circleci/pull/3


zaro0508:
done in https://github.com/Sceptre/sceptre/pull/991

PR #990, opened 2021-03-31
Disable integration tests

zaro0508:
Disable integration tests because we don't have access to the
AWS account it depends on.  We'll need to get an account before
we can run these tests again.

The intention is to disable it to get the builds passing again and
to be able to merge non-code changes. We also inted to get these
tests running again before merging any code related changes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -136,28 +136,31 @@ jobs:
             . venv/bin/activate
             make sonar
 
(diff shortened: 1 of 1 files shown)


rdkr (approved):

zaro0508:
done in https://github.com/Sceptre/sceptre/pull/991

Issue #988, opened 2021-03-31
InvalidClientTokenId during DescribeStacks


alexharv074:
Hi there, I am receiving this failure that I cannot debug due to insufficient feedback from Sceptre:

```
[2021-04-01 04:29:25] - iam-cseo - termination protection set to 'enabled'
"An error occurred (InvalidClientTokenId) when calling the DescribeStacks operation: The security token included in the request is invalid."
**********************************************************************************************************************************************************************************                                                       launch on foundations/config/foundations.yaml returned exit status 1
**********************************************************************************************************************************************************************************
```

Outside of Sceptre, I am unable to reproduce any such issue about InvalidClientTokenId.

I have spent a couple of hours placing print statements variously through the Sceptre code but I just can't find where or what is emitting that error.

Can anyone give me any clues on where it might be coming from?

craighurley:
Looks like a boto error.  Are you using a particular IAM role when running sceptre?  Does that role have perms to `DescribeStacks`.

alexharv074:
@craighurley Yes and yes. `aws cloudformation describe-stacks` runs fine for instance.

craighurley:
Did you figure out the cause in the end?

craighurley:
Relates to #897

Issue #897, opened 2020-04-27
Improve error message for failed deployments


zaro0508:
I'll see the below error messages in my sceptre deploy log however it doesn't tell me which file contains the error.  It would be great if sceptre could provide a little more info in these situations.  I I tried using the sceptre `--debug` flag however that didn't provide anymore info. 

```
..
..
[2020-04-27 08:18:01] - prod/sc-product-ec2-linux-jumpcloud sc-product-ec2-linux-jumpcloud AWS::CloudFormation::Stack UPDATE_COMPLETE
"An error occurred (ValidationError) when calling the CreateStack operation: Template format error: unsupported structure."
```

```
..
..
[2020-04-27 08:46:42] - prod/sc-product-ec2-linux-jumpcloud sc-product-ec2-linux-jumpcloud AWS::CloudFormation::Stack UPDATE_COMPLETE
"An error occurred (ValidationError) when calling the CreateStack operation: Template format error: YAML not well-formed. (line 44, column 7)"
```

ngfgrant:
Error handling is something that has come up in https://github.com/Sceptre/sceptre/issues/640 and https://github.com/Sceptre/sceptre/issues/564.

I think my suggestion in #640 is probably breaking but could be evolved?

PR #997, opened 2021-04-06
Stop hiding critical debug info in helpers (#988)

alexharv074:
Before this, the catch_exceptions function in cli/helpers would catch a
range of exceptions and then hide all but the error message from the
caller.

Over the years, this has caused myself and others quite a lot of lost
time, as it is now often quite unclear what caused Sceptre to fail.

Simply re-raising the original exception provides valuable information
to allow users of Sceptre to debug their failing code.

[Your PR description here]

## PR Checklist

## Other Information


alexharv074:
I am not able to see why the build is failing here but `make lint` and `make test` work fine for me.

zaro0508:
build fixed with PR #998

zaro0508:
I'm not sure doing this across the board is better than the existing functionality.  For example when i run scpeptre (ver 2.4.0) with an invalid profile i get this..
```
➜ sceptre --var "profile=my-invalid-profile" --var "region=us-east-1" --ignore-dependencies launch develop/docker-runner.yaml
Do you want to launch 'develop/docker-runner.yaml' [y/N]: y
[2021-04-08 08:42:20] - develop/docker-runner - Launching Stack
"The config profile (my-invalid-profile) could not be found"
```

when i run with this change I get this..
<details>
<summary>invalid profile console out</summary>
➜  sceptre --var "profile=my-invalid-profile" --var "region=us-east-1" --ignore-dependencies launch develop/docker-runner.yaml
Do you want to launch 'develop/docker-runner.yaml' [y/N]: y
[2021-04-08 08:41:07] - develop/docker-runner - Launching Stack
"The config profile (my-invalid-profile) could not be found"
  File "/Users/zaro0508/.virtualenvs/bla9/bin/sceptre", line 8, in <module>
    sys.exit(cli())
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/click/core.py", line 829, in __call__
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/click/core.py", line 610, in invoke
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/click/decorators.py", line 21, in new_func
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/sceptre/cli/helpers.py", line 35, in decorated
    return func(*args, **kwargs)
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/sceptre/cli/launch.py", line 38, in launch_command
    responses = plan.launch()
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/sceptre/plan/plan.py", line 130, in launch
    return self._execute(*args)
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/sceptre/plan/plan.py", line 40, in _execute
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/sceptre/plan/executor.py", line 55, in execute
  File "/usr/local/var/pyenv/versions/3.8.6/lib/python3.8/concurrent/futures/_base.py", line 432, in result
    return self.__get_result()
  File "/usr/local/var/pyenv/versions/3.8.6/lib/python3.8/concurrent/futures/_base.py", line 388, in __get_result
    raise self._exception
  File "/usr/local/var/pyenv/versions/3.8.6/lib/python3.8/concurrent/futures/thread.py", line 57, in run
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/sceptre/plan/executor.py", line 62, in _execute
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/sceptre/hooks/__init__.py", line 104, in decorated
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/sceptre/plan/actions.py", line 192, in launch
    existing_status = self._get_status()
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/sceptre/plan/actions.py", line 748, in _get_status
    status = self._describe()["Stacks"][0]["StackStatus"]
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/sceptre/plan/actions.py", line 740, in _describe
    return self.connection_manager.call(
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/sceptre/connection_manager.py", line 51, in decorated
    return func(*args, **kwargs)
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/sceptre/connection_manager.py", line 252, in call
    client = self._get_client(service, region, profile, stack_name, iam_role)
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/sceptre/connection_manager.py", line 216, in _get_client
    self._clients[key] = self._get_session(
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/sceptre/connection_manager.py", line 144, in _get_session
    session = boto3.session.Session(**config)
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/boto3/session.py", line 80, in __init__
    self._setup_loader()
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/boto3/session.py", line 120, in _setup_loader
    self._loader = self._session.get_component('data_loader')
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/botocore/session.py", line 685, in get_component
    return self._components.get_component(name)
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/botocore/session.py", line 924, in get_component
    self._components[name] = factory()
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/botocore/session.py", line 158, in <lambda>
    lambda:  create_loader(self.get_config_variable('data_path')))
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/botocore/session.py", line 240, in get_config_variable
    return self.get_component('config_store').get_config_variable(
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/botocore/configprovider.py", line 313, in get_config_variable
    return provider.provide()
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/botocore/configprovider.py", line 410, in provide
    value = provider.provide()
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/botocore/configprovider.py", line 471, in provide
    scoped_config = self._session.get_scoped_config()
  File "/Users/zaro0508/.virtualenvs/bla9/lib/python3.8/site-packages/botocore/session.py", line 340, in get_scoped_config
    raise ProfileNotFound(profile=profile_name)
botocore.exceptions.ProfileNotFound: The config profile (my-invalid-profile) could not be found
</details>

Would it make sense to pinpoint the problem you were having in issue  #998 first then determine a fix that's more targeted?

alexharv074:
> Would it make sense to pinpoint the problem you were having in issue #988 first then determine a fix that's more targeted? Also would you not get more debug info if you simply pass in the sceptre `--debug` flag?

We already use `--debug` and that does not help. I solved the issue in #988 and it took me a very long time because the stack trace was hidden. When I finally figured out how to unhide the stack trace, it then became obvious what the problem was. (In that case, it was because I didn't have IAM permissions in the account it was trying to read a stack_output_external from.)

I don't think users should lose hours of development time trying to debug something just to make output look prettier when an exception is raised. If pretty output is required, I think the specific exception where pretty output is required should be handled. We end up in this code block precisely because the developers haven't bothered to handle the exception. It's also by no means the first time we lost development time as a consequence of the stack trace being hidden.

alexharv074:
> If --debug is not displaying the exception, maybe that's the change that should be made?

@craighurley Yes true. Would you like me to refactor and only show the stack trace in debug mode?

zxiiro:
I agree with making `--debug` display more information including the stack trace. If it's not passed displaying less information.

craighurley:
Hi @alexharv074 , refactoring to only show the strace in debug would be great.

alexharv074:
> Hi @alexharv074 , I think refactoring to only show the strace in debug would be great, but we should consider addressing the root cause as @zaro0508 mentioned.

@craighurley Could you clarify what you mean by fixing the "root cause"? The root cause was that I did not have permission to read from an S3 bucket.

craighurley:
Hi @alexharv074 , for the root cause what I meant was that if the exception you were getting in #988 was not helpful, could it be caught with better matching and print something more useful for you (without `--debug`), for example https://github.com/Sceptre/sceptre/blob/881c6bd1877685118845d5e3e1b582e7ed1011e1/sceptre/template.py#L262-L269

That plus adding the stack trace as you described once `--debug` is used.

What do you think?

craighurley:
Also relates to #897

alexharv074:
> Hi @alexharv074, 
> [snip]
> What do you think?

I agree but I have no idea without testing whether that's the right try/except block that was causing my original issue. I do know that my comments from above were:

> ... it was because I didn't have IAM permissions in the account it was trying to read a stack_output_external from

Not sure if that helps.

craighurley:
> Whereas, simply refactoring this PR to only hide stack traces when mode != debug is something we can and should get merged quickly. Thoughts?

I agree.

ngfgrant:
I agree more useful debugging and errors is definitely something the project needs (I think there are plenty of open issues about this). 

The main issue with this change is that users may well (and probably) do rely on status exit codes on the CLI. Removing the status code and replacing it with a dump of the stack trace isn't going to be a good solution.

Instead - better debug logging statements will help imo.

alexharv074:
> > Whereas, simply refactoring this PR to only hide stack traces when mode != debug is something we can and should get merged quickly. Thoughts?
> 
> I agree.

@craighurley I have a bit of time this weekend, but `catch_exceptions` is a decorator, and I don't know how to make the logger available inside a decorator. Any ideas?

sceptre/cli/helpers.py
```diff
@@ -30,14 +29,14 @@ def catch_exceptions(func):
...
-        exits sceptre with a non-zero exit code.
+        re-raises.
         """
         try:
             return func(*args, **kwargs)
         except (SceptreException, BotoCoreError, ClientError, Boto3Error,
                 TemplateError) as error:
             write(error)
-            sys.exit(1)
+            raise
```
jfalkenstein, line 39: So I know I'm late to the game here, and this PR is a little old, but I have a couple thoughts here:
First:
> The main issue with this change is that users may well (and probably) do rely on status exit codes on the CLI. Removing the status code and replacing it with a dump of the stack trace isn't going to be a good solution.
@ngfgrant, I _believe_ blowing up with an exception _will_ result in a nonzero status code. Are you concerned that the code might not be 1?

Second:
@alexharv074 , have you considered using `traceback.print_exc()` instead of `write(error)` here? This would give you the full traceback and we could still have the `sys.exit(1)`

Finally: 
We could always have DEBUG_ERRORS environment variable or something like that which would reraise rather than exit with just the message. Has that been considered?

alexharv074: @jfalkenstein This seems like a good compromise to me.

sceptre/cli/helpers.py
```diff
@@ -36,8 +39,9 @@ def decorated(*args, **kwargs):
         try:
             return func(*args, **kwargs)
         except (SceptreException, BotoCoreError, ClientError, Boto3Error,
-                TemplateError) as error:
-            write(error)
+                TemplateError):
+            if "DEBUG_ERRORS" in os.environ:
```
jfalkenstein, line 43: This probably should be reflected in the docs if we're going to do this.

jfalkenstein: Also, I think other envs used by sceptre are prefixed by SCEPTRE_ or something like that. You should check into that.

sceptre/cli/helpers.py
```diff
@@ -36,8 +39,9 @@ def decorated(*args, **kwargs):
...
+                traceback.print_exc()
```
jfalkenstein, line 44: Why wouldn't we just always want to do this? I don't know why we'd want to buy the stack track ever? Whenever there IS an error... it'd always be useful to know where the error is coming from. That's just my own two cents.

craighurley: > Why wouldn't we just always want to do this?

@jfalkenstein The pattern has generally been to capture specific exceptions and print relative message in a normal run.  The original plan for this PR was to print the exception when `--debug` was present, and possibly also add logic to capture the original exception that was causing Alex pain in the source issue, like has been done for other exceptions.

@alexharv074 this PR was close with printing the exception when `--debug` was present, but it has drifted into printing the exception when a specific env var is set.  I would prefer to see the exception with printed with `--debug` rather than an env var, as IMO setting/unsetting the env var is more clunky.

alexharv074: @craighurley How would you make the logger available inside this decorator function?

alexharv074:
@craighurley how is this version?

PR #1003, opened 2021-04-11
use standard AWS ENV VARS

zaro0508:
Switch to using more standard AWS environment variables.


zaro0508:
@zxiiro, this seems like it should work but circleci logs show a bunch of `AccessDenied` 😞

zxiiro:
Seems like an authorization error rather than authentication?

"... is not authorized to perform: cloudformation:DescribeStacks on resource: ...."

zaro0508:
it's working so i'm going to leave as-is.

PR #1006, opened 2021-04-12
Use pre-commit instead of flake8

zaro0508:
Run pre-commit in the CI, which includes flake8 and other checks, instead of running
just flake8.  Move flake8 configurations from setup.cfg to .flake8 config file.


zaro0508:
could you please take a look @zxiiro?

Makefile
```diff
@@ -53,7 +53,7 @@ clean-test:
 	rm -f test-results.xml
 
 lint:
-	flake8 .
+	pre-commit run --all-files
```
zxiiro, line 56: Add ` --show-diff-on-failure` here so that when it does fail on something the pre-commit modifies it will print out the diffs making it easier to spot what the cause of the faillure was.

zxiiro (changes requested):

setup.cfg
```diff
@@ -24,15 +24,6 @@ universal = 1
...
 [flake8]
-exclude =
-	.git,
-	__pycache__,
-	build,
-	dist,
-	.tox,
-	venv
-max-complexity = 12
 per-file-ignores =
```
zaro0508, line 27: i couldn't figure out how to move this config to .pre-ommit-config.yaml, all my attempts will cause the linter to fail.  do you happen to know how to do it @zxiiro ?

Makefile
zaro0508: done

setup.cfg
zxiiro: Use an exclude block like this in pre-commit https://github.com/opendaylight/releng-builder/blob/master/.pre-commit-config.yaml#L27-L31 it only accepts regex if I recall so even the list must be in regex form.

tarkatronic:
There is one big problem I see with moving the config, and that is IDE integration. If I have my IDE set up to run `flake8` on save, I will no longer be able to see the settings, so I will receive different results when working, vs when I try to commit.

zxiiro:
> There is one big problem I see with moving the config, and that is IDE integration. If I have my IDE set up to run `flake8` on save, I will no longer be able to see the settings, so I will receive different results when working, vs when I try to commit.

If that's the case I think we can support both by moving the config to `.flake8` file (https://flake8.pycqa.org/en/latest/user/configuration.html). Since it's a standard config location pre-commit and IDEs should be able to pick it up.

tarkatronic (approved):

PR #1009, opened 2021-04-14
Set up a basic pyproject.toml for PEP-518 builds

tarkatronic:
This made a number of changes, working toward a singular goal: Implementing a `pyproject.toml` to allow for PEP-518 builds. This will allow for better testing in tox, the possibility of using other packaging systems, and this file is also the only one where config for `black` can be stored. In addition to that, a lot of projects are now allowing you to put their config into this file. Such as I have done with pytest!

A summary of my commits:

- Ignore .venv folders
- Don't force Python development versions
  - I removed the `.python-version` file as it was forcing developers to use `3.6-dev` and `3.7-dev` versions of Python which are very old and probably not good to be using at this point.
- Read the version from the file instead of importing it
  - With the way that PEP-518 builds work, the project is put into an isolated environment before being built. In that isolated environment, nothing is installed, so the `from sceptre import __version__` call was failing. Now, that file is scraped manually for the version.
- Upgrade to the latest pytest
  - This was 3 major versions behind, and this latest allows for placing config in `pyproject.toml`!
- Add a basic pyproject.toml & shift pytest config over

## PR Checklist

## Other Information


setup.py
```diff
@@ -37,9 +47,9 @@
 
 setup(
     name="sceptre",
-    version=__version__,
+    version=get_version("sceptre/__init__.py"),
```
zxiiro, line 50: On some projects I worked on we pull the version information in from Git tags:

https://github.com/lfit/releng-lftools/blob/master/docs/conf.py#L26-L31

Do we want to do something similar here so that we don't have to maintain a file that hardcodes a version number?

tarkatronic: Possibly, that is in fact one of the solutions (the last one) suggested here: https://packaging.python.org/guides/single-sourcing-package-version/ -- this was the first solution, and more of a quick fix, as opposed to trying to rework the whole versioning system right now.

One thing I would like to propose for the future is using `poetry`, which would completely change this, and the version number would then be kept in the `pyproject.toml`. But that's a whole separate discussion. 😄

zxiiro: Personally I prefer to infer the version number from `git tag` since then you can guarentee that the tag and the version number in the project files match. Otherwise you can get into a situation where pyproject.toml says 1 version and the git tag says a different version.

What I like about the pbr method is for tagged commits it uses the tag version. For versions after the last tagged commit it auto increments a dev version like:

- tag: 1.0.0
- tag+1: 1.0.0-dev0
- tag+2: 1.0.0-dev1
- tag: 2.0.0
- etc...

This way every single commit has a unique version number that's automatically created from previous tagged releases. No one needs to remember to bump the version in any files and release automation can happen through pushes of git tags.

tarkatronic: I definitely see the merit of that; my point being that I'm trying to limit the scope of this PR and not try to rethink the origin of the versioning at this point. I only modified the way in which it's retrieved, to not be dependent on a particular environment. I think there should definitely be separate discussions on how the versioning is ultimately managed.

zxiiro: Yeah totally. I didn't request changes to this PR and meant for the version thing to just be a comment in passing. We should definitely move that discussion out of here.

zxiiro (approved):

zaro0508 (approved):

PR #1014, opened 2021-04-19
[Resolves #1013] Fix virtual-hosted-style uri

### LINKED ISSUE #1013: Region missing from S3 Url
The S3 Url returned by the [upload_to_s3][1] method does not include a region. The AWS docs [suggest it should][2]. One consequence of this is that sceptre cannot be used with gov-cloud. 

[1]: https://github.com/Sceptre/sceptre/blob/v2.4.0/sceptre/template.py#L234-L235
[2]: https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#virtual-hosted-style-access

nicholasphair:
S3 virutal-hosted-style urls include the region name.
Currently, template.py only properly formats urls in
the China region. This commit adds the region to the url.
In doing so, gov cloud becomes supported.

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/template.py
+++ b/sceptre/template.py
@@ -207,7 +207,7 @@ class Template(object):
         # Remove any leading or trailing slashes the user may have added.
         bucket_name = self.s3_details["bucket_name"]
         bucket_key = self.s3_details["bucket_key"]
-        bucket_region = self.s3_details["bucket_region"]
+        bucket_region = self._bucket_region(bucket_name)
 
         self.logger.debug(
             "%s - Uploading template to: 's3://%s/%s'",
@@ -224,16 +224,9 @@ class Template(object):
             }
         )
 
-        china_regions = ["cn-north-1", "cn-northwest-1"]
-
-        if bucket_region in china_regions:
-            url = "https://{0}.s3.{1}.amazonaws.com.cn/{2}".format(
-                bucket_name, bucket_region, bucket_key
-            )
-        else:
-            url = "https://{0}.s3.amazonaws.com/{1}".format(
-                bucket_name, bucket_key
-            )
+        url = "https://{}.s3.{}.amazonaws.{}/{}".format(
+            bucket_name, bucket_region, self._domain_from_region(bucket_region), bucket_key
+        )
 
         self.logger.debug("%s - Template URL: '%s'", self.name, url)
 
@@ -318,6 +311,18 @@ class Template(object):
             return {"TemplateURL": url}
         else:
             return {"TemplateBody": self.body}
+
+    def _bucket_region(self, bucket_name):
+        region = self.connection_manager.call(
+            service="s3",
+            command="get_bucket_location",
+            kwargs={"Bucket": bucket_name}
+        ).get("LocationConstraint")
+        return region if region else "us-east-1"
+
+    @staticmethod
+    def _domain_from_region(region):
+        return "com.cn" if region.startswith("cn-") else "com"
 
     @staticmethod
     def _render_jinja_template(template_dir, filename, jinja_vars):
--- a/sceptre/config/reader.py
+++ b/sceptre/config/reader.py
@@ -454,16 +454,13 @@ class ConfigReader(object):
                 )
             ])
 
-            bucket_region = config.get("region", None)
-
             if "template_key_prefix" in config:
                 prefix = config["template_key_prefix"]
                 template_key = "/".join([prefix.strip("/"), template_key])
 
             s3_details = {
                 "bucket_name": config["template_bucket_name"],
-                "bucket_key": template_key,
-                "bucket_region": bucket_region
+                "bucket_key": template_key
             }
         return s3_details
 
--- a/tests/test_config_reader.py
+++ b/tests/test_config_reader.py
@@ -186,8 +186,7 @@ class TestConfigReader(object):
             },
             {
                 "bucket_name": "bucket-name",
-                "bucket_key": "prefix/name/2012-01-01-00-00-00-000000Z.json",
-                "bucket_region": "eu-west-1",
+                "bucket_key": "prefix/name/2012-01-01-00-00-00-000000Z.json"
             }
         ),
         (
@@ -198,8 +197,7 @@ class TestConfigReader(object):
             },
             {
                 "bucket_name": "bucket-name",
-                "bucket_key": "name/2012-01-01-00-00-00-000000Z.json",
-                "bucket_region": "eu-west-1",
+                "bucket_key": "name/2012-01-01-00-00-00-000000Z.json"
             }
         ),
         (
@@ -209,8 +207,7 @@ class TestConfigReader(object):
             },
             {
                 "bucket_name": "bucket-name",
-                "bucket_key": "name/2012-01-01-00-00-00-000000Z.json",
-                "bucket_region": None,
+                "bucket_key": "name/2012-01-01-00-00-00-000000Z.json"
             }
         ),
         (
--- a/tests/test_template.py
+++ b/tests/test_template.py
@@ -59,13 +59,20 @@ class TestTemplate(object):
         mock_bucket_exists.return_value = True
         self.template.s3_details = {
             "bucket_name": "bucket-name",
-            "bucket_key": "bucket-key",
-            "bucket_region": "eu-west-1"
+            "bucket_key": "bucket-key"
         }
 
         self.template.upload_to_s3()
 
-        self.template.connection_manager.call.assert_called_once_with(
+        get_bucket_location_call, put_object_call = self.template.connection_manager.call.call_args_list
+        get_bucket_location_call.assert_called_once_with(
+            service="s3",
+            command="get_bucket_location",
+            kwargs={
+                "Bucket": "bucket-name"
+            }
+        )
+        put_object_call.assert_called_once_with(
             service="s3",
             command="put_object",
             kwargs={
@@ -76,13 +83,17 @@ class TestTemplate(object):
             }
         )
 
+    def test_domain_from_region(self):
+        assert self.template._domain_from_region("us-east-1") == "com"
+        assert self.template._domain_from_region("cn-north-1") == "com.cn"
+        assert self.template._domain_from_region("cn-northwest-1") == "com.cn"
+
     def test_bucket_exists_with_bucket_that_exists(self):
         # connection_manager.call doesn't raise an exception, mimicing the
         # behaviour when head_bucket successfully executes.
         self.template.s3_details = {
             "bucket_name": "bucket-name",
-            "bucket_key": "bucket-key",
-            "bucket_region": "eu-west-1"
+            "bucket_key": "bucket-key"
         }
 
         assert self.template._bucket_exists() is True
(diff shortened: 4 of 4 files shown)


zaro0508 (approved):

zxiiro (approved):

zaro0508 (changes requested):
@nicholasphair integration test failed.. https://app.circleci.com/pipelines/github/Sceptre/sceptre/458/workflows/4674bdce-7b2f-4793-9255-de6935308af1/jobs/5595/tests#failed-test-0

zxiiro (approved):

nicholasphair:
I only looked at this briefly this weekend, but it seems the bucket region in the integration tests is not being reported properly.   

For all tests, the bucket (and bucket region) being used is always the one defined in the [config][1]. However, the region that gets reported in [template.py][2] is not that bucket's region. For a concrete example, in [this step][3], the bucket region is reported as `eu-west-2`, when in fact it is `eu-west-1`. The reported bucket region being misaligned with the actual buckets region is what causes the failed test. 

Now, I'm still learning the way these values get set so If someone more familiar quicklt knows the right fix I'm happy to implement it. Otherwise, I'll find some time this week to work through this.

Thanks!

[1]: https://github.com/Sceptre/sceptre/blob/v2.4.0/integration-tests/sceptre-project/config/config.yaml#L2
[2]: https://github.com/Sceptre/sceptre/blob/v2.4.0/sceptre/template.py#L208
[3]: https://github.com/Sceptre/sceptre/blob/v2.4.0/integration-tests/features/stack-output-resolver.feature#L28

nicholasphair:
Okay, I've believe got a better grasp on this.

[Here][1] is where the bucket region is set, and it is just set to whatever value is in the config. However, there is no guarantee that the region in the config is in fact the bucket region. And indeed, in our case, it is not. 

Instead, the value for that region should be looked up. Boto3 gives the [api call][2] needed to do just that.

So, my only question is where to actually make this call. The reader does not seem like the best place. @zxiiro, @zaro0508, any suggestions?

[1]: https://github.com/Sceptre/sceptre/blob/v2.4.0/sceptre/config/reader.py#L452
[2]: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.get_bucket_location

zaro0508:
Why do you think `The reader does not seem like the best place` @nicholasphair ?

The sceptre template bucket needs to exist before we ca request the bucket location with S3 API.   I believe sceptre will create an S3 bucket to store templates if one doesn't exist and that's why it's getting the region from the config.   Maybe we need to do the region lookup if bucket exists and fallback to region in config if bucket does not exist?

nicholasphair:
Well, I believe adding that logic to the reader would be out of scope. 

In it's current state, the `ConfigReader` simply parses the local config files into a `Stack` object. To add bucket lookup logic, would be to extend the class to take parsed values and query external apis for even more context. In fact, the reader does not even expose a `ConnectionManager` to make such requests, which maybe further hints that it shouldn't be done.

What I have done instead, is ripped out the `bucket_region` from the `s3_details`. Instead, I preferred to simply [lookup the region][1] when its time to write to the bucket. `bucket_region` is only ever used by the upload call, so it seemed like a natural fit to keep all the logic there. 

I have updated the commit and my local integration tests are passing. (Still trying to get circleci going on my fork though).


[1]: https://github.com/nicholasphair/sceptre/blob/s3_vhost_style/sceptre/template.py#L210

zaro0508:
I think there's something wrong with your latest update of this PR @nicholasphair.  Would you mind rebasing your changes onto the `latest` master branch and push again?

nicholasphair:
sure thing @zaro0508 , I just rebased.

sceptre/template.py
```diff
@@ -319,6 +312,17 @@ def get_boto_call_parameter(self):
         else:
             return {"TemplateBody": self.body}
 
+    def _bucket_region(self, bucket_name):
+        return self.connection_manager.call(
+            service="s3",
+            command="get_bucket_location",
+            kwargs={"Bucket": bucket_name}
+        ).get('LocationConstraint')
```
zaro0508, line 320: [aws docs](https://docs.aws.amazon.com/cli/latest/reference/s3api/get-bucket-location.html) says.. "Returns the Region the bucket resides in. **You set the bucket's Region using the LocationConstraint request parameter in a CreateBucket request**."  Therefore this request can return `None`,  it's doing that for all my buckets because I don't set `LocationConstraint` when i create buckets. 

```
aws s3api get-bucket-location --bucket bootstrap-awss3cloudformationbucket-l0axns15gl9h
None
```

This will cause sceptre to fail when setting `template_bucket_name: bootstrap-awss3cloudformationbucket-l0axns15gl9h` in config.yaml

<details>
<summary>sceptre run log</summary>

		[2021-05-01 08:45:15] - prod/noop - Launching Stack
		[2021-05-01 08:45:15] - No cloudformation client found, creating one...
		[2021-05-01 08:45:15] - Getting Boto3 session
		[2021-05-01 08:45:15] - No Boto3 session found, creating one...
		[2021-05-01 08:45:15] - Using cli credentials...
		[2021-05-01 08:45:15] - Using credential set from assume-role: {'AccessKeyId': '****************', 'SecretAccessKey': '************************************', 'Region': 'us-east-1'}
		[2021-05-01 08:45:15] - Boto3 session created
		[2021-05-01 08:45:16] - prod/noop - Stack is in the PENDING state
		[2021-05-01 08:45:16] - prod/noop - Creating Stack
		[2021-05-01 08:45:16] - noop - Uploading template to S3...
		[2021-05-01 08:45:16] - noop - Attempting to find template bucket 'bootstrap-awss3cloudformationbucket-l0axns15gl9h'
		[2021-05-01 08:45:16] - No s3 client found, creating one...
		[2021-05-01 08:45:16] - Getting Boto3 session
		[2021-05-01 08:45:16] - noop - Found template bucket 'bootstrap-awss3cloudformationbucket-l0axns15gl9h'
		[2021-05-01 08:45:17] - noop - Uploading template to: 's3://bootstrap-awss3cloudformationbucket-l0axns15gl9h/testing/prod/noop/2021-05-01-15-45-13-362578Z.json'
			File "/usr/local/var/pyenv/versions/py38-sceptredev/bin/sceptre", line 8, in <module>
				sys.exit(cli())
			File "/usr/local/var/pyenv/versions/3.8.6/envs/py38-sceptredev/lib/python3.8/site-packages/click/core.py", line 829, in __call__
			File "/usr/local/var/pyenv/versions/3.8.6/envs/py38-sceptredev/lib/python3.8/site-packages/click/core.py", line 782, in main
				rv = self.invoke(ctx)
			File "/usr/local/var/pyenv/versions/3.8.6/envs/py38-sceptredev/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
			File "/usr/local/var/pyenv/versions/3.8.6/envs/py38-sceptredev/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
			File "/usr/local/var/pyenv/versions/3.8.6/envs/py38-sceptredev/lib/python3.8/site-packages/click/core.py", line 610, in invoke
			File "/usr/local/var/pyenv/versions/3.8.6/envs/py38-sceptredev/lib/python3.8/site-packages/click/decorators.py", line 21, in new_func
			File "/usr/local/var/pyenv/versions/3.8.6/envs/py38-sceptredev/lib/python3.8/site-packages/sceptre/cli/helpers.py", line 37, in decorated
				return func(*args, **kwargs)
			File "/usr/local/var/pyenv/versions/3.8.6/envs/py38-sceptredev/lib/python3.8/site-packages/sceptre/cli/launch.py", line 38, in launch_command
				responses = plan.launch()
			File "/usr/local/var/pyenv/versions/3.8.6/envs/py38-sceptredev/lib/python3.8/site-packages/sceptre/plan/plan.py", line 130, in launch
				return self._execute(*args)
			File "/usr/local/var/pyenv/versions/3.8.6/envs/py38-sceptredev/lib/python3.8/site-packages/sceptre/plan/plan.py", line 40, in _execute
			File "/usr/local/var/pyenv/versions/3.8.6/envs/py38-sceptredev/lib/python3.8/site-packages/sceptre/plan/executor.py", line 55, in execute
			File "/usr/local/var/pyenv/versions/3.8.6/lib/python3.8/concurrent/futures/_base.py", line 432, in result
				return self.__get_result()
			File "/usr/local/var/pyenv/versions/3.8.6/lib/python3.8/concurrent/futures/_base.py", line 388, in __get_result
				raise self._exception
			File "/usr/local/var/pyenv/versions/3.8.6/lib/python3.8/concurrent/futures/thread.py", line 57, in run
			File "/usr/local/var/pyenv/versions/3.8.6/envs/py38-sceptredev/lib/python3.8/site-packages/sceptre/plan/executor.py", line 62, in _execute
			File "/usr/local/var/pyenv/versions/3.8.6/envs/py38-sceptredev/lib/python3.8/site-packages/sceptre/hooks/__init__.py", line 104, in decorated
			File "/usr/local/var/pyenv/versions/3.8.6/envs/py38-sceptredev/lib/python3.8/site-packages/sceptre/plan/actions.py", line 201, in launch
				status = self.create()
			File "/usr/local/var/pyenv/versions/3.8.6/envs/py38-sceptredev/lib/python3.8/site-packages/sceptre/hooks/__init__.py", line 104, in decorated
			File "/usr/local/var/pyenv/versions/3.8.6/envs/py38-sceptredev/lib/python3.8/site-packages/sceptre/plan/actions.py", line 74, in create
				self.stack.template.get_boto_call_parameter())
			File "/usr/local/var/pyenv/versions/3.8.6/envs/py38-sceptredev/lib/python3.8/site-packages/sceptre/template.py", line 310, in get_boto_call_parameter
				url = self.upload_to_s3()
			File "/usr/local/var/pyenv/versions/3.8.6/envs/py38-sceptredev/lib/python3.8/site-packages/sceptre/template.py", line 228, in upload_to_s3
				bucket_name, bucket_region, self._domain_from_region(bucket_region), bucket_key
			File "/usr/local/var/pyenv/versions/3.8.6/envs/py38-sceptredev/lib/python3.8/site-packages/sceptre/template.py", line 324, in _domain_from_region
				return "com.cn" if region.startswith("cn-") else "com"
		AttributeError: 'NoneType' object has no attribute 'startswith'

</details>

zaro0508 (changes requested):

sceptre/template.py
nicholasphair: Hi @zaro0508, good catch, you are right, the request can return None. Fortunately, that api call should still work for us. None/null is returned [when the bucket is in `us-east-1`][1], not specifically because you didn't set the constraint. Of course, since `us-east-1` is the [default][2] when creating buckets, it does lead to this problem. Looking at the logs, it appears you are using `us-east-1`, so that checks out.

I can update the pr to default to `us-east-1` to align with the (seemingly odd) boto behavior. Thanks again for finding this. 

[1]: https://docs.aws.amazon.com/cli/latest/reference/s3api/get-bucket-location.html#output
[2]: https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateBucket.html

nicholasphair: Sorry for the delay. I just updated the behavior to use `us-east-1` when the request returns None. Again, this is appears to be how to interpret a None return value as [per the docs][1]. 

[1]: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.get_bucket_location

zaro0508 (approved):

PR #1015, opened 2021-04-22
add cfn-lint to pre-commit checks

zaro0508:
Run cfn-lint[1] check with pre-ccommit to validate cloudformation
templates.

[1] https://github.com/aws-cloudformation/cfn-python-lint

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
     hooks:
       - id: yamllint
   - repo: https://github.com/awslabs/cfn-python-lint
-    rev: v0.48.3
+    rev: v0.47.2
     hooks:
       - id: cfn-python-lint
         files: .*\.(json|yml|yaml)$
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -8,8 +8,12 @@ repos:
...
   - repo: https://gitlab.com/pycqa/flake8
-    rev: 3.9.0
+    rev: 3.9.1
     hooks:
       - id: flake8
+  - repo: https://github.com/awslabs/cfn-python-lint
+    rev: v0.48.3
+    hooks:
+      - id: cfn-python-lint
+        files: .*\.(json|yml|yaml)$
```
zxiiro, line 19: Since we're using the `.cfnlintrc.yaml` file this is probably redundant here (maybe it will fix the circleci issue?). Other route is to move the ignore_templates to pre-commit's excludes instead. If I recall `pre-commit` passes one file at a time to the linter tool so filtering in the tool's config is redundant with what pre-commit is doing.

zxiiro (approved):

PR #1022, opened 2021-04-28
[Resolves #1005] Testing uplift

### LINKED ISSUE #1005: setup tests for python 3.8 and 3.9
Currently sceptre tests are run with python 3.6 & 3.7, https://github.com/Sceptre/sceptre/blob/master/tox.ini#L2 

We should also run tests with python 3.8 and possibly 3.9 so those can be supported as well.

tarkatronic:
This PR adds testing for Python 3.8 and 3.9 along with a little bit of an uplift in the general testing tooling.

* Upgraded tox
* Started using isolated/PEP-518 builds in tox
* Using the `pytest-cov` plugin now, so that testing and gathering coverage are consolidated into a single step. This should shorten CI runs since we're no longer running tests twice.
* Removed `pytest-runner`. The documentation itself recommends removing it (https://pypi.org/project/pytest-runner/)

## PR Checklist

## Other Information

Fixes #1005 


tarkatronic:
One note I forgot to add in my summary:

Previously, CI would theoretically fail if coverage fell below 92%. I had to lower that to 90%, as I was seeing only 90.11% coverage.

I suspect this is because I turned on branch coverage measurement (https://coverage.readthedocs.io/en/coverage-5.5/branch.html).

Makefile
```diff
@@ -56,22 +54,14 @@ lint:
...
 
-coverage-all:
-		coverage erase
-		coverage run --source sceptre -m pytest
-		coverage xml
-
-coverage: coverage-all
-		coverage report --show-missing --fail-under 92
-
 sonar:
```
zaro0508, line 65: hmm, isn't sonar for aggregating coverage reports?  Do you know anything about sonar @tarkatronic?  I've never used it before.

pyproject.toml
```diff
@@ -2,5 +2,25 @@
...
+  "if __name__ == .__main__.:", # Don't complain about mypy-specific code
+  "if TYPE_CHECKING:",
+]
+ignore_errors = true
+fail_under = 90.0
+show_missing = true
+
 [tool.pytest.ini_options]
-addopts = "tests/ --ignore=env/ --ignore=venv/ --junitxml=test-results/junit.xml -s"
+addopts = "-s --cov --cov-report term --cov-report term-missing --cov-report xml"
```
zaro0508, line 26: sorry, i don't really know how this toml file works.  does running tox automatically generated coverage reports?

tox.ini
```diff
@@ -1,9 +1,14 @@
 [tox]
-envlist = py36,py37
+minversion = 3.23.0
+isolated_build = True
+envlist = py{36,37,38,39},flake8
```
zaro0508, line 4: does this run flake8? if so do we really want tox to run flake8?

zaro0508 (changes requested):
There's some info in CONTRIBUTING.md file that's related to this change.  would you mind updating some of that info?

Makefile
tarkatronic: [Sonar](https://www.sonarqube.org/) is generally used for static analysis. I don't personally use it; I prefer using tools build specifically for and by the Python ecosystem including flake8, pylint, and maybe even bandit.

I wasn't sure I wanted to rip out a whole other part of the tooling yet, until I better understood how and why it's used.

pyproject.toml
tarkatronic: This toml file is just a way of configuring many tools in one place. This particular part is the equivalent of having a separate `pytest.ini` file. Many tools in the Python ecosystem have been adopting the ability to add configuration to this file. You can see this repo for a list of such tools: https://github.com/carlosperate/awesome-pyproject

As to the specific configuration listed here, this makes it so that any time `pytest` is run, it automatically generates a coverage report. And since `tox` runs `pytest` yes, that does generate a coverage report. 😄

tox.ini
tarkatronic: It does! And I think we should do both, for the same reason that we run it in both pre-commit and in CI. The more places, the better. Not all contributors will install `pre-commit` in their local environment, and not all contributors will run `tox`. But the more places we run these tools, the higher chance that they will get run at least once by contributors.

pyproject.toml
zaro0508: Any reason you didn't move configs in `tox.ini` into this  toml file?

tarkatronic: I could, I've just been hesitant to move that particular config yet in my projects. According to [the documentation](https://tox.readthedocs.io/en/latest/example/basic.html#pyproject-toml-tox-legacy-ini), it's a little bit of a hack at the moment, and intentionally so. Basically you have to write all of the config into a single big multi-line string, still using the INI format. They do have plans for native TOML config in the future, but it's just not there yet.

So it would look something like this:
```toml
[tool.tox]
legacy_tox_ini = """
[tox]
minversion = 3.23.0
isolated_build = True
envlist = py{36,37,38,39},flake8
skip_missing_interpreters = True

[testenv]
deps =
     -rrequirements/prod.txt
     -rrequirements/dev.txt
commands = pytest {posargs}

[testenv:flake8]
commands = flake8 .
"""
```
Maybe not the worst thing in the world. But my personal preference so far has been to wait until they adopt native TOML configuration. Especially when the tox files [can tend to grow a bit larger](https://github.com/godaddy/tartufo/blob/main/tox.ini).

zaro0508: ok, let's just leave as-is then.

zaro0508 (approved):

Makefile
zxiiro: Agreed, Sonar's not that great with Python, it's much more useful for Java projects and C/C++ with a paid subscription. I feel we got what Sonar provides sufficiently with pre-commit linters. Unless someone wants a nice UI and historical view of the data.

zxiiro (approved):

PR #1030, opened 2021-05-03
Remove Sceptre V1 docs

zaro0508:
Sceptre V1 EOL was on 06/01/2019.  Remove docs for V1


craighurley (approved):
Looks good.  I assume we are keeping the actual migraiton doc around _just-in-case_ ? 

https://github.com/sceptre/project/wiki/Migration-Guide:-V1-to-V2

zaro0508:
meh.. i nearly forgot wiki was there @craighurley. let's just just keep wiki as-is for now. it seems like it has some good info that we may want to move into CONTRIBUTIONS.md?

tarkatronic (approved):

PR #1033, opened 2021-05-07
Add .editorconfig

zxiiro:
Introduce an .editorconfig file so that we can sync common editor
configuration across developer systems. Developers can take advantage
of this file through a plugin for their preferred editor if it is
not already built-in.

More details at https://editorconfig.org/

Signed-off-by: Thanh Ha <zxiiro@gmail.com>

[Your PR description here]

## PR Checklist

## Other Information

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,26 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_style = space
+
+[*.json]
+indent_size = 2
+
+[*.{markdown,md}]
+indent_size = 4
+max_line_length = 80
+trim_trailing_whitespace = false
+
+[*.py]
+indent_size = 4
+max_line_legth = 120
+
+[*.{yaml,yml}]
+indent_size = 2
+
+[Makefile]
+indent_style = tab


.editorconfig
```diff
@@ -0,0 +1,21 @@
...
+[*.json]
+indent_size = 2
+
+[*.{markdown,md}]
+indent_size = 4
+max_line_length = 80
+
+[*.py]
+indent_size = 4
+max_line_legth = 79
```
tarkatronic, line 18: I think this is going to be the one sticking point here in this file. I know that PEP 8 recommends a line length of 79, but I think there is good reason not to follow that nowadays. The two main reasons I would argue against it in this particular case are current convention in the project, and `black` which I think we should aim to adopt. Currently, [flake8 is configured](https://github.com/Sceptre/sceptre/blob/8a805555886e160b5325e6260949e7e9c9fa9e89/.flake8#L15) to use a line length of 120. I'm okay with using that, but nowadays I tend to use the default from `black`, [which is 88](https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length).

.editorconfig
```diff
@@ -0,0 +1,21 @@
...
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_style = space
+
+[*.json]
+indent_size = 2
+
+[*.{markdown,md}]
+indent_size = 4
+max_line_length = 80
```
tarkatronic, line 14:
```suggestion
max_line_length = 80
trim_trailing_whitespace = false
```
In Markdown, trailing whitespace is significant. Specifically, two spaces at the end of the line indicate a line break.

tarkatronic:
I just remembered another very important piece to add in here!
```ini
[Makefile]
indent_style = tab
```

tarkatronic:
Oh and `charset = utf-8` at the top level. Sorry, I haven't had enough coffee yet this morning haha.

zxiiro:
> Oh and `charset = utf-8` at the top level. Sorry, I haven't had enough coffee yet this morning haha.

Updated to add charset and Makefile settings.

.editorconfig
```diff
@@ -0,0 +1,26 @@
...
+
+[*.py]
+indent_size = 4
+max_line_legth = 120
+
+[*.{yaml,yml}]
+indent_size = 2
+
+[Makefile]
+indent_style = tab
```
zaro0508, line 26: did you set this to tab to match existing indent?  should we change to spaces?

zaro0508 (changes requested):
some of these checks are already in `pre-commit-config.yaml`.  should we remove duplicates in pre-commit and have pre-commit execute checks from this file using https://github.com/editorconfig-checker/editorconfig-checker.python ?

.editorconfig
zxiiro: Yes, I didn't intend to change what was already there. Also I believe tabs are the standard syntax for Makefiles. Depending on which version of `make` you have changing it to spaces might actually break the Makefile for some folks so I'd recommend keeping it as tabs.

zxiiro:
> some of these checks are already in `pre-commit-config.yaml`. should we remove duplicates in pre-commit and have pre-commit execute checks from this file using https://github.com/editorconfig-checker/editorconfig-checker.python ?

Interesting I wasn't aware of this editorconfig plugin. I'll take a look and see if I can improve this PR to utilize it.

zxiiro:
> > some of these checks are already in `pre-commit-config.yaml`. should we remove duplicates in pre-commit and have pre-commit execute checks from this file using https://github.com/editorconfig-checker/editorconfig-checker.python ?
> 
> Interesting I wasn't aware of this editorconfig plugin. I'll take a look and see if I can improve this PR to utilize it.

Ok I just took at look at the editorconfig-checker linter and it fails a lot of Python and Markdown code that passes the flake8 linter we currently have. It would take a lot of effort to reformat all the Python and Markdown files. I think we should merge this patch as is and revisit adding editorconfig-checker after we implement python-black as I suspect black will solve most of the issues the checker is picking up.

.editorconfig
zaro0508: ok

zaro0508 (approved):

PR #1039, opened 2021-05-17
Include sceptre-file-resolver as a core resolver

zxiiro:
This includes the sceptre-file-resolver
(https://github.com/Sceptre/sceptre-file-resolver) into Sceptre as
a core resolver.

Signed-off-by: Thanh Ha <zxiiro@gmail.com>

## PR Checklist

## Other Information

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,6 +28,7 @@ install_requirements = [
     "Jinja2>=2.8,<3",
     "colorama>=0.3.9",
     "packaging>=16.8,<17.0",
+    "sceptre-file-resolver>=1.0.3,<2",
     "six>=1.11.0,<2.0.0",
     "networkx>=2.4,<2.6"
 ]
--- a/docs/_source/docs/resolvers.rst
+++ b/docs/_source/docs/resolvers.rst
@@ -39,8 +39,15 @@ Example:
    parameters:
        database_password: !environment_variable DATABASE_PASSWORD
 
+file
+~~~~
+
+Refer to `sceptre-file-resolver <https://github.com/Sceptre/sceptre-file-resolver/>`_ for documentation.
+
 file_contents
 ~~~~~~~~~~~~~
+
+**deprecated**: Consider using the `file`_ resolver instead.
 
 Reads in the contents of a file.
 


zxiiro:
@zaro0508 not sure what you meant by replacing the !file_contents resolver but here I replaced it with the code from the !file resolver. Considering the 2 plugins are pretty similar with the latter being more advanced I would consider this an upgrade to the file_contents resolver.

I think for the most part this should just work for people already using the !file_contents resolver but maybe the parsing of JSON/YAML parsing will come at a surprise to those using this plugin.

sceptre/resolvers/file.py
```diff
@@ -0,0 +1,81 @@
+# -*- coding: utf-8 -*-
+
+import os
+import json
+import yaml
+import requests
```
zaro0508, line 6: is requests in the python 3.6 standard library?

zaro0508 (changes requested):
To add a new plugin we also need to add an entrypoint in `setup.py` and we will want to move the docs to `docs/_source/docs/resolvers.rst`.   Also we want to deprecate the  `!file_contents` resolver and tell everyone to use this new resolver instead.  Could you please add some documentation for that?

sceptre/resolvers/file.py
zxiiro: No it's a dependency.

zxiiro:
> To add a new plugin we also need to add an entrypoint in `setup.py` and we will want to move the docs to `docs/_source/docs/resolvers.rst`. Also we want to deprecate the `!file_contents` resolver and tell everyone to use this new resolver instead. Could you please add some documentation for that?

Sure, I'll look into that.

craighurley:
Is the functionality of this existing `!file_contents` resolver identical to a subset of of funcationality in this PRs `!file` resolver?  If so, should this PR be treated as an upgrade to the existing `!file_contents` resolver?  Then there would be no need to deprecate `!file_contents`.

zxiiro:
> Is the functionality of the existing `!file_contents` resolver identical to a subset of of funcationality in this PRs `!file` resolver? If so, should this PR be treated as an upgrade to the existing `!file_contents` resolver? Then there would be no need to deprecate `!file_contents`.

My first implementation in this PR was to upgrade to the functionality of the existing `!file_contents` resolver but I changed it to introduce a new resolver due to @zaro0508's request. I'm happy to re-patch this PR in either direction if the 2 of you can get together and hash it out and get an agreement on which direction we go.

zxiiro:
This has been updated to include the file-resolver by installing it as part of requirements.txt rather than copying the code over from the other repo. I'm not sure what we want to do about the docs. I think keeping it in this repo's docs makes it more official so people can read about it directly in the documentation. If we instead link to the README it feels less official to me. Thoughts?

zaro0508:
i prefer a doc reference otherwise doc changes would need to happen in two repos, correct?

zxiiro:
Updated to reference the GitHub repo for the sceptre-file-resolver for the documentation.

setup.py
```diff
@@ -28,6 +28,7 @@ def get_version(rel_path):
     "Jinja2>=2.8,<3",
     "colorama>=0.3.9",
     "packaging>=16.8,<17.0",
+    "sceptre-file-resolver>=1.0.1,<2",
```
zaro0508, line 31: latest version is v1.0.3  https://github.com/Sceptre/sceptre-file-resolver/tags

zaro0508 (changes requested):

zaro0508:
@zxiiro did you forget to push your latest changes?  the diff still shows doc and code additions instead of just making the resolver a dependency.

zxiiro:
> @zxiiro did you forget to push your latest changes? the diff still shows doc and code additions instead of just making the resolver a dependency.

Indeed you are right. I'm not sure how that happened. I just pushed my code up.

zaro0508 (approved):

PR #1043, opened 2021-05-20
[Resolves #1042] Optimize Sceptre Start Time by only Processing Dependent Configuration Files

### LINKED ISSUE #1042: Sceptre Commands Slow to Start with a Large Number of Configuration Files
### Subject of the issue

When a large number of Sceptre configuration yaml files exist within the Sceptre config directory, running any sceptre command is very slow to start. For example, we have over 2,500 Sceptre configuration files and it now takes approx 3 minutes to run any sceptre command up to the point of performing an action.

This is because when any sceptre command is run specifying a subdirectory path, all configuration files are processed within the config directory rather than only the configuration files existing in the specified path and their dependencies.

This also causes the issue whereby a syntax error in any configuration file within the config directory will cause sceptre to fail, even if that configuration file is not a dependency of the configurations in the specified path.

### Your environment

* Sceptre version: 2.4.0
* python version: 3.8.3
* OS: Mac/Windows

### Steps to reproduce

For example, run `sceptre --debug status` against any individual configuration file, and you will see all configuration files processed within the debug output.

### Expected behaviour

Only configuration files specified on the sceptre command line and their dependencies should be processed.

### Actual behaviour

All configuration files are processed.

lucasvieirasilva:
Fix for: #1042 

This changes the `ConfigReader` to always use the `full_command_path` instead of using `full_config_path` then add the stack dependencies to the `todo` set to be processed.

## Evidences

I tested locally in my environment with over about 2,500 Sceptre configuration files.

**Current version:**
image

**Changed version:**
image

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/config/reader.py
+++ b/sceptre/config/reader.py
@@ -230,12 +230,9 @@ class ConfigReader(object):
                 if not path.exists(full_dep):
                     raise DependencyDoesNotExistError(
                             "{stackname}: Dependency {dep} not found. "
-                            "Valid dependency names are: "
-                            "{stackkeys}. "
                             "Please make sure that your dependencies stack_outputs "
                             "have their full path from `config` defined."
-                            .format(stackname=stack.name, dep=dep,
-                                    stackkeys=", ".join(stack_map.keys())))
+                            .format(stackname=stack.name, dep=dep))
 
                 if full_dep not in full_todo and full_dep not in deps_todo:
                     todo.add(full_dep)
--- a/sceptre/config/reader.py
+++ b/sceptre/config/reader.py
@@ -195,10 +195,7 @@ class ConfigReader(object):
         """
         stack_map = {}
         command_stacks = set()
-        if self.context.ignore_dependencies:
-            root = self.context.full_command_path()
-        else:
-            root = self.context.full_config_path()
+        root = self.context.full_command_path()
 
         if path.isfile(root):
             todo = {root}
@@ -212,6 +209,8 @@ class ConfigReader(object):
                     todo.add(path.join(directory_name, filename))
 
         stack_group_configs = {}
+        full_todo = todo.copy()
+        deps_todo = set()
 
         while todo:
             abs_path = todo.pop()
@@ -226,6 +225,22 @@ class ConfigReader(object):
                     self.read(path.join(directory, self.context.config_file))
 
             stack = self._construct_stack(rel_path, stack_group_config)
+            for dep in stack.dependencies:
+                full_dep = path.join(self.context.full_config_path(), dep)
+                if not path.exists(full_dep):
+                    raise DependencyDoesNotExistError(
+                            "{stackname}: Dependency {dep} not found. "
+                            "Valid dependency names are: "
+                            "{stackkeys}. "
+                            "Please make sure that your dependencies stack_outputs "
+                            "have their full path from `config` defined."
+                            .format(stackname=stack.name, dep=dep,
+                                    stackkeys=", ".join(stack_map.keys())))
+
+                if full_dep not in full_todo and full_dep not in deps_todo:
+                    todo.add(full_dep)
+                    deps_todo.add(full_dep)
+
             stack_map[sceptreise_path(rel_path)] = stack
 
             full_command_path = self.context.full_command_path()
--- a/tests/test_config_reader.py
+++ b/tests/test_config_reader.py
@@ -299,25 +299,25 @@ class TestConfigReader(object):
         (
             "Abd",
             ["Abc/1.yaml", "Abd/1.yaml"],
-            {"Abd/1", "Abc/1"},
+            {"Abd/1"},
             {"Abd/1"}
         ),
         (
             "Abd",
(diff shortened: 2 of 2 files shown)


zxiiro (approved):
I downloaded and tested this and can confirm running `sceptre status` on my own repos appear to run faster than without this patch.

lucasvieirasilva:
> Loving the change (we have hundreds of Sceptre stack configs). However, would it be better to make such behaviour optional via CLI param? The reason is currently it validates whole config/* directory which is useful in case you've changed many files, but deploying one of them which does not depends on other changed files, and in case of a typo/etc anywhere it will let you know by failing.
> 
> Another option will be making this behaviour default and adding CLI param to validate whole config folder (or vice versa).
> @zaro0508 @lucasvieirasilva @zxiiro thoughts?

Hi @ykhalyavin thanks for your feedback, in my point of view the syntax validation should continue the responsibility of the `validate` command (https://sceptre.cloudreach.com/2.5.0/docs/cli.html#sceptre-validate), there you can specify a more generic command path.

(e.g `sceptre validate dev` in command in my point of view should be used in a different stage of your CI/CD maybe in PR validation)
(e.g `sceptre launch dev/dns/zone.yaml` and this one to actually apply the changes that you want to apply to the environment, so you already have checked the syntax of the changed files before, you don't need to that again in this command)

ykhalyavin:
> Hi @ykhalyavin thanks for your feedback, in my point of view the syntax validation should continue the responsibility of the `validate` command (https://sceptre.cloudreach.com/2.5.0/docs/cli.html#sceptre-validate), there you can specify a more generic command path.

@lucasvieirasilva 
My thouhgs were about validation of config/* files (proper jinja2 rendering with all variables available, validation of the required properties being in place, etc)

sceptre/config/reader.py
```diff
@@ -226,6 +225,22 @@ def construct_stacks(self):
...
+                full_dep = path.join(self.context.full_config_path(), dep)
+                if not path.exists(full_dep):
+                    raise DependencyDoesNotExistError(
+                            "{stackname}: Dependency {dep} not found. "
+                            "Valid dependency names are: "
+                            "{stackkeys}. "
+                            "Please make sure that your dependencies stack_outputs "
+                            "have their full path from `config` defined."
+                            .format(stackname=stack.name, dep=dep,
+                                    stackkeys=", ".join(stack_map.keys())))
```
zaro0508, line 238: `stackkeys` will be empty if user passes in a single file...

```
➜ sceptre status develop/essentials.yaml
"develop/essentials: Dependency develop/bootstrap2.yaml not found. Valid dependency names are: . Please make sure that your dependencies stack_outputs have their full path from `config` defined."
```

maybe we should exclude the `Valid dependency names are: ` for this condition?

zaro0508 (changes requested):
This PR generally LGTM,  I just have a minor concern.  In regards to the commands I general  agree with @lucasvieirasilva.  Sceptre already has the status, validate, and generate commands that can be used to validate the all Sceptre config files before deployment (or launch).

sceptre/config/reader.py
lucasvieirasilva: Yes, I agree, I've removed this statement from the message

zaro0508 (approved):

ngfgrant:
I agree that this change would be safer as an optional argument. From just having a quick look at the code this is probably quite a significant change. I would be nervous about introducing it as the default behaviour as I suspect it will have unintended consequences.

zaro0508:
> I agree that this change would be safer as an optional argument. From just having a quick look at the code this is probably quite a significant change. I would be nervous about introducing it as the default behaviour as I suspect it will have unintended consequences.

I've been using sceptre with this change on my projects and it seems fine to me.  I prefer this as the sceptre default behaviour so how about we get a few more users to confirm that it works fine on their projects before merging?

ykhalyavin:
> > I agree that this change would be safer as an optional argument. From just having a quick look at the code this is probably quite a significant change. I would be nervous about introducing it as the default behaviour as I suspect it will have unintended consequences.
> 
> I've been using sceptre with this change on my projects and it seems fine to me. I prefer this as the sceptre default behaviour so how about we get a few more users to confirm that it works fine on their projects before merging?

I can try it on my dev machine and in non prod environment starting today and write back later (Mon-Tue?) if any issues arised under _our normal usage_

lucasvieirasilva:
> I've been using sceptre with this change on my projects and it seems fine to me. I prefer this as the sceptre default behaviour so how about we get a few more users to confirm that it works fine on their projects before merging?

I've been using this version of sceptre in all my projects since I've raised this PR, my colleagues have been using this few days ago as well, until now, no issues.

lloydpick:
I've also been using it on one of our largest projects (1100 stacks), had no issues so far.

michal-sujkowski:
I have been using the @lucasvieirasilva version for more than 2 weeks on my project with over 2k stacks. It is much faster than the original command (way more than 10 times faster). I would rather keep the @lucasvieirasilva version as a default behaviour and have an option to run a full validation with `sceptre validate`.

PR #1057, opened 2021-06-11
[Resolves  #961]: Hook.stack is not necessarily the stack actually being updated

### LINKED ISSUE #961: Stack attribute in sceptre.hooks.Hook is not thread-safe
I am using a custom Hook because I want to report the start and end time of updating our stacks. I also want to know the name of the Stack for which the Hook has been called. According to the documentation this is done with `self.stack.name`. See this simplified example:

```python
class CustomHook(Hook):
    def __init__(self, *args, **kwargs):
        super(CustomHook, self).__init__(*args, **kwargs)

    def run(self):
        print(f'Hook: {self.stack.name}')
```

The content of `self.stack` seems to be overwritten by other concurrent running updates, probably because all use the same `CustomHook`-instance. So when I call `self.stack.name` I am not getting the information of stack for which the Hook has been called, but the information of the latest stack that has been assigned to the `CustomHook`-instance.

To clarify this, I have added a print-statement to the `add_stack_hooks(func)` in `hooks.__init__`:
```python
def add_stack_hooks(func):
    """
    A function decorator to trigger the before and after hooks, relative
    to the decorated function's name.
    :param func: a function that operates on a stack
    :type func: function
    """
    @wraps(func)
    def decorated(self, *args, **kwargs):
        print(f'before_{func.__name__}: {self.stack.name} {self.stack.hooks}')
        execute_hooks(self.stack.hooks.get("before_" + func.__name__))
        execute_hooks(self.stack.hooks.get("after_" + func.__name__))

        return response

    return decorated
```

I have the following in my `config/config.yaml`:
```yaml
profile: [....]
project_code: [....]
hooks:
  before_update:
    - !stack_deploy_time "deploy_start"
  after_update:
    - !stack_deploy_time "deploy_end"
```
The output will be something like this (I have sanitized it to be readable):
```
before_update: d/e/f {'before_update': [<hook.hook.CustomHook object at 0x7fdb5ffe2ed0>]} Hook: a/b/c
before_update: x/y/z {'before_update': [<hook.hook.CustomHook object at 0x7fdb5ff09e90>]} Hook: x/y/z
before_update: g/h/i {'before_update': [<hook.hook.CustomHook object at 0x7fdb5ffe2ed0>]} Hook: a/b/c
before_update: j/k/l {'before_update': [<hook.hook.CustomHook object at 0x7fdb5ffe2ed0>]} Hook: a/b/c
before_update: m/n/o {'before_update': [<hook.hook.CustomHook object at 0x7fdb5ffe2ed0>]} Hook: a/b/c
```
Only one Hook has the correct `stack` available (the `x/y/z`). The other four hooks all have `a/b/c` as `stack`, which is different than the one being updated. Note how these four all use the same hook-instance, which probably causes this problem.

Because the correct stack is available in the `add_stack_hooks`-decorator, I suggest adding the `stack` as parameter to the `Hook.run` method. The `stack` in the hook is than always the same as the one in the decorator. Also, sceptre can still get around with a limited amount of hook-instances. This would also mean that `Hook.stack` is no longer needed:

```python
import abc
import logging
from functools import wraps

from sceptre.helpers import _call_func_on_values


class Hook(object):
    """
    Hook is an abstract base class that should be inherited by all hooks.

    :param argument: The argument of the hook.
    :type argument: str
    """
    __metaclass__ = abc.ABCMeta

    def __init__(self, argument=None):
        self.logger = logging.getLogger(__name__)
        self.argument = argument

    def setup(self):
        """
        setup is a method that may be overwritten by inheriting classes. Allows
        hooks to run so initalisation steps when config is first read.
        """
        pass  # pragma: no cover

    @abc.abstractmethod
    def run(self, stack):
        """
        run is an abstract method which must be overwritten by all
        inheriting classes. Run should execute the logic of the hook.

        :param stack: The associated stack of the hook.
        :type stack: sceptre.stack.Stack
        """
        pass  # pragma: no cover


class HookProperty(object):
    [... as is ...]


def execute_hooks(hooks, stack):
    """
    Searches through dictionary or list for Resolver objects and replaces
    them with the resolved value. Supports nested dictionaries and lists.
    Does not detect Resolver objects used as keys in dictionaries.

    :param attr: A complex data structure to search through.
    :type attr: dict or list
    :param stack: The associated stack of the hook.
    :type stack: sceptre.stack.Stack
    :return: A complex data structure without Resolver objects.
    :rtype: dict or list
    """
    if isinstance(hooks, list):
        for hook in hooks:
            if isinstance(hook, Hook):
                hook.run(stack)


def add_stack_hooks(func):
    """
    A function decorator to trigger the before and after hooks, relative
    to the decorated function's name.
    :param func: a function that operates on a stack
    :type func: function
    """
    @wraps(func)
    def decorated(self, *args, **kwargs):
        execute_hooks(self.stack.hooks.get("before_" + func.__name__), self.stack)
        execute_hooks(self.stack.hooks.get("after_" + func.__name__), self.stack)

        return response

    return decorated

```

gilleswijnker:
[Resolves #961]: Hook.stack is not necessarily the stack actually being updated.

Please note this request introduces a breaking change for custom Hooks. `Hook.stack` has been removed. `Hook.run` now accepts a single parameter `stack`.

## PR Checklist

- [ X ] Wrote a good commit message & description [see guide below].
- [ X ] Commit message starts with `[Resolve #issue-number]`.
- [ X ] Added/Updated unit tests.
- [ NA ] Added/Updated integration tests (if applicable).
- [ X ] All unit tests (`make test`) are passing.
- [ X ] Used the same coding conventions as the rest of the project.
- [ X ] The new code passes pre-commit validations (`pre-commit run --all-files`).
- [ X ] The PR relates to _only_ one subject with a clear title.

## Other Information

### DIFF
--- a/docs/_source/docs/hooks.rst
+++ b/docs/_source/docs/hooks.rst
@@ -116,10 +116,10 @@ integrate additional functionality into Sceptre projects.
 A hook is a Python class which inherits from abstract base class ``Hook`` found
 in the ``sceptre.hooks module``.
 
-Hooks are require to implement a ``run()`` function that takes no parameters
-and to call the base class initializer.
+Hooks are require to implement a ``run(stack)`` function that takes a single parameter
+``stack`` and to call the base class initializer.
 
-Hooks may have access to ``argument``, and ``stack`` as object attributes. For example ``self.stack``.
+Hooks may have access to the ``argument`` object attribute. For example ``self.argument``.
 
 Sceptre uses the ``sceptre.hooks`` entry point to locate hook classes. Your
 custom hook can be written anywhere and is installed as Python package.
@@ -152,26 +152,28 @@ custom_hook.py
         argument: str
             The argument is available from the base class and contains the
             argument defined in the Sceptre config file (see below)
-        stack: sceptre.stack.Stack
-             The associated stack of the hook.
-        connection_manager: sceptre.connection_manager.ConnectionManager
-            Boto3 Connection Manager - can be used to call boto3 api.
 
         """
         def __init__(self, *args, **kwargs):
             super(CustomHook, self).__init__(*args, **kwargs)
 
-        def run(self):
+        def run(self, stack):
             """
             run is the method called by Sceptre. It should carry out the work
             intended by this hook.
+
+            Parameters
+            ----------
+            stack: sceptre.stack.Stack
+                The associated stack of the hook.
+            stack.connection_manager: sceptre.connection_manager.ConnectionManager
+                Boto3 Connection Manager - can be used to call boto3 api.
 
             To use instance attribute self.<attribute_name>.
 
             Examples
             --------
             self.argument
-            self.stack_config
 
             """
             print(self.argument)
--- a/sceptre/hooks/__init__.py
+++ b/sceptre/hooks/__init__.py
@@ -11,15 +11,12 @@ class Hook(object):
 
     :param argument: The argument of the hook.
     :type argument: str
-    :param stack: The associated stack of the hook.
-    :type stack: sceptre.stack.Stack
     """
     __metaclass__ = abc.ABCMeta
 
-    def __init__(self, argument=None, stack=None):
+    def __init__(self, argument=None):
         self.logger = logging.getLogger(__name__)
         self.argument = argument
-        self.stack = stack
 
     def setup(self):
         """
@@ -29,10 +26,13 @@ class Hook(object):
         pass  # pragma: no cover
 
     @abc.abstractmethod
-    def run(self):
+    def run(self, stack):
         """
         run is an abstract method which must be overwritten by all
         inheriting classes. Run should execute the logic of the hook.
+
+        :param stack: The associated stack of the hook.
+        :type stack: sceptre.stack.Stack
         """
         pass  # pragma: no cover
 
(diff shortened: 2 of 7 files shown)


ykhalyavin:
@zaro0508 could you please review? In our work we have some hooks where we have to execute certain boto3 calls after the stack A is deployed and we need to pass to our hook X outputs from the stack A (work with TGW) and this PR straightens our workarounds

docs/_source/docs/hooks.rst
```diff
@@ -152,26 +152,28 @@ custom_hook.py
...
-             The associated stack of the hook.
-        connection_manager: sceptre.connection_manager.ConnectionManager
-            Boto3 Connection Manager - can be used to call boto3 api.
 
         """
         def __init__(self, *args, **kwargs):
             super(CustomHook, self).__init__(*args, **kwargs)
 
-        def run(self):
+        def run(self, stack):
```
zaro0508, line 160: thanks for the contribution however I don't think we can break every Sceptre custom hook out in the wild.  I would prefer a more general solution that would allow hooks or resolvers to access the stack object.  

If this is the only solution then i think it would need to be done with a new interface and then deprecating the old one.

zaro0508 (changes requested):

alexharv074:
@gilleswijnker @ykhalyavin Is it ok for me to close this PR?

PR #1059, opened 2021-06-20
[Resolves #715] Support for Jsonnet templates

### LINKED ISSUE #715: Idea: Add ability for custom template filetypes
Sceptre comes with out-of-the-box support for json, yaml, Jinja2, and Troposphere based templates.

I'd like to register the `.jsonnet` extension for rendering [Jsonnet](https://jsonnet.org) templates before upload.

But rather than make a PR to add Jsonnet rendering and indirectly force the sceptre developers to support this filetype indefinitely, why not make template rendering customizable like hooks and resolvers? That way, the Jsonnet renderer can be developed and maintained independently.

Template rendering is already nicely contained in the [Template#body](https://github.com/cloudreach/sceptre/blob/master/sceptre/template.py#L67-L95) method. I could make a PR to call out to custom renderers based on the template extension, and refactor the existing filetype rendering logic to their own classes so they can exist as builtin renderers.

@ngfgrant Do you think custom template renderer support is inline with Sceptre's unopinionated philosophy? How likely would you approve a PR that adds, tests, and documents this functionality?

For me, adding support for custom renderers means I can add support for Jsonnet rendering to my project without having to fork Sceptre, and I very much want to avoid forking if I can.

rubendg:
#715 Support for Jsonnet templates

This PR adds support for templates written in [Jsonnet](https://jsonnet.org/). In particular, through the [Google Go Jsonnet implementation](https://github.com/google/go-jsonnet) and its Python bindings: [gojsonnet](https://pypi.org/project/gojsonnet/).

Details:

The implementation follows the structure of the existing template implementations. One difference from the other template implementations is that next to `sceptre_user_data`, the stack parameters are also made available. This gives the template writer direct access to the stack parameters instead of having to rely solely on the CloudFormation intrinsic functions for variable binding and manipulation.

Concerning the Jsonnet integration into Sceptre, there are a couple of implementations to consider:
the original [Google C++ implementation](https://github.com/google/jsonnet), the newer and faster [Google Go implementation](https://github.com/google/go-jsonnet), and a [JVM based implementation](https://github.com/databricks/sjsonnet) by Databricks. 

AFAIK there is currently no direct way to delay binding to a particular template implementation. Hence I've decided to integrate with the Go implementation through its [Python bindings](https://pypi.org/project/gojsonnet/), which to me seems the next best thing after the C++ client.

Data from outside the template is provided through the top-level arguments construct instead of external arguments - the difference is explained [here](https://jsonnet.org/learning/tutorial.html#parameterize-entire-config). This decision follows the recommendation from the Jsonnet docs:

> Generally, top-level arguments are the safer and easier way to parameterize an entire config, because the variables are not global and it's clear what parts of the config are dependent on their environment.

## PR Checklist

## Other Information


rubendg:
Currently the build fails because installing the `gojsonnet` package requires `go` to be installed.

rubendg:
Hi @zaro0508 thanks for you reply. I agree that sounds better. When I have time I'll migrate the changes and let you know.

Issue #937, opened 2020-09-28
Jinja variables cached between StackGroups


NathanWilliams:
While reading config files, Sceptre caches the `templating_vars` between StackGroups (in `config/reader.py`).

If a variable is omitted in one stack group's config, the incorrect value from another group is used.
(Note: Sceptre uses a `set` for its file list, which is non-deterministic, as such the following bug will only be seen on some runs)

I have a minimal example of the bug here:
https://github.com/NathanWilliams/sceptrebug

I'm working on a PR to fix it, and will update this ticket when it is ready.

PR #1083, opened 2021-08-17
[Resolves #813] Fix nested config Jinja render

### LINKED ISSUE #813: Variable is undefined
When defining the same variable in the config.yaml in different levels, causes Sceptre to intermittently return an error: <VARIABLE_NAME> is undefined. 
Usually if I run again it works, but sometimes I have to run a few times. This is in the scope of creating a changeset.

e.g of structure
-config
  - dev
     - config.yaml (variable : x: a)
     - appb
         - config.yaml (variable : x: b)   

The error is happening in Sceptre 2.1.0 and 2.1.5 (I havent' tested other versions, but I don't see this issue reported nor fixed in 2.2.0).

lucasvieirasilva:
This PR fixes nested `config.yaml` jinja substitutions (e.g `{{ someKey }}`).

Fixes issue: https://github.com/Sceptre/sceptre/issues/813

#### Scenario

Currently, when the project structure has multiple nested config files and one of them tries to use a property from another config file sometimes works, and sometimes not, when not work the error message is `<prop> is undefined.`.

Example structure:

```text
root/
  config/
    prod/
      A.yaml
      config.yaml
      2/
        B.yaml
        config.yaml
        3/
          C.yaml
          config.yaml
```

`config/prod/config.yaml`
```yaml
keyA: A
```

`config/prod/2/config.yaml`
```yaml
keyB: {{ keyA }}-B
```

`config/prod/2/3/config.yaml`
```yaml
keyA: {{ keyB }}-C
```

`keyB` should render the value `A-B` and `keyC` should render the value `A-B-C`

When running `sceptre launch prod` command the `ConfigReader` class doesn't follow a correct order to process the files, and sometimes the config object already has all the properties, sometimes not.

#### Solution

Sceptre already has a concept of recursive read in the `ConfigReader` class that reads all the `config.yaml` files recursively based on the parent path, and already has the config `dict` object with all the properties in the correct place, but when call the `_render` function just pass the `stack_group_config` that only contains the processed `stack_group_config`, which is fine when the files follow the correct order to be processed.

However, if the `ConfigReader` combines the current recursive object with the `stack_group_config` the `_render` will have all the required properties, doesn't matter which file order will be used to process the stacks.

## PR Checklist

## Other Information

### DIFF
--- a/integration-tests/features/create-stack.feature
+++ b/integration-tests/features/create-stack.feature
@@ -48,3 +48,14 @@ Feature: Create stack
     and the template for stack "10/A" is "sam_template.yaml"
     When the user creates stack "10/A"
     Then stack "10/A" exists in "CREATE_COMPLETE" state
+
+  Scenario: create new stack with nested config jinja resolver
+    Given stack_group "12/1" does not exist
+    When the user launches stack_group "12/1"
+    Then all the stacks in stack_group "12/1" are in "CREATE_COMPLETE"
+    and stack "12/1/A" has "Project" tag with "A" value
+    and stack "12/1/A" has "Key" tag with "A" value
+    and stack "12/1/2/B" has "Project" tag with "B" value
+    and stack "12/1/2/B" has "Key" tag with "A-B" value
+    and stack "12/1/2/3/C" has "Project" tag with "C" value
+    and stack "12/1/2/3/C" has "Key" tag with "A-B-C" value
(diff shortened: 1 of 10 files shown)


zaro0508:
Does this also fix issue #937 ?

lucasvieirasilva:
> Does this also fix issue #937 ?

No, it is a different issue.

zaro0508 (approved):

PR #1089, opened 2021-08-30
 [Resolves #1087] Add YAML document markers to template body

mrowlingfox:
This changes adds the `---` document marker to the beginning of each template body if it doesn't already exist.
It also updates the file upload file ext from `json` `yaml`
In the template tests `json.loads` is replaced with `yaml.safe_loads` as they now always return yaml

The reasoning behind this change can be found here https://github.com/Sceptre/sceptre/issues/1087

## PR Checklist

- [x] Wrote a good commit message & description [see guide below].
- [x] Commit message starts with `[Resolve #issue-number]`.
- [x] Added/Updated unit tests.
- [ ] Added/Updated integration tests (if applicable).
- [x] All unit tests (`make test`) are passing.
- [x] Used the same coding conventions as the rest of the project.
- [x] The new code passes pre-commit validations (`pre-commit run --all-files`).
- [x] The PR relates to _only_ one subject with a clear title.
      and description in grammatically correct, complete sentences.

## Approver/Reviewer Checklist

- [ ] Before merge squash related commits.

## Other Information

[Guide to writing a good commit](http://chris.beams.io/posts/git-commit/)

PR #1091, opened 2021-08-31
[Resolves #1087] Add YAML doc markers in generate

### LINKED ISSUE #1087: Add YAML document markers to template body
# Proposal
Add Yaml start document markers to the template body.

Initially, I ran into this lack of markers when using the `generate` command. I was trying to process the output programmatically and ran into issues trying to separate the documents from multiple stacks.

This is 100% compliant with [YAML Ain’t Markup Language (YAML™) Version 1.2 Document Markers](https://yaml.org/spec/1.2/spec.html#id2800401)

Another addition that **_may be considered_**, is the use of `...` at the end of documents. This is YAMl compliant, however much less common in usage. Also would need to confirm if CloudFormation supports them or not.

### Expected behaviour
When the template body is generated:
- If the `---` marker is not present at the start of the template, insert it.
- If the `...` marker is not present at the end of the template, insert it. (potentially)

alexharv074:
This change adds the YAML document marker to the beginning
of each template body if it doesn't already exist.

The reasoning behind this change can be found here
https://github.com/Sceptre/sceptre/issues/1087.

## PR Checklist

- [ ] Added/Updated unit tests. TODO
- [ ] Added/Updated integration tests (if applicable). N/A

## Other Information

### DIFF
--- a/sceptre/cli/template.py
+++ b/sceptre/cli/template.py
@@ -63,7 +63,15 @@ def generate_command(ctx, path):
 
     plan = SceptrePlan(context)
     responses = plan.generate()
-    output = [template for template in responses.values()]
+
+    output = []
+    templates = responses.values()
+    for template in templates:
+        if template.startswith("---"):
+            output.append(template)
+        else:
+            output.append("---\n" + template)
+
     write(output, context.output_format)
 
 


alexharv074:
This is an alternative implementation of https://github.com/Sceptre/sceptre/pull/1089 for discussion. @mrowlingfox and I discussed this internally after I had independently implemented the feature in a slightly different way. We decided it is worth raising both PRs for discussion.

I argued that `template.body` could be expected to always hold a true representation of what is in the actual template, and if so, perhaps the best place to edit the output to add missing YAML markers is in the CLI. I also wonder if there is less risk of side effects using this approach.

I have not written tests for this implementation yet, but will do so if the team decides to use this PR.

PR #1092, opened 2021-09-04
[Resolves #582] update imp to importlib

### LINKED ISSUE #582: [For V3] `imp` module is deprecated in favour of `importlib`
`template.py` is using deprecated `imp` module

the imp module is deprecated in favour of importlib
https://github.com/cloudreach/sceptre/blob/master/sceptre/template.py#L10

possible fix: is to take similar approach like here: https://github.com/pyinvoke/invoke/issues/214

zaro0508:
The imp.py library has been deprecated since ver 3.4 in favor of
importlib[1]. This updates to using importlib.

[1] https://docs.python.org/3/library/imp.html

### DIFF
--- a/integration-tests/steps/templates.py
+++ b/integration-tests/steps/templates.py
@@ -4,6 +4,7 @@ import imp
 import yaml
 
 from botocore.exceptions import ClientError
+from importlib.machinery import SourceFileLoader
 from sceptre.plan.plan import SceptrePlan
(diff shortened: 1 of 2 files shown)


ngfgrant:
For some context https://github.com/Sceptre/sceptre/issues/582

lucasvieirasilva (approved):

tarkatronic (approved):

PR #1093, opened 2021-09-04
[Resolve #683] Creating a Diff command (Prototype)

jfalkenstein:
I'm posting this at this point for feedback. It's pretty important for my company's deployment process (and I suspect many others as well) to generate the "diff" on what this set of changes will do to the currently deployed group of stacks.

This creates what has been requested in a few different issues, most notably in #683 

The syntax here is basically `sceptre diff {command path}`

## What does this do, exactly?
This compares the CURRENTLY DEPLOYED stacks in the stack group with the current configuration of stacks and templates. 
It both compares configurations and generated templates, indicating the difference.

The heavy-lifting on this is done with the excellent DeepDiff library as well as the cfn-flip library that lets you flip between CF json and yaml.

The output can be directed to a file or dumped to stdout.

## What sort of output does this create?
********************************************************************************
```
--> Difference detected for stack sceptre-dev-jf-demo-bucket-website-http-bucket!
No config difference
--------------------------------------------------------------------------------
Template difference:
values_changed:
  root['Resources']['S3Bucket07682993']['Properties']['BucketName']:
    new_value: changed-environment-sceptre-demo
    old_value: dev-jf-sceptre-demo

********************************************************************************
********************************************************************************
--> Difference detected for stack sceptre-dev-jf-demo-demo-lambda-artifact-bucket!
This stack is not deployed yet!
--------------------------------------------------------------------------------
New config:
name: sceptre-dev-jf-demo-demo-lambda-artifact-bucket
notification_arns: []
parameters:
  ArtifactBucketName: dev-jf-random-artifacts
tags:
  Deployed_Using: Sceptre
  Eng_Business_Product: demo-lambda
  Environment: dev-jf
  Reason_For_Existance: To demonstrate Sceptre

--------------------------------------------------------------------------------
New template:
Parameters:
  ArtifactBucketName:
    Type: String
    Description: The name of the Amazon S3 bucket where uploaded SAM artifacts will
      be stored.
Resources:
  ArtifactBucket7410C9EF:
    Type: AWS::S3::Bucket
    Properties:
      BucketName: !Ref 'ArtifactBucketName'
    UpdateReplacePolicy: Delete
    DeletionPolicy: Delete
    Metadata:
      aws:cdk:path: CdkStack/ArtifactBucket/Resource
  CDKMetadata:
    Type: AWS::CDK::Metadata
    Properties:
      Analytics: >-
[...]

PR #225, opened 2017-10-18
Revert "Add stack diff command (#212)"

theseanything:
This reverts commit 53f364aae48a312a337980be0fc7eb5c592ee7c2.

`sceptre diff` command is being removed. Diff was done comparing the string representations of the local template and the remote template. However, when you upload the template to AWS - its stored a JSON object and hence can be re-ordered, have whitespace changes etc - and you can’t retrieve the actual string you uploaded in the first place.

-----------------

Before submitting the PR make sure the following are checked:

* [x] Wrote [good commit messages][1].
* [x] [Squashed related commits together][2] after the changes have been approved.
* [x] Commit message starts with `[Resolve #issue-number]` (if a related issue exists).
* [x] Added unit tests.
* [x] Added integration tests (if applicable).
* [x] All unit tests (`make test`) are passing.
* [x] Used the same coding conventions as the rest of the project.
* [x] The new code doesn't generate flake8 (`make lint`) offenses.
* [x] The PR relates to *only* one subject with a clear title
  and description in grammatically correct, complete sentences.

[1]: http://chris.beams.io/posts/git-commit/
[2]: https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

PR #212, opened 2017-10-09
Add Stack Diff Command

sp-joseluis-ledesma:
Although ChangeSets provides insightful information, sometimes it does not give enough information (for example when changing Instance Tags or PolicyDocuments). This diff command will show just a raw diff between the local generated template and the Cloudformation remote one.

-----------------

* [x] Wrote [good commit messages][1].
* [x] [Squashed related commits together][2].
* [ ] Commit message starts with `[Resolve #issue-number]` (if a related issue exists).
* [x] Added unit tests.
* [x] Added integration tests (if applicable).
* [x] All unit tests (`make test`) are passing.
* [x] Used the same coding conventions as the rest of the project.
* [x] The new code doesn't generate flake8 (`make lint`) offenses.
* [x] The PR relates to *only* one subject with a clear title
  and description in grammatically correct, complete sentences.

PR #1094, opened 2021-09-07
[Resolve #683] Create a diff command

### LINKED ISSUE #683: Enhancement: Show diff (and change set details) before updating a stack

alexharv074:
Post as alternative implementation to open PR
https://github.com/Sceptre/sceptre/pull/1093

## PR Checklist

- [ ] Wrote a good commit message & description [see guide below]. TBD

## Other Information

### DIFF
--- a/requirements/prod.txt
+++ b/requirements/prod.txt
@@ -1,6 +1,7 @@
 boto3>=1.3.0,<2
 click>=7.0,<9.0
 colorama>=0.2.5,<0.4.4
+dictdiffer==0.8.1
 Jinja2>=2.8,<3
 networkx>=2.4,<2.6
 packaging>=16.8,<17.0
--- a/sceptre/cli/template.py
+++ b/sceptre/cli/template.py
@@ -100,3 +100,65 @@ def estimate_cost_command(ctx, path):
             response = response["Url"]
             webbrowser.open(response, new=2)
         write(response + "\n", 'text')
+
+
+@click.command(name="fetch-remote-template", short_help="Prints the remote template.")
+@click.argument("path")
+@click.pass_context
+@catch_exceptions
+def fetch_remote_template_command(ctx, path):
+    """
+    Prints the remote template used for stack in PATH.
+    \f
+
+    :param path: Path to execute the command on.
+    :type path: str
+    """
+    context = SceptreContext(
+        command_path=path,
+        project_path=ctx.obj.get("project_path"),
+        user_variables=ctx.obj.get("user_variables"),
+        options=ctx.obj.get("options"),
+        output_format=ctx.obj.get("output_format"),
+        ignore_dependencies=ctx.obj.get("ignore_dependencies")
+    )
+
+    plan = SceptrePlan(context)
+    responses = plan.fetch_remote_template()
+    output = [template for template in responses.values()]
+    write(output, context.output_format)
+
+
+@click.command(name="diff", short_help="Show diffs with running stack.")
+@click.argument("path")
+@click.option(
+    "-D", "--differ", type=click.Choice(["difflib", "dictdiffer"]),
+    default="difflib", help="Specify diff library, default difflib."
+)
+@click.pass_context
+@catch_exceptions
+def diff_command(ctx, path, differ):
+    """
+    Show diffs between the running and generated stack.
+    \f
+
+    :param path: The path to execute the command on.
+    :type path: str
+    :param differ: The diff library to use, default difflib.
+    :type differ: str
+    """
+    context = SceptreContext(
+        command_path=path,
+        project_path=ctx.obj.get("project_path"),
+        user_variables=ctx.obj.get("user_variables"),
+        options=ctx.obj.get("options"),
+        ignore_dependencies=ctx.obj.get("ignore_dependencies")
+    )
+
+    plan = SceptrePlan(context)
+    responses = plan.diff(differ)
+    output = "\n".join([
+        stack_name + ": " + template
+        for stack_name, template in responses.values()
+    ])
+    write(output, context.output_format)
--- a/sceptre/plan/actions.py
+++ b/sceptre/plan/actions.py
@@ -15,6 +15,10 @@ from datetime import datetime, timedelta
 
 import botocore
 import json
+import yaml
+import difflib
+import dictdiffer
+
 from dateutil.tz import tzutc
 
 from sceptre.connection_manager import ConnectionManager
@@ -873,3 +877,97 @@ class StackActions(object):
             return StackChangeSetStatus.DEFUNCT
         else:  # pragma: no cover
             raise Exception("This else should not be reachable.")
+
+    @add_stack_hooks
+    def fetch_remote_template(self):
+        """
+        Returns the Template for the remote Stack
+
+        :returns: the template body.
+        :rtype: str
+        """
+        self.logger.debug("%s - Fetching remote template", self.stack.name)
+
+        response = self.connection_manager.call(
+            service="cloudformation",
+            command="get_template",
+            kwargs={
+                "StackName": self.stack.external_name
+            }
+        )
+
+        return response.get("TemplateBody")
+
+    @add_stack_hooks
+    def diff(self, differ="difflib"):
+        """
+        Returns a diff of Template and Remote Template
+        using a specific diff library.
+
+        :param differ: The diff lib to use, default difflib.
+        :type: str
+
+        :returns: The stack name and diffs.
+        :rtype: Tuple[str, str]
+        """
+        remote_template_stream = self.fetch_remote_template()
+        local_template_stream = self.stack.template.body
+
+        if differ == "difflib":
+            function = self._diff_by_difflib
+        elif differ == "dictdiffer":
+            function = self._diff_by_dictdiffer
+
+        response = function(
+            remote_template_stream,
+            local_template_stream
+        )
+
+        return (self.stack.external_name, response)
+
+    def _diff_by_difflib(self, rt_stream, lt_stream):
+        """
+        Diffs remote and local templates using difflib.unified_diff.
+
+        :param rt_stream: remote template as a stream.
+        :type rt_stream: str
+        :param lt_stream: local template as a stream.
+        :type lt_stream: str
+
+        :returns: The diff.
+        :rtype: str
+        """
+
+        remote_template = rt_stream.split("\n")
+        local_template = lt_stream.split("\n")
+
+        diffs = difflib.unified_diff(
+            remote_template, local_template,
+            fromfile="remote_template", tofile="local_template",
+            lineterm=""
+        )
+
+        return "\n".join(diffs)
+
+    def _diff_by_dictdiffer(self, rt_stream, lt_stream):
+        """
+        Diffs remote and local templates using dictdiffer.
+
+        :param rt_stream: remote template as a stream.
+        :type rt_stream: str
+        :param lt_stream: local template as a stream.
+        :type lt_stream: str
+
+        :returns: The diff.
+        :rtype: str
+        """
+        remote_template = yaml.load(
+            rt_stream, Loader=yaml.BaseLoader
+        )
+        local_template = yaml.load(
+            lt_stream, Loader=yaml.BaseLoader
+        )
+
+        diffs = dictdiffer.diff(remote_template, local_template)
+
+        return str(list(diffs))
--- a/sceptre/cli/__init__.py
+++ b/sceptre/cli/__init__.py
@@ -25,7 +25,9 @@ from sceptre.cli.list import list_group
 from sceptre.cli.policy import set_policy_command
 from sceptre.cli.status import status_command
 from sceptre.cli.template import (validate_command, generate_command,
-                                  estimate_cost_command)
+                                  estimate_cost_command,
+                                  fetch_remote_template_command,
+                                  diff_command)
 from sceptre.cli.helpers import catch_exceptions, setup_vars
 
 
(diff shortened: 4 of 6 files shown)


sceptre/cli/template.py
```diff
@@ -100,3 +100,65 @@ def estimate_cost_command(ctx, path):
...
+    )
+
+    plan = SceptrePlan(context)
+    responses = plan.fetch_remote_template()
+    output = [template for template in responses.values()]
+    write(output, context.output_format)
+
+
+@click.command(name="diff", short_help="Show diffs with running stack.")
+@click.argument("path")
```
jfalkenstein, line 133: The convention, I think is to call this "command-path"

jfalkenstein:
Lol. @alexharv074 Did I beat you to the punch by a few days? lol.

I'd be interested to see what kind of output this produces. I'm not familiar with dict-differ. 

I also notice that you're not diffing on the _configurations_, only the templates. I originally started on my work doing a "diff command" only comparing the templates. But what I found is that (for purposes of reviewing a deployment), changes in configuration can be extremely important. That's why #1093 also does config diffing. 

Anyway, if you wanted to collaborate on this, I'd be game. I'm trying to get Sceptre adopted at my company and this functionality is going to be pretty important for us.

sceptre/cli/template.py
```diff
@@ -100,3 +100,65 @@ def estimate_cost_command(ctx, path):
...
+        user_variables=ctx.obj.get("user_variables"),
+        options=ctx.obj.get("options"),
+        output_format=ctx.obj.get("output_format"),
+        ignore_dependencies=ctx.obj.get("ignore_dependencies")
+    )
+
+    plan = SceptrePlan(context)
+    responses = plan.fetch_remote_template()
+    output = [template for template in responses.values()]
+    write(output, context.output_format)
```
jfalkenstein, line 129: I think there's a real need to be able to output the diff to a file. If you rely only on stdout, that will include all sorts of logs and printed statements from sceptre and hooks and such... it doesn't make for a "clean" output that can be posted somewhere for review by the likes of Jenkins or whatever.

sceptre/cli/template.py
```diff
@@ -100,3 +100,65 @@ def estimate_cost_command(ctx, path):
...
+        ignore_dependencies=ctx.obj.get("ignore_dependencies")
+    )
+
+    plan = SceptrePlan(context)
+    responses = plan.diff(differ)
+    output = "\n".join([
+        stack_name + ": " + template
+        for stack_name, template in responses.values()
+    ])
+    write(output, context.output_format)
```
jfalkenstein, line 164: I think there's a real need to be able to output the diff to a file. If you rely only on stdout, that will include all sorts of logs and printed statements from sceptre and hooks and such... it doesn't make for a "clean" output that can be posted somewhere for review by the likes of Jenkins or whatever

sceptre/plan/actions.py
```diff
@@ -873,3 +877,97 @@ def _get_cs_status(self, change_set_name):
...
+
+        response = self.connection_manager.call(
+            service="cloudformation",
+            command="get_template",
+            kwargs={
+                "StackName": self.stack.external_name
+            }
+        )
+
+        return response.get("TemplateBody")
```
jfalkenstein, line 899: So weird thing I found: Sometimes this returns a `dict` and sometimes this returns a string. https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation.html#CloudFormation.Client.get_template
image


Most of the time, it returns a dict, actually. But then sometimes, it returns a string, such as when the template has a transform on it, like a SAM template.

jfalkenstein: (that last bit isn't documented behavior, but it _did_ blow up my own diff command until I accounted for it)

requirements/prod.txt
```diff
@@ -1,6 +1,7 @@
 boto3>=1.3.0,<2
 click>=7.0,<9.0
 colorama>=0.2.5,<0.4.4
+dictdiffer==0.8.1
```
jfalkenstein, line 4: have you looked at DeepDiff? In my experience it's a lot more robust of a diffing library. It can be very exact about what has changed and how it changed. It even has useful serialization formats that are fairly easy to read.

sceptre/cli/template.py
jfalkenstein, line 129: I'm not sure there's a compelling usecase for this as a full-blown command.

mrowlingfox: We've been using it at Foxtel for a while now to make it easier for our developers to see the name of the actual cfn stack.

jfalkenstein: Well then I take that back...

requirements/prod.txt
mrowlingfox: So in this implementation the default diffing library we use is difflib.unified_diff. We found that gives us the most easily digestible format for humans. There are some issues with it though, specifically around how it handles ordering with JSON templates. Also I'm pretty sure it's only python 3 (but is that really a problem? 😛 ).

Note that a feature based on difflib.unified_diff was added upstream in #212 but then reverted again later in #225. 

For those 2 reasons we also wanted to provide an option for a much more robust option if the user required it.

We did look into using DeepDiff, however, found that dictdiffer provided extra information in a format that could more easily be parsed. YMMV.  If DeepDiff can be configured to give us a friendly diff format similar to difflib.unified_diff, that would be the best of both worlds!

Examples Below:

difflib.unified_diff.
```diff
test-test: --- remote_template
+++ local_template
@@ -23,10 +23,14 @@
         - Key: Project
           Value: GitOps
+        - Key: Foo
+          Value: Bar
 
 Outputs:
   InstanceArn:
     Value: !Ref MyInstance
+  AvailabilityZone:
+    Value: !GetAtt MyInstance.AvailabilityZone
 
 # vim: set ft=yaml:
```

dictdiffer
```
[aws:personal1] /src $ sceptre --dir=test --var-file=/env/common-env.yaml --var-file=/env/test.yaml diff test.yaml 
["test-test: [('add', 'Resources.MyInstance.Properties.Tags', [(6, {'Key': 'Foo', 'Value': 'Bar'})]), ('remove', 'Resources.MyInstance.Properties', [('InstanceType', 'm1.small')])]"]
```

DeepDiff
```
[aws:personal1] /src $ sceptre --dir=test --var-file=/env/common-env.yaml --var-file=/env/test.yaml diff test.yaml 
test-test: Item root['Resources']['MyInstance']['Properties']['InstanceType'] removed from dictionary.
Item root['Resources']['MyInstance']['Properties']['Tags'][6] added to iterable.
```

sceptre/plan/actions.py
mrowlingfox: That's interesting, will have to test that out for myself 😄

sceptre/cli/template.py
mrowlingfox: With this implementation, so long as `--debug` isn't specified, only the stack name and then the diff is written to stdout.

mrowlingfox:
> Lol. @alexharv074 Did I beat you to the punch by a few days? lol.
> 
> I'd be interested to see what kind of output this produces. I'm not familiar with dict-differ.
> 
> I also notice that you're not diffing on the _configurations_, only the templates. I originally started on my work doing a "diff command" only comparing the templates. But what I found is that (for purposes of reviewing a deployment), changes in configuration can be extremely important. That's why #1093 also does config diffing.
> 
> Anyway, if you wanted to collaborate on this, I'd be game. I'm trying to get Sceptre adopted at my company and this functionality is going to be pretty important for us.

That's an interesting point you raise that we aren't diffing on the configuration, you're completely right. For our use case we are focused on the template. I think this would be a modification worthwhile considering.

jfalkenstein:
@mrowlingfox Take a look at how I implemented that here: https://github.com/Sceptre/sceptre/pull/1093/files#diff-8c47d49187ea99574c8352db11966cd678e3cd648161c315b2a058900ab24116R106

There are a few tricky parts about comparing stack configurations I found:
* You can't compare _everything_ (hooks, etc...), but I think the core bits can be compared (role_arn, parameters, tags, notification arns)
* Sometimes there are resolvers for parameters that can't be resolved because they rely on stacks that don't currently exist

sceptre/cli/template.py
jfalkenstein: So we have hooks that do things like SAM Build / Package and printing outputs. What could be done is it defaults to stdout, but you can pass an `-o/--output-file` argument that instead writes to that file. This way we can output that file easily enough.

requirements/prod.txt
jfalkenstein: I found it a lot easier to read in the template file into a dictionary and do the diff on the dictionary itself rather than doing the diff on the templates. AWS's cfn-flip library lets you load a template, regardless of whether it is in json or yaml and whether it has `!Ref` or things like it in it. It loads it as a dictionary that can be compared key-for-key. This was really useful and tackles the resource and parameter ordering issue.

Take a look at #1093 and the output I posted there. I'm essentially doing this process:
1. Load the template to a dict with cfn_flip.load
2. diff the two with DeepDiff(deployed, generated, verbosity=2) (the verbosity of 2 shows exactly what the different values are so it's not just "item added" and "item removed"
3. I serialize the value to a readable json using `diff.to_json()` (which has some type serializatin functionality). If the sceptre output_format isn't "json", I convert the json to yaml.

This produces something like this:
```yaml
type_changes:
  root['Resources']['BucketPolicy']['Properties']['PolicyDocument']['Statement'][0]['Action']:
    new_type: list
    new_value:
      - s3:GetObject
      - s3:PutObject
    old_type: str
    old_value: s3:GetObject
```

sceptre/cli/template.py
alexharv074: It's actually path:

```
▶ grep -r "click.argument.*path" sceptre/cli
sceptre/cli/update.py:@click.argument("path")
sceptre/cli/delete.py:@click.argument("path")
sceptre/cli/describe.py:@click.argument("path")
sceptre/cli/describe.py:@click.argument("path")
sceptre/cli/list.py:@click.argument("path")
sceptre/cli/list.py:@click.argument("path")
sceptre/cli/list.py:@click.argument("path")
sceptre/cli/create.py:@click.argument("path")
sceptre/cli/execute.py:@click.argument("path")
sceptre/cli/policy.py:@click.argument("path")
sceptre/cli/launch.py:@click.argument("path")
sceptre/cli/template.py:@click.argument("path")
sceptre/cli/template.py:@click.argument("path")
sceptre/cli/template.py:@click.argument("path")
sceptre/cli/status.py:@click.argument("path")
```

jfalkenstein: And so it is! I'm totally wrong on this. Sorry for the misdirection.

sceptre/plan/actions.py
jfalkenstein: I think we should use TemplateStage="Original". I say this because the default is to use the processed template, which includes the post-transformed template when there IS a transform. This means that if you have a template with a transform, like a SAM template, your diff will be off, even if you haven't changed the template at all, since the templates we',re generating are all the "original". 

I even played around with aws-sam-translator to translate any Sam templates pre-diff... And there STILL was a difference, since the translator will give auto-generated resources (like API gateways) unique names every time.

sceptre/plan/actions.py
```diff
@@ -873,3 +877,97 @@ def _get_cs_status(self, change_set_name):
...
+        :type lt_stream: str
+
+        :returns: The diff.
+        :rtype: str
+        """
+        remote_template = yaml.load(
+            rt_stream, Loader=yaml.BaseLoader
+        )
+        local_template = yaml.load(
+            lt_stream, Loader=yaml.BaseLoader
```
jfalkenstein, line 968: I think we need to use scepter's Cfn yaml loader, once the intrinsic functions can blow up the BaseLoader.

alexharv074:
Closing in favour of @jfalkenstein 's branch.

PR #1099, opened 2021-09-19
[Resolve #1096] Gracefully executing SAM Change sets

### LINKED ISSUE #1096: Executing a change set on a SAM template returns a different "No changes" message, breaking Sceptre
### Subject of the issue
When Sceptre is used to create/execute a change set for a stack when there are no changes, the "FAILURE" message is different when there is a Transform on the template, such as the case on a SAM template.

As you can see below, the change set StatusReason when there are no changes on a sam template is "No updates are to be performed" with a "FAILED" status.

image

Sceptre handles this _type_ of response gracefully in a normal CF template and there's a condition for the reason returned on those responses.

### Your environment
2.6.3
3.9.4
* which OS/distro
Mac OS 10.15.7

### Steps to reproduce
Create a stack where the template is a SAM template, with a Transform on it. Apart from using hooks, this could be done manually by:
1. Using the SAM cli to create the hello-world sample project (which can be deployed with no changes) (use `sam init` for this)
2. Package up the project using `sam package --output-template-file /path/into/sceptre/templates/directory`
3. Make a Sceptre Stack config where the template_path references that packaged template
4. Launch it: `sceptre launch` your stack
5. Create a change set with no changes: `sceptre create {your-stack} change-set
6. Execute the change set: `sceptre execute {your-stack} change-set`

Doing this will blow up with an error

### Expected behaviour
This should gracefully skip this stack since there were no changes. 

### Actual behaviour
It raises an error: `"An error occurred (InvalidChangeSetStatus) when calling the ExecuteChangeSet operation: ChangeSet [arn:aws:cloudformation:us-west-2:557625315814:stack/bb-hb-cnc-beatbox-sam-application/f33eb110-1356-11ec-84d0-06c3b2675421] cannot be executed in its current status of [FAILED]"`

### My thoughts on implementation:
The _lowest_ hanging fruit on this would be simply to add "No updates" to the screen-shotted area of code in StackActions. Thus, it would essentially be something like this:

```python
ok_reasons = ("submitted information didn't contain changes", "No updates")
if status == "FAILED" and any(ok_reason in reason for ok_reason in ok_reasons):
    return 0
``` 

With that said, the idea of "chasing" error strings when these aren't even documented (so long as I can find) seems like a losing proposition. I think a _better_ option would be to tie in with one of the diff solutions being proposed in #1094 and #1093 so that we could avoid either creating or attempting to execute a change set when there is no diff observed between the current set of templates/configs and the one deployed for a given stack.

jfalkenstein:
This resolves issue #1096. SAM Templates result in a different failure message when a change set is created with no actual changes. Currently Spectre handles the same situation for _normal_ CloudFormation change sets, but it blows up if there's a SAM template since the StatusReason is a little different.

## PR Checklist

## Other Information

### DIFF
--- a/integration-tests/features/execute-change-set.feature
+++ b/integration-tests/features/execute-change-set.feature
@@ -21,8 +21,14 @@ Feature: Execute change set
     Then stack "1/A" does not have change set "A"
     And stack "1/A" was updated with change set "A"
 
-  Scenario: execute a change set for SAM template with no changes
-    Given stack "1/A" exists using "sam_template.yaml"
-    And stack "1/A" has change set "A" using "sam_template.yaml"
-    When the user executes change set "A" for stack "1/A"
-    Then stack "1/A" does not have change set "A"
+  Scenario: execute a change set that failed creation for no changes
+    Given stack "2/A" exists using "valid_template.json"
+    And stack "2/A" has change set "A" using "valid_template.json"
+    When the user executes change set "A" for stack "2/A"
+    Then stack "2/A" has change set "A" in "FAILED" state
+
+  Scenario: execute a change set that failed creation for a SAM template with no changes
+    Given stack "3/A" exists using "sam_template.yaml"
+    And stack "3/A" has change set "A" using "sam_template.yaml"
+    When the user executes change set "A" for stack "3/A"
+    Then stack "3/A" has change set "A" in "FAILED" state
--- a/integration-tests/features/execute-change-set.feature
+++ b/integration-tests/features/execute-change-set.feature
@@ -2,21 +2,27 @@ Feature: Execute change set
 
   Scenario: execute a change set that exists
     Given stack "1/A" exists in "CREATE_COMPLETE" state
-    and stack "1/A" has change set "A" using "updated_template.json"
+    And stack "1/A" has change set "A" using "updated_template.json"
     When the user executes change set "A" for stack "1/A"
     Then stack "1/A" does not have change set "A"
-    and stack "1/A" was updated with change set "A"
+    And stack "1/A" was updated with change set "A"
 
   Scenario: execute a change set that does not exist
     Given stack "1/A" exists in "CREATE_COMPLETE" state
-    and stack "1/A" does not have change set "A"
+    And stack "1/A" does not have change set "A"
     When the user executes change set "A" for stack "1/A"
     Then a "ClientError" is raised
-    and the user is told "change set does not exist"
+    And the user is told "change set does not exist"
 
   Scenario: execute a change set that exists with ignore dependencies
     Given stack "1/A" exists in "CREATE_COMPLETE" state
-    and stack "1/A" has change set "A" using "updated_template.json"
+    And stack "1/A" has change set "A" using "updated_template.json"
     When the user executes change set "A" for stack "1/A" with ignore dependencies
     Then stack "1/A" does not have change set "A"
-    and stack "1/A" was updated with change set "A"
+    And stack "1/A" was updated with change set "A"
+
+  Scenario: execute a change set for SAM template with no changes
+    Given stack "1/A" exists using "sam_template.yaml"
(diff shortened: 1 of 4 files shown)


integration-tests/features/execute-change-set.feature
```diff
@@ -2,21 +2,33 @@ Feature: Execute change set
...
     When the user executes change set "A" for stack "1/A" with ignore dependencies
     Then stack "1/A" does not have change set "A"
-    and stack "1/A" was updated with change set "A"
+    And stack "1/A" was updated with change set "A"
+
+  Scenario: execute a change set that failed creation for no changes
+    Given stack "2/A" exists using "valid_template.json"
+    And stack "2/A" has change set "A" using "valid_template.json"
+    When the user executes change set "A" for stack "2/A"
+    Then stack "2/A" has change set "A" in "FAILED" state
```
jfalkenstein, line 28: It turns out we didn't have any existing integration test for the other scenario that wasn't SAM. It felt important to have coverage for both types of message.

integration-tests/steps/stacks.py
```diff
@@ -401,7 +401,13 @@ def get_stack_status(context, stack_name, region_name=None):
...
     retry_boto_call(
         context.client.create_stack,
-        StackName=stack_name, TemplateBody=body, **kwargs
+        StackName=stack_name,
+        TemplateBody=body, **kwargs,
+        Capabilities=[
+            'CAPABILITY_IAM',
+            'CAPABILITY_NAMED_IAM',
+            'CAPABILITY_AUTO_EXPAND'
+        ]
```
jfalkenstein, line 410: You can't build a SAM stack without capabilities.

tests/test_actions.py
```diff
@@ -676,6 +676,29 @@ def test_execute_change_set_sends_correct_request(
...
+    def test_execute_change_set__change_set_is_failed_for_no_changes__returns_0(self):
+        def fake_describe(service, command, kwargs):
+            assert (service, command) == ('cloudformation', 'describe_change_set')
+            return {
+                'Status': 'FAILED',
+                'StatusReason': "The submitted information didn't contain changes",
+            }
+        self.actions.connection_manager.call.side_effect = fake_describe
+        result = self.actions.execute_change_set(sentinel.change_set_name)
+        assert result == 0
```
jfalkenstein, line 688: We didn't have a unit test for this scenario, so I made sure we had one.

jfalkenstein:
@zaro0508, there hasn't yet been comment on #1096, but getting this issue resolved is going to be pretty important for my company and our plans to use Sceptre with SAM. I see that you're one of the more active maintainers on Spectre. What is it I need to do to move this forward?

zaro0508:
@jfalkenstein I don't have any issues with this change but I also don't use CFN changesets.  To move this forward you'll want to resolve the conflicts and then have it reviewed by someone that does care about changesets.  Maybe @alexharv074  or @mrowlingfox might be interested in reviewing this?

sceptre/plan/actions.py
```diff
@@ -555,6 +555,23 @@ def execute_change_set(self, change_set_name):
         status = self._wait_for_completion()
         return status
 
+    def change_set_creation_failed_due_to_no_changes(self, reason: str) -> bool:
```
alexharv074, line 558: I think that consistency with the rest of the code base here would mean we don't provide type hints—as much as I also like type hints! Normally, rather, the return type is indicated in the doc string. (See elsewhere in the file for examples.) (I daresay `no_change_substrings` below could be a list to be more consistent too, although I didn't look at it closely and have no strong opinion.)

alexharv074 (changes requested):
Minor inconsistency with the rest of the code base I think should be fixed up, otherwise LGTM.

alexharv074 (approved):
Lgtm

jfalkenstein:
@zaro0508 now that this has been approved by @alexharv074, what more must be done to this and the issue to move this along?

zxiiro (approved):

PR #1104, opened 2021-09-20
[Resolves #1090] Install troposphere as an extra package

### LINKED ISSUE #1090: troposphere not installed with sceptre
Scepter supports the deployment of [troposphere templates](https://github.com/cloudtools/troposphere) however the troposphere library does not get installed on a sceptre install because the library dependency is not setup.  This means that troposphere must be installed separately from sceptre in order to use it.  The problem with that is that users may install a version of troposphere that may not work with sceptre.  To fix this issue I think we should add a versioned troposphere dependency to sceptre so that users will get a version that is tested and will most likey work with the installed version of scepter.

zaro0508:
Scepter supports the deployment of troposphere templates however the
troposphere library does not get installed on a sceptre install because
the library dependency is not setup. This means that troposphere must
be installed separately from sceptre in order to use it. The problem
with that is that users may install a version of troposphere that may
not work with sceptre. To fix this issue I think we should add a versioned
troposphere dependency to sceptre so that users will get a version that
is tested and will most likey work with the installed version of scepter.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,6 +34,11 @@ install_requirements = [
     "networkx>=2.4,<2.6"
 ]
 
+extra_requirements={
+    "troposphere": ["troposphere>=2.0.0,<2.1.0"],
+}
+
+
 setup(
     name="sceptre",
     version=get_version("sceptre/__init__.py"),
@@ -88,4 +93,5 @@ setup(
     ],
     test_suite="tests",
     install_requires=install_requirements,
+    extra_require=extra_requirements,
 )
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -88,4 +93,5 @@ def get_version(rel_path):
     ],
     test_suite="tests",
     install_requires=install_requirements,
+    extra_require=extra_requirements,
```
tarkatronic, line 96:
```suggestion
    extras_require=extra_requirements,
```

zaro0508: fixed

tarkatronic (approved):

zxiiro (approved):

PR #1035, opened 2021-05-07
[Resolves #894] Convert the project to use poetry

tarkatronic:
The intent here is to change the tooling over to use [poetry](https://python-poetry.org) instead of `setuptools`. This is unfortunately not a small change, especially for a project of this size. However as you can see, it will lead to a great consolidation of tools and files.

Some important things to note:

* Poetry is smart enough to automatically include data files such as the `*.json` that had to be manually included by setuptools.
* `twine` is no longer necessary as we can simply use `poetry publish`.
* Poetry handles all virtualenv matters automatically. So we no longer need to instruct users to construct virtualenvs for development purposes, nor do we need to manually create and manage them for CI.
* Dependency specifications are a bit different with Poetry, and a bit more "modern", [as seen here](https://python-poetry.org/docs/dependency-specification/).
* I have told git (via the `.gitattributes`) file to treat the `poetry.lock` file as binary. This is a fairly standard practice, also followed in the JS community with the `package-lock.json` file. This does two primary things:
  * Tells GitHub to not attempt to display a diff for this file
  * Prevents frequent merge conflicts, since this tends to be a frequently updated file
* I have included `awscli` and all of the Sphinx-related requirements as dev requirements. This means that they will be installed when a developer runs `poetry install`. If that's not desirable, I can remove `awscli`, and move the Sphinx requirements to an "extra", [as seen here](https://github.com/godaddy/tartufo/blob/74291f93d0f33b5279bf6d1f62d2df453658281a/pyproject.toml#L65-L66). This latter part would mean that you would have to run `poetry install -E docs` to get the Sphinx requirements.

Note that I am starting this as a draft PR while I work on it, to show my progress. There will likely need to be significant work done to uplift the CI to work with poetry. I can pretty much guarantee that it WILL NOT pass right now. But luckily, CircleCI [may have an easy way](https://circleci.com/developer/orbs/orb/circleci/python#usage-work-with-poetry) to help facilitate this!

## Checklist for completion of this change

- [x] Set up base poetry config in pyproject.toml
- [x] Update tox to use poetry
- [x] Update Makefile to use poetry
- [x] Update Dockerfile to use poetry
- [x] Update CircleCI to use poetry
- [x] Remove setuptools config

## PR Checklist

- [X] Wrote a good commit message & description [see guide below].
- [ ] Commit message starts with `[Resolve #issue-number]`.
- [x] Added/Updated unit tests.
- [ ] Added/Updated integration tests (if applicable).
[...]

PR #1105, opened 2021-09-21
Setup to build docs on readthedocs.org

zaro0508:
This is to partially resolve issue #1098.

We plan to move sceptre docs away from hosting on a
[cloudreach domain](https://sceptre.cloudreach.com/2.6.3/) to
https://readthedocs.org/project/sceptre

This PR adds the file to build the sphinx docs on readthedocs.org

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,21 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+   configuration: docs/_source/conf.py
+
+# Optionally build your docs in additional formats such as PDF
+formats:
+   - pdf
+
+# Optionally set the version of Python and requirements required to build your docs
+python:
+   version: "3.8"
+   install:
+   - requirements: requirements/prod.txt
+   - requirements: requirements/dev.txt


ngfgrant:
What are the benefits of moving away from GitHub pages?

zxiiro (approved):

zaro0508:
@ngfgrant github pages is markdown format while the sceptre docs is reStructuredText format

ngfgrant:
Oh right - the docs are currently being served from GitHub Pages (the format of the content doesn't matter as it's all generated to html in the end).

ngfgrant:
To change the url you just need to edit the CNAME file here https://github.com/Sceptre/sceptre.github.io/tree/master

Alternatively, just remove it and you can access the docs at sceptre.github.io

zaro0508:
thanks @ngfgrant.    can we just replace cloudreach domain in the CNAME with the domain you've setup @tarkatronic ?

craighurley:
> Alternatively, just remove it and you can access the docs at sceptre.github.io

I didn't realise that was an option.  The discussion about registering a domain for the project was mainly focused on a domain for the docs.  If the docs can be accissible from sceptre.github.io, isn't that good enough and removes the need for maintaining a domain?

ngfgrant:
@zaro0508 yep that is correct. Alternatively, as Craig said we can just use the GitHub provided domain.

zxiiro:
> What are the benefits of moving away from GitHub pages?

One of the nice things that ReadTheDocs provides as a built-in UI and support for multiple versions of docs. Switching to GitHub pages would mean we lose that right?

According to https://sceptre.cloudreach.com/ on the bottom left, it loooks like CloudReach's Sceptre docs are hosoted through ReadTheDocs right now or am I missing something?

ngfgrant:
@zxiiro  we wouldn't be switching to GitHub Pages as we are already using it. We also maintain versioned docs already.. 

Sceptre docs are "hosted" at https://github.com/Sceptre/sceptre.github.io

zxiiro:
> @zxiiro  we wouldn't be switching to GitHub Pages as we are already using it. We also maintain versioned docs already.. 
> 
> 
> 
> Sceptre docs are "hosted" at https://github.com/Sceptre/sceptre.github.io

Interesting. I guess I haven't seen a readthedocs theme deployment outside of readthedocs which was what threw me off. Good to know it's possible.

ngfgrant (changes requested):
I would leave this for now as there is nothing wrong or restrictive about the current set up.

tarkatronic:
> > Alternatively, just remove it and you can access the docs at sceptre.github.io
> 
> I didn't realise that was an option. The discussion about registering a domain for the project was mainly focused on a domain for the docs. If the docs can be accissible from sceptre.github.io, isn't that good enough and removes the need for maintaining a domain?

The initial discussion around registering the domain was to get the GitHub org verified, and eventually to be able to become a verified publisher, specifically for the GitHub Action.

The domain itself can be pointed at either GitHub Pages or ReadTheDocs. That part doesn't really matter. Although I would still recommend using RTD, as a way to reduce complexity. That build process right now is managed by a series of self-authored scripts which could very easily break at any time, whereas using RTD is simply a matter of setting up a config and installing the application on GitHub.

I personally advocate for any tooling that can reduce complexity in a build pipeline.

ngfgrant:
> The initial discussion around registering the domain was to get the GitHub org verified, and eventually to be able to become a verified publisher, specifically for the GitHub Action.
> 
> The domain itself can be pointed at either GitHub Pages or ReadTheDocs. That part doesn't really matter. Although I would still recommend using RTD, as a way to reduce complexity. That build process right now is managed by a series of self-authored scripts which could very easily break at any time, whereas using RTD is simply a matter of setting up a config and installing the application on GitHub.
> 
> I personally advocate for any tooling that can reduce complexity in a build pipeline.

IIRC the factors around original decision for how we handled docs were: 

- reduce "tooling sprawl"
- remove dependency on 3rd party service and any incompatible version changes (e.g. RTD v1 vs v2). 
- the flexibility/control of having specific docs available (e.g. we wanted to keep some of the old v1 docs alive). 

The scripts could break at any time, but in reality they haven't since they were published 2 years ago... famous last words eh?

Ideally, we would reduce tool sprawl further by bringing the CI/CD to GitHub too... but that is another story. 

A side note having these sorts of suggestions is fine but I would say they add limited value. Thanks to me (and some others) bigger problems exist in Sceptre - such as useful error messages or more meaningful debug statements. If we have people that are engaged with the project it would be good to get stuck in to the core concerns rather than around the periphery.

tarkatronic:
> Ideally, we would reduce tool sprawl further by bringing the CI/CD to GitHub too... but that is another story.

We have definitely talked about moving to GitHub Actions! The complexity of CircleCI has been quite a frustration, especially in #1035. A lot of the problems I've had to deal with are very specific to CircleCI, and (in my experience) don't exist, or manifest the same, on GitHub Actions.

> A side note having these sorts of suggestions is fine but I would say they add limited value. Thanks to me (and some others) bigger problems exist in Sceptre - such as useful error messages or more meaningful debug statements. If we have people that are engaged with the project it would be good to get stuck in to the core concerns rather than around the periphery.

I agree there are definitely improvements to be made in the core, but the tooling around the project is a place where people like myself, less intimately familiar with the core, can help immediately.

tarkatronic:
The other goal here is to lend further legitimacy to the project; having an official domain (which is totally free btw) and that little verified checkmark wherever possible can tend to add to that. I know that things like that give me a little more confidence in projects when I see them.

Plus, selfishly, my employer only allows us to use GitHub Actions from verified publishers. So I'm hoping that the Sceptre action can become one of those viable for use.

ngfgrant:
100% onboard for new domain and GitHub Action sounds very cool - hopefully we can utilise the existing Dockerfile?

zaro0508:
> A side note having these sorts of suggestions is fine but I would say they add limited value. Thanks to me (and some others) bigger problems exist in Sceptre - such as useful error messages or more meaningful debug statements. If we have people that are engaged with the project it would be good to get stuck in to the core concerns rather than around the periphery.

tarkatronic:
> 100% onboard for new domain and GitHub Action sounds very cool - hopefully we can utilise the existing Dockerfile?

The goal is that the `Dockerfile` in this project will be able to translate over with no problem. The only modifications are in #1035; in there, I have changed it to a multi-stage build, utilizing Poetry to install dependencies. In this way we eliminate the fragmentation of requirements specifications (`setup.py`, `requirements/dev.txt`, and `requirements/prod.txt`) and, with the multi-stage build, it will ideally generate a smaller image overall!

If you're referring to the `sceptre-circleci` Docker image, then we should actually be able to fully eliminate that and just use the default GitHub Actions runner images 🤞🏻 -- at least, that's the hope.

And I'm super excited to work on that transition, if I can ever get the Poetry work finalized and landed!

PR #1110, opened 2021-09-24
[Resolves #1106] S3 template handler jinja and python support

### LINKED ISSUE #1106: S3 template handler: support jinja and python templates
Sceptre allows referencing templates directly in S3 buckets using the [S3 template handler](https://github.com/Sceptre/sceptre/blob/master/docs/_source/docs/template_handlers.rst#s3).  However that handler does not support jinja and python based templates.  It would be great if we could add support for jinja and troposphere (python) templates.

zaro0508:
Currently the S3 template handler does not support jinja and python
based templates (troposphere).  This adds Support for those template
types.


sceptre/template_handlers/s3.py
```diff
@@ -22,11 +31,51 @@ def schema(self):
...
+        if file_extension in {".json", ".yaml", ".template"}:
+            return template
+        elif file_extension in {".j2", ".py"}:
+            try:
+                template = self._get_template(bucket, key)
+                file = tempfile.NamedTemporaryFile(prefix=f'{filename}_')
+                with file as f:
+                    f.write(template)
+                    f.seek(0)
+                    f.read()
```
jfalkenstein, line 51: What are you doing on this line?

sceptre/template_handlers/s3.py
```diff
@@ -22,11 +31,51 @@ def schema(self):
...
     def handle(self):
-        self.logger.debug("Downloading file from S3: %s" % self.arguments["path"])
-
+        logger.debug(f'test200')
         segments = self.arguments["path"].split('/')
         bucket = segments[0]
         key = "/".join(segments[1:])
+        filename = segments[-1]
+        extension = filename.split('.')[1]
+        file_extension = f'.{extension}'
```
jfalkenstein, line 40: You could probably use pathlib.Path and get a lot of this extension stuff for free.

PR #1111, opened 2021-09-24
Update troposphere library

zaro0508:
When attempting to deploy troposphere template I get the following error

```
AttributeError: 'Template' object has no attribute 'to_yaml'

```

Accoring to https://github.com/cloudtools/troposphere/issues/759  it
seems like we need to update trophoshere to a version where the
`to_yaml` was added back.  The current dependency to trophosphere (v2.0.2)
is pretty old anyways so we are updating the version to the latest 2.x
release

The latest version is 3.0.3 however updating to 3.x causes sceptre unit
tests to fail so we are sticking to the latest 2.x version.

### DIFF
--- a/requirements/prod.txt
+++ b/requirements/prod.txt
@@ -9,4 +9,4 @@ PyYaml>=5.1,<6.0
 sceptre-cmd-resolver>=1.1.3,<2
 sceptre-file-resolver>=1.0.4,<2
 six>=1.11.0,<2.0.0
-troposphere>=2.0.0,<2.1.0
+troposphere>=2.0,<3
--- a/requirements/prod.txt
+++ b/requirements/prod.txt
@@ -9,4 +9,4 @@ PyYaml>=5.1,<6.0
 sceptre-cmd-resolver>=1.1.3,<2
 sceptre-file-resolver>=1.0.4,<2
 six>=1.11.0,<2.0.0
-troposphere>=2.0,<3
+troposphere>=2.7,<3
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,7 @@ install_requirements = [
 ]
 
 extra_requirements = {
-    "troposphere": ["troposphere>=2,<3"],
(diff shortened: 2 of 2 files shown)


craighurley (approved):

tarkatronic (commented):
Given that we know 2.0.x doesn't work, should we keep the restriction higher? Perhaps `>=2.1.0,<3`? Or even `>=2.2.0,<3`?

It might be worth figuring out what versions actually work. Because with this specification, somebody could still install a broken version.

zaro0508:
> It might be worth figuring out what versions actually work. Because with this specification, somebody could still install a broken version.

Good point @tarkatronic , i have verified that the latest 2.x  version (v2.7.1) works so i have adjusted the version accordingly.

requirements/prod.txt
```diff
@@ -9,4 +9,4 @@ PyYaml>=5.1,<6.0
 sceptre-cmd-resolver>=1.1.3,<2
 sceptre-file-resolver>=1.0.4,<2
 six>=1.11.0,<2.0.0
-troposphere>=2.0.0,<2.1.0
+troposphere>=2.7,<3
```
craighurley, line 12: @zaro0508 recently in slack, you mentioned `2.0.2` works.  Should this be `>=2.0.2,<3` or are there versions greater than `2.0.2` and less than `2.7` that cause issues?

zaro0508: sorry, i think the slack conversation was a bit confusing.  To clarify, I get the `AttributeError` on a troposphere template deployment when run with ver 2.0.2.  I reran the deployment with troposphere ver 2.7.1 and it works without error.  I did not test with any other versions.

craighurley: got it

craighurley (approved):

zaro0508:
are you satisfied @tarkatronic ?

tarkatronic (approved):
Much better!

PR #1088, opened 2021-08-28
[Resolves #213] Add support for template handlers 

zaro0508:
This PR is an update to PR #864

This PR adds support for template handlers, a new entity used to load templates from external sources. The `Template` class now always expects a configuration object for the template handler, and will instantiate it based on the entry points found for `sceptre.template_handlers`. In this way, others can add new template handlers and use them in the Sceptre config.

`TemplateHandler` objects define a JSON schema that will be validated upon use. This will make sure that template handler authors and users provide the correct configuration.

An example of the S3 template handler is as follows:

```yaml
template:
    type: s3
    path: my-bucket/my-fancy-template.yaml
parameters:
    ...
sceptre_user_data:
    ...
```

All the properties of the complex `template` object will be passed into the template handler as arguments (except for `type`). 

To not break backwards compatibility, the `template_path` property can still be used. It will be wired into a `file` template handler in `stack.py`, so the 'ugliness' is contained there. `Template` can then always work with handlers, without considering `template_path`. The `file` template handler contains all the logic for loading Jinja2 templates and/or Python scripts, that was moved from `Template`. So that logic is not in the core of Sceptre any more (which is good IMHO).

The `s3` template handler that is included here does not support Jinja2 templates or Python scripts. That is a limitation that can be solved by writing some code that will check the extension and parse the S3 contents from memory (similar to what `file` does, but that only supports files on disk, not strings in memory).

PR #1112, opened 2021-09-25
Document deprecation of the template_path key

zaro0508:
We now have template handlers (PR https://github.com/Sceptre/sceptre/pull/1088)
with a new `template` key in sceptre cconfig.  This means we can
deprecate the `template_path` key.


craighurley:
Is the intention to only update the docs, leaving the `template_path` functionality in place for backwards compatibility?  If `template_path` functionality be removed entirely, when?  I ask out of concern that we don't break things for users without clear notice and time to migrate.

zaro0508:
> Is the intention to only update the docs, leaving the `template_path` functionality in place for backwards compatibility? 

Yes, that is the idea.  New sceptre users should be using the`template` instead of the `template_path` key.  Also we don't want to support multiple ways to reference local template files.  

> If `template_path` functionality will be removed entirely, when? I ask out of concern that we don't break things for users without clear notice and time to migrate.

This PR was meant to raise discussion for deprecating `template_path` and when it should be removed.  I'm open to ideas on how it should be done.  Next major release?  A pre-set  EOL date?

zxiiro:
I think this is fine. If we follow SemVer then that right time to remove the functionality is when we bump the first number of the version. So for Sceptre that means a version 3.0.0 release.

docs/_source/docs/stack_config.rst
```diff
@@ -37,6 +37,9 @@ Sceptre treats the template as CloudFormation, Jinja2 or Python depending on
 the template’s file extension. Note that the template filename may be different
 from the Stack config filename.
 
+.. note::
+  This key is deprecated in favor of the `template`_ key.
```
zxiiro, line 41: Maybe we can make it more explicit by adding something like "will be removed in 3.0.0".

zaro0508: done

zxiiro (approved):

zaro0508:
are you ok with this PR @craighurley ?

craighurley (approved):

craighurley:
> are you ok with this PR @craighurley ?

All good :)

PR #1115, opened 2021-10-01
Update docs

zaro0508:
* Contribotrs doc is outdated and not needed since users can just look
  at the Sceptre org users and teams for that info.
* Add docs about pycharm debugger not working with project settings
  and how to fix it.


craighurley (dismissed):

tarkatronic (approved):
Considering how out-of-date this doc was, I'd say this is a good change!

Another thing to consider, perhaps in the future, would be something like the All Contributors bot: https://allcontributors.org/

pyproject.toml
```diff
@@ -23,4 +23,8 @@ fail_under = 90.0
 show_missing = true
 
 [tool.pytest.ini_options]
+# The "--cov" option will the disable the pycharm debugger.  To enable the debugger either
```
zaro0508, line 26: will the/will disable

PR #1116, opened 2021-10-01
[RESOLVE #946] Fixing bug preventing StackGroup dependencies

### LINKED ISSUE #946: stack dependencies in cascading configs
assuming i have the following project structure

```
.
├── config.yaml
├── prod
│   ├── config.yaml
│   ├── vpc.yaml
│   └── vpcPeering.yaml
├── test
│   ├── config.yaml
│   ├── vpc.yaml
│   └── vpcPeering.yaml
├── vpc.yaml
└── vpcPeering.yaml
```

i would like to define a dependency in vpcPeering.yaml to vpc.

at the moment the only way i could get this work is to define the dependencies in both ./test/vpcPeering.yaml and ./prod/vpcPeering.yaml

by adding 

```
dependencies:
    - {{ stack_group_config.profile }}/vpc.yam
```

is there a smarter way to do this. ideally i want to configure the dependency just 1 time in ./vpcPeering.yaml but i was not able to do so.

jfalkenstein:
This resolves issue #946 by solving the issue of inherited dependencies. Turns out, the problem was just a small bug with inherited objects in memory.

There was a small bug in the ConfigReader that has to do with the way list-joining works on individual stacks and how dependencies are resolved.

The basic situation was this:

* Sceptre uses the "list_join" strategy to combine StackGroup dependencies with Stack dependencies. 
* The list_join strategy creates a NEW list by combining A + B... **but only if both A and B exist**. 
* In the case where the StackGroup dependencies exist _but there are no Stack dependencies_, Sceptre just selects A and passes each stack in that group the exact same dependencies list (i.e. **same object in memory**) from the StackGroup
* The process of resolving stacks replaces the stack string with a Stack object _in place_.
* If two stacks have inherited the exact same dependencies list because neither of them defined their own dependencies lists, the first stack to be resolved will replace those objects in memory with Stack objects and the second stack to be resolved will receive those Stack objects rather than strings.
* In the attempt to resolve the Stack object (believing it's a string), Sceptre will blow up because you cannot call `.replace()` on a Stack instance. But we don't need to do that at all if it's already a Stack.

This PR fixes that bug by skipping resolution for dependencies that are already Stack objects.
## PR Checklist

## Other Information

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,6 @@ __pycache__/
 env/
 venv/
 .venv/
-.python-version
 build/
 develop-eggs/
 dist/
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ __pycache__/
 env/
 venv/
 .venv/
+.python-version
 build/
 develop-eggs/
 dist/
--- a/tests/test_config_reader.py
+++ b/tests/test_config_reader.py
@@ -525,6 +525,39 @@ class TestConfigReader(object):
         else:
             assert False
 
+    @pytest.mark.parametrize(
+        "filepaths, dependency, parent_config_path", [
+            (["A/1.yaml", "A/2.yaml", "B/1.yaml"], "B/1.yaml", 'A/config.yaml'),
+            (["A/1.yaml", "A/2.yaml"], "A/1.yaml", 'A/config.yaml'),
+        ]
+    )
+    def test_inherited_dependency_already_resolved(self, filepaths, dependency, parent_config_path):
+        project_path, config_dir = self.create_project()
+        parent_config = {
+            'dependencies': [dependency]
+        }
+        abs_path = os.path.join(config_dir, parent_config_path)
+        self.write_config(abs_path, parent_config)
+
+        for rel_path in filepaths:
+            # Set up config with reference to an existing stack
+            config = {
+                "project_code": "project_code",
+                "region": "region",
+                "template_path": rel_path,
+            }
+
+            abs_path = os.path.join(config_dir, rel_path)
+            self.write_config(abs_path, config)
+        self.context.project_path = project_path
+        try:
+            config_reader = ConfigReader(self.context)
+            all_stacks, command_stacks = config_reader.construct_stacks()
+        except Exception:
+            raise
+        else:
+            assert True
+
     def test_resolve_node_tag(self):
         mock_loader = MagicMock(yaml.Loader)
         mock_loader.resolve.return_value = "new_tag"
--- a/sceptre/config/reader.py
+++ b/sceptre/config/reader.py
@@ -272,7 +272,10 @@ class ConfigReader(object):
             if not self.context.ignore_dependencies:
                 for i, dep in enumerate(stack.dependencies):
                     try:
-                        stack.dependencies[i] = stack_map[sceptreise_path(dep)]
+                        if not isinstance(dep, Stack):
+                            # If the dependency was inherited from a stack group, it might already
+                            # have been mapped and so doesn't need to be mapped again.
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -11,6 +11,7 @@ __pycache__/
 env/
 venv/
 .venv/
+.python-version
```
craighurley, line 14: Why was this included?

craighurley: I could have sworn I saw folks discussing adding adding that file to the repo, but it looks like it is not there, adding to ignore might be ok

jfalkenstein: This is used by pyenv to set up a virtual environment. It's similar to a venv directory. Nothing specific to the project, just prevents accidentally committing this to the local repo.

craighurley: I know the purpose of `.python-version`, my point is that it has been included in this repo in the past and I know folks discussed keeping it in the past.

https://github.com/Sceptre/sceptre/tree/20579c49ac62ec720cfedca097bbe55daf825a93
https://github.com/Sceptre/sceptre/search?q=%22.python-version%22&type=commits

Ideally a PR should only touch a limited set of files related to the functionality being added/removed/updated.

jfalkenstein: That's fine. I can pull it out. Not a big deal.

jfalkenstein: This has been pulled out.

jfalkenstein:
@craighurley any chance I could get a follow-up review on this?

zaro0508 (changes requested):
Issue #946 seems to indicate that this fix will provide new functionality to sceptre.  would you be willing to add some documentation telling users about cascading dependencies and how they configure sceptre for it?

tests/test_config_reader.py
```diff
@@ -525,6 +525,39 @@ def test_missing_dependency(
...
+                "region": "region",
+                "template_path": rel_path,
+            }
+
+            abs_path = os.path.join(config_dir, rel_path)
+            self.write_config(abs_path, config)
+        self.context.project_path = project_path
+        try:
+            config_reader = ConfigReader(self.context)
+            all_stacks, command_stacks = config_reader.construct_stacks()
```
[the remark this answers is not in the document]
jfalkenstein: So we already test stack resolution in other tests (see here: https://github.com/Sceptre/sceptre/pull/1116/files#diff-b2719a0eb862c151c5232ec3d1d3f09086a76ad10a01f5dfa26deb37085b770eL465-L493). This particular test is covering the bug. With the code currently in master, this test fails. With my fix, it passes. I feel like that's sufficient.

jfalkenstein:
@zaro0508 That's an interesting thought. #946 seems to call cascading dependencies as a feature, but in reality, there's nothing new about what I have done. In fact, we were already using the "list_join" strategy on dependencies. Sceptre cascades almost all configs and there was explicit code in place to handle this in the case of dependencies... but in specific circumstances, this cascade raised an error. So I would categorize this change more as a "bug fix" than a new feature.

With that said, I could absolutely submit a follow-up PR to improve the docs on "[Cascading Config](https://sceptre.cloudreach.com/2.6.3/docs/stack_group_config.html#cascading-config)", since Sceptre's cascading is one of its best features and it doesn't really ever highlight how awesome and flexible this feature-set is. I think adding docs that indicate the merge strategies for each stack configuration would also be super helpful, since you have to consult the source code to find that. Would you accept a follow-up PR that does this, @zaro0508 ?

tests/test_config_reader.py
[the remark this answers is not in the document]
zaro0508: sounds good

zaro0508 (approved):
> Would you accept a follow-up PR that does this, @zaro0508 ?

Yes, that would be great @jfalkenstein!

PR #1127, opened 2021-10-14
Set file as the default template handler type

zaro0508:
Now that we have template handlers we want to encourage users to
migrate from using `template_path` to `template` in their Sceptre
stack config files.  To make that transition easier we set the
default `template` type to 'file' which will save the user from
having to specify a full `template` block if they just want to
load a file from disk


zaro0508:
this is related to PR #1112

PR #1145, opened 2021-10-29
Add retry and timeout to http template handler

zaro0508:
The python request library does not provide retries by
default. To make the http template handler more robust
we retry the download a few times.  Also by default the
request library timeout is set to off which means it
can hang indefinitely if it get stuck downloading templates.
We add a timeout to make sure downloads don't get stuck.


sceptre/template_handlers/http.py
```diff
@@ -75,9 +77,28 @@ def _get_template(self, url):
         :rtype: str
         """
         self.logger.debug("Downloading file from: %s", url)
+        session = self._get_retry_session()
         try:
-            response = requests.get(url)
+            response = session.get(url, timeout=(2, 5))
             return response.content
-        except requests.exceptions.RequestException as e:
+        except Exception as e:
```
zxiiro, line 84: Is there a more specific exception we are expecting here? I realize that changing request to session might have changed the exception here.

zxiiro (approved):
I just have a question but I don't think it's blocking.

sceptre/template_handlers/http.py
```diff
@@ -75,9 +77,28 @@ def _get_template(self, url):
...
             self.logger.fatal(e)
```
jfalkenstein, line 85: Why `fatal` and not error here?

sceptre/template_handlers/http.py
```diff
@@ -75,9 +77,28 @@ def _get_template(self, url):
         :rtype: str
         """
         self.logger.debug("Downloading file from: %s", url)
+        session = self._get_retry_session()
         try:
-            response = requests.get(url)
+            response = session.get(url, timeout=(2, 5))
```
jfalkenstein, line 82: I think the retry settings should probably be configured as handler options. Without being an explicit option, I wonder what's the purpose of ALWAYS retrying?

sceptre/template_handlers/http.py
zaro0508: fixed

sceptre/template_handlers/http.py
zaro0508: changed to raise exception

sceptre/template_handlers/http.py
zaro0508: I've changed it so that users can pass the timeout and retry as handler options.

sceptre/template_handlers/http.py
```diff
@@ -66,18 +77,57 @@ def handle(self):
...
+        Get the template handler options
+        :param url: The option name
+        :type: str
+        :param default: The default value if option is not set.
+        :rtype: int
+        """
+        if HANDLER_OPTION_KEY in self.stack_group_config:
+            option = self.stack_group_config.get(HANDLER_OPTION_KEY)
+            if name in option:
+                return option.get(name)
```
jfalkenstein, line 131: Why aren't these argument options? I'm confused as to why you're pulling these in via the stack_group_config rather than arguments defined in the handler's schema.

zaro0508: It's because the template schema is defined per sceptre config.   I think  it makes more sense to apply these options at a higher level, across all sceptre configs.  That's why I've opted to pull from stack_group_config.

jfalkenstein: I see... then I think this should definitely be included in the docs. Otherwise, how would anyone know about these settings?

zaro0508: good call, I completely forgot.  It's there now.  Thanks!

docs/_source/docs/stack_group_config.rst
```diff
@@ -92,6 +93,18 @@ This will impact :ref:`Templating` of stacks by modifying the behavior of jinja.
       trim_blocks: True
       newline_sequence: \n
 
+http_template_handler
+~~~~~~~~~~~~~~~~~~~~~
+
+Optios passed to the `http template handler`_.
```
jfalkenstein, line 99:
```suggestion
Options passed to the `http template handler`_.
```

docs/_source/docs/stack_group_config.rst
```diff
@@ -92,6 +93,18 @@ This will impact :ref:`Templating` of stacks by modifying the behavior of jinja.
...
+  * retries - The number of retry attempts (default is 5)
+  * timeout - The timeout for the session in seconds (default is 5)
```
jfalkenstein, line 101: I'm thinking that we might want to clarify these a little more so they don't accidentally conflict with other sceptre configs since these are so common of names. 

What about:
* http_handler_retries
* http_handler_timeout
?

jfalkenstein: Never mind! I just saw that these are really sub-keys.

docs/_source/docs/stack_group_config.rst
```diff
@@ -92,6 +93,18 @@ This will impact :ref:`Templating` of stacks by modifying the behavior of jinja.
...
+
+.. code-block:: yaml
+
+   http_template_handler:
+      retries: 10
+      timeout: 20
```
jfalkenstein, line 107: 👍

docs/_source/docs/template_handlers.rst
```diff
@@ -67,8 +67,10 @@ Example:
 http
 ~~~~~~~~~~~~~
 
-Downloads a template from a url on the web.  This handler supports templates with .json, .yaml,
-.template, .j2 and .py extensions.
+Downloads a template from a url on the web.  By default, this handler will attempt to download
+templates with 5 retries and a download timeout is set to 5 seconds.  The retry and timeout
```
jfalkenstein, line 71:
```suggestion
templates with 5 retries and a download timeout of 5 seconds.  The retry and timeout
```

sceptre/template_handlers/http.py
```diff
@@ -4,10 +4,19 @@
...
+from requests.packages.urllib3.util.retry import Retry
 from sceptre.exceptions import UnsupportedTemplateFileTypeError
 from sceptre.template_handlers import TemplateHandler
 from urllib.parse import urlparse
 
+HANDLER_OPTION_KEY = "http_template_handler"
+HANDLER_RETRIES_OPTION_PARAM = "retries"
+DEFAULT_RETRIES_OPTION = 5
+HANDLER_TIMEOUT_OPTION_PARAM = "timeout"
+DEFAULT_TIMEOUT_OPTION = 5
```
jfalkenstein, line 18: 5 seems a bit low, I guess. But I also can't see why something as simple as a text document like a template should take 5 seconds to download. 🤷

sceptre/template_handlers/http.py
```diff
@@ -66,18 +77,57 @@ def handle(self):
...
         """
         Get template from the web
         :param url: The url to the template
         :type: str
-        :returns: The body of the CloudFormation template.
-        :rtype: str
+        :returns: The number of retry attempts.
+        :rtype: int
+        :returns: The timeout for the session in seconds.
+        :rtype: int
```
jfalkenstein, line 88: You know, with type annotations, you wouldn't need even half of these...

docs/_source/docs/stack_group_config.rst
zaro0508: fixed

docs/_source/docs/template_handlers.rst
zaro0508: fixed

sceptre/template_handlers/http.py
zaro0508: yeah, i've noticed that it's mostly just getting the connection that fails occasionally not actually downloading the file.

sceptre/template_handlers/http.py
zaro0508: those are supposed to be param not returns.  i have fixed.

jfalkenstein (approved):

PR #1146, opened 2021-10-31
Remove .python-version file

zaro0508:
Having .python-version file requires contributors to have python 3.6 installed
which should not be a requirement.  It also messes up pyenv's python
management.


craighurley:
Should you also be adding the file to .gitignore?

zxiiro:
yes let's add this to `.gitignore`. IMO pyenv is only significant for the user and not the project.

zxiiro (approved):

Issue #1148, opened 2021-11-03
Sceptre fails to find template if sceptre_dir is set


alexharv074:
### Subject of the issue

When a _relative path_ is passed to `--dir` and `template_path` contains a path relative to `sceptre_dir + "/templates"`, Sceptre builds an  incorrect path to the template.

### Your environment
* version of sceptre - latest master branch
* version of python (python --version) - 3.6.9
* which OS/distro - Ubuntu

### Steps to reproduce

A config like:

```yaml
# redshift-cluster/config/redshift-cluster.yaml
template_path: redshift-cluster.template.yaml.j2
```

and:

```text
tree redshift-cluster/templates/
redshift-cluster/templates/
└── redshift-cluster.template.yaml.j2
```

Run:

```text
sceptre --debug --dir=redshift-cluster validate redshift-cluster.yaml
```

### Expected behaviour

Should work fine and validate the stack.

### Actual behaviour

```text
  File "/usr/local/lib/python3.6/dist-packages/sceptre/template.py", line 90, in body                                                                                                                 
    body = handler.handle()                                                                                                                                                                           
  File "/usr/local/lib/python3.6/dist-packages/sceptre/template_handlers/file.py", line 54, in handle                                                                                                 
    raise e                                                                                                                                                                                           
  File "/usr/local/lib/python3.6/dist-packages/sceptre/template_handlers/file.py", line 43, in handle                                                                                                 
    with open(path) as template_file:                                                                                                                                                                 
FileNotFoundError: [Errno 2] No such file or directory: 'redshift-cluster/templates/redshift-cluster/templates/redshift-iam-roles.template.yaml'                                                      
```

alexharv074:
This bug appears to have been introduced in 5c0ab398b3a634336ce9dbd234b3d4b7d288ae8c. Note also that prior to introduction of template handlers, a hack had existed to solve this problem [here](https://github.com/Sceptre/sceptre/blob/354e87e8fbf2dafd2692c7c87abc1ecc35ea3f40/sceptre/template.py#L152-L167).

zaro0508:
hmm, this seems to work for me.

```➜ pwd
/Users/zaro0508

➜ tree /tmp/itsandbox
/tmp/itsandbox
├── README.md
├── config
│   ├── config.yaml
│   └── file
│       ├── a.yaml
│       ├── b.yaml
│       └── c.yaml
└── templates
    └── noop.j2
    
[...]

PR #1151, opened 2021-11-07
Refactor template handler

alexharv074:
In 0abe66966bc1bd86df3884389a61edc675a8f98c, some unnecessary
refactoring occurred while fixing Issue #1148. This reverts that
refactoring and adds unit tests for the File handler.

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/template_handlers/file.py
+++ b/sceptre/template_handlers/file.py
@@ -1,12 +1,9 @@
 # -*- coding: utf-8 -*-
 import sceptre.template_handlers.helper as helper
 
-from os import path
 from pathlib import Path
-
 from sceptre.exceptions import UnsupportedTemplateFileTypeError
 from sceptre.template_handlers import TemplateHandler
-from sceptre.helpers import normalise_path
 
 
 class File(TemplateHandler):
@@ -27,8 +24,12 @@ class File(TemplateHandler):
         }
 
     def handle(self):
+        project_path = self.stack_group_config.get("project_path")
         input_path = Path(self.arguments["path"])
-        path = self._resolve_template_path(str(input_path))
+        if input_path.is_absolute():
+            path = str(input_path)
+        else:
+            path = str(Path(project_path) / 'templates' / input_path)
 
         if input_path.suffix not in self.supported_template_extensions:
             raise UnsupportedTemplateFileTypeError(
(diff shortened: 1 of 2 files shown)


zxiiro (approved):

sceptre/template_handlers/file.py
```diff
@@ -50,16 +51,3 @@ def handle(self):
...
-        """
-        Return the project_path joined to template_path as
-        a string.
-
-        Note that os.path.join defers to an absolute path
-        if the input is absolute.
-        """
-        return path.join(
-            self.stack_group_config["project_path"], "templates",
-            normalise_path(template_path)
```
zaro0508, line 64: wasn't this put in to make template handler work with windows paths?

alexharv074: Perhaps @jfalkenstein could comment. At the time, I thought `normalise_path` was only required because I had moved some legacy code to the template handler class. Looking at it more closely, the new code may also have needed `normalise_path` to be applied to the `path` argument. What if I apply this patch here:

```diff
--- a/sceptre/template_handlers/file.py
+++ b/sceptre/template_handlers/file.py
@@ -4,7 +4,7 @@ import sceptre.template_handlers.helper as helper
 from pathlib import Path
 from sceptre.exceptions import UnsupportedTemplateFileTypeError
 from sceptre.template_handlers import TemplateHandler
-
+from sceptre.helpers import normalise_path
 
 class File(TemplateHandler):
     """
@@ -25,7 +25,7 @@ class File(TemplateHandler):
 
     def handle(self):
         project_path = self.stack_group_config.get("project_path")
-        input_path = Path(self.arguments["path"])
+        input_path = Path(normalise_path(self.arguments["path"]))
         if input_path.is_absolute():
             path = str(input_path)
         else:
```

Alternatively, I can just abort this part of the change and go with the version of the code that was already merged. However, I feel that I added unnecessary complexity.

jfalkenstein: normalise_path takes a "sceptre_path" and turns it into an os-compatible path. This is important in the case of windows, where the user (per Sceptre's documentation) provides a template path like "directory/my-stack-config.yaml". Normalize_path will convert that to be windows-compatible. This is important for windows support.

jfalkenstein: So yeah, I think we should probably still use it. Honestly... It's not clear to me why we're trying to put the code _back_ to it's previous state. The current state of this handler works well and accounts for all the edge cases we've discussed over the last week.

alexharv074: @jfalkenstein Ok, how about we just add the tests on this branch.

zaro0508 (approved):

Issue #1114, opened 2021-09-26
Sceptre should manage sceptre-required resources


jfalkenstein:
### Subject of the issue
One of the "weird" things about the way Sceptre works right now is how it works with some settings like...
* `template_bucket_name` (It will create an S3 buckets for your templates if it doesn't exist already)
* `iam_role`: The role used to execute sceptre actions on a stack group
* `role_arn`: The Cloudformation service role to be assigned to the stack upon creation
* `notification_arns`: Arns for SNS topics to notify about stack actions

All of these are AWS service resources that Sceptre can operate with, but Sceptre doesn't (smoothly) support using Sceptre to create those things. They are either automatically created as byproducts of the first use (template_bucket_name) or they just have to already exist.

Here's what I mean:
* Only notification_arns is a resolvable property, so you can't use `!stack_output` to establish dependencies to those configurations, so you can't you pipe the outputs of other stacks into those settings.
* `template_bucket_name` really expects to be set at the highest level config (though it isn't strictly necessary), so you can't have a `template_bucket_name` set if you want to create a cloudformation template to create your template bucket.
* You can't `unset` template_bucket_name on a stack where it's been set at the project-level by setting the value to `null` or `~`. I tried doing this to define the template bucket in a stack config, but that doesn't work. It just blows up.
* You could get a little creative about having one stack group for the "sceptre necessities" and one for everything else, but then you have to set those "top-level" configurations one level down and duplicate it in a bunch of places
* **Sceptre doesn't currently support setting stack dependencies at the StackGroup level; It errors out when you try.** You can only set dependencies on individual stacks.

In other words, the only way to have Sceptre manage the sceptre-used dependencies for these things feels pretty hacky. And it isn't foolproof. You need to hardcode duplicated values between stack configurations to make that work and set up explicit dependencies on _every...single...stack_. And I can't really figure out a way to manage the template bucket via Sceptre very well.

### What I think could make Sceptre a lot better on this front
Other infrastructure management tools (I'm thinking SAM CLI) have a command that can generate cloudformation stack that defines the required infrastructure to run SAM CLI, so it's like a secondary shared stack. I think this is pretty slick.

[...]

PR #847, opened 2019-09-26
WIP [Resolve #426] resolver support for stack_tags

zaro0508:
This change is to add resolver functionality to stack_tags.  The
idea behind this implementation is to make it work exactly like
sceptre_user_data.  Not only make tags resolvable but also make
them available to jinja templates.

PR #1152, opened 2021-11-07
[Resolve #1114,#426] Project Dependencies part 1: ResolvableProperty refactor + tags resolvable property

jfalkenstein:
This is the first in a series of pull requests that addresses #1114 , allowing Sceptre to manage its own dependencies.

## In this PR:
* **ResolvableProperty is refactored into a base class with a single subclass (for now): ResolvableContainerProperty.** _This refactor will be necessary so we can build on that base class to make a ResolvableValueProperty in a later PR so we can resolve single values like template_bucket_name, etc..._. 
* **The merge strategies list_join and dict_merge now perform deepcopy on the product of those joins,** preventing changes in child stack configs from modifying the same dict/list in memory that has been inherited by other child stack configs. This change is necessary so inheriting mutable values from parent configs becomes a lot safer.
* **A few bugs have been fixed in ResolvableProperty** in this refactor in preparation for changes that are to come in future Pull requests. These all have to do with being able to inherit resolvers and having more resolvers on stack configurations. 
    - Previously, the `_no_recursive_get` method was effectively blocking access to the same property on a DIFFERENT stack. If a resolver accesses the same property on another stack, there shouldn't be a problem with this. This is relevant in cases like the StackOutput resolver that accesses _another_ stack's attributes. This bug has been fixed now. `_no_recursive_get` specifically disallows recursive gets on the _current stack only_.
    - If a resolver was set on a StackGroup config and then "inherited" by multiple stack groups, the _exact same_ object in memory would now be on two stacks. Worse, they would both have the same Stack object assigned to them... which means that Stack A could use a resolver that was referencing Stack B. This has now been solved by resolvers being cloned when they are assigned to a stack, guaranteeing that every stack has an independent resolver in memory.
    - Deferred resolution is properly handled _within_ the property. Previously it wasn't, so we needed a weird double-layer property on the Stack class to handle sceptre_user_data. This is now no longer necessary and is handled where it should have been handled all along: in the property itself.
* **stack_tags is now a ResolvableProperty**, ultimately finishing what #847 attempted but never finished.

## PR Checklist

## Other Information

### DIFF
--- a/integration-tests/steps/helpers.py
+++ b/integration-tests/steps/helpers.py
@@ -73,7 +73,7 @@ def get_cloudformation_stack_name(context, stack_name):
 
 
 def retry_boto_call(func, *args, **kwargs):
-    delay = 2
+    delay = 5
     max_retries = 150
     attempts = 0
     while attempts < max_retries:
--- a/sceptre/config/strategies.py
+++ b/sceptre/config/strategies.py
@@ -6,6 +6,7 @@ sceptre.config.strategies
 This module contains the implementations of the strategies used to merge config
 attributes.
 """
+from copy import deepcopy
 
 
 def list_join(a, b):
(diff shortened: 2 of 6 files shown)


zaro0508 (approved):
only a few minor issues otherwise LGTM

Issue #886, opened 2020-03-20
Enable the use of !stack_output for stack_config role_arn


schlichtanders:
Dear sceptre team,

I am currently in the process of deploying EKS with sceptre and it turns out that AWS EKS cloudformation has the specificality that only the ARN which created the cluster can also change it.

Hence this is a case for sceptre's stack_config ``role_arn``. However the ``role_arn`` is defined within another stack, hence the intuitive thing todo would be
```
# my config.yaml for the stack create-eks-cluster.yaml
template_path: create-eks-cluster.yaml
role_arn: !stack_output create-role-arn.yaml::RoleArn
```

However this gives the following error:
```
Parameter validation failed:\nInvalid type for parameter RoleARN, value: <sceptre.resolvers.stack_output.StackOutput object at 0x7fc4183a99d0>, type: <class 'sceptre.resolvers.stack_output.StackOutput'>, valid types: <class 'str'>
```
which indicates that !stack_output is not  supported currently for ``role_arn`` (but only for ``parameters:`` I guess)

craighurley:
Resolvers (e.g. stack_output) only work in [Stack Config files](https://sceptre.cloudreach.com/2.3.0/docs/stack_config.html).  I believe you are trying to use a resolver in a [StackGroup Config file](https://sceptre.cloudreach.com/2.3.0/docs/stack_group_config.html).

craighurley:
_I deleted my previous comment as my response was incorrect._

From the [Resolvers](https://sceptre.cloudreach.com/2.3.0/docs/resolvers.html) doc:

> Sceptre implements resolvers, which can be used to resolve a value of a CloudFormation parameter or sceptre_user_data value at runtime. 

i.e. not for `role_arn`, although, it would be useful especially in your case.

schlichtanders:
I made a minimal implementation, missing tests and everything
I really don't know whether I can find the time to add the rest for a full pull-request,
I hope that even if not, this implementation would help someone else to complete the pull-request

(the implementation is working for me)

schlichtanders:
just checked ``iam_role`` because it apparently is quite similar.

Unfortunately the required amount of code-changes are much larger, because ``iam_role`` triggers the creation of a ``ResolveLater`` which seems really hard right now to support plain non-dict, non-list arguments. Maybe the ResolveLater logic would be good to review if ``iam_role`` should also be supported.

Luckily `role_arn` works super fine, and does not trigger any ResolveLater at all.

[...]

Issue #491, opened 2018-10-25
template bucket encryption


craighurley:
```
$ sceptre --version
Sceptre, version 1.4.2
```

When using `template_bucket_name` (https://sceptre.cloudreach.com/latest/docs/environment_config.html#template_bucket_name), the bucket is automatically created by sceptre if it does not already exist:

> If the bucket does not exist, Sceptre creates one using the given name, in the AWS region specified by region.

... it would be nice if server side encryption was enabled by default; got to keep ticking those encryption boxes to keep security folks happy :)

ngfgrant:
Hi @craighurley yeh I agree happy to look over a PR for this.

ngfgrant:
Closing as #730 gives a more broad template bucket config option including encryption.

jfalkenstein:
The spirit of this ticket has been acomplished now that the template_bucket_name is resolvable, from #1153 .

PR #1153, opened 2021-11-07
[Resolve #1114,#886,#491] Project Dependencies part 2: Resolvable role_arn and template_bucket_name

jfalkenstein:
This is the second in a series of pull requests that addresses #1114 , allowing Sceptre to manage its own dependencies.

## In this PR:
* The ResolvableValueProperty is created, creating a property that can resolve to a single value (not a list or dict)
* role_arn and template_bucket_name are now fully resolvable properties
    - role_arn also resolves the issue #886 
    - template_bucket_name also provides a solution to the issue #491 
* If template_bucket_name is set to None on a stack, it will be interpreted as if it wasn't there at all. _This will be important in future Pull Requests so that if template_bucket_name is inherited from the stack_group, it won't actually create a circular dependency on the stack that outputs that bucket name._
* Documentation is clarified on setting stack dependencies on a StackGroup to be shared across all stacks in that group as well as a warning about circular dependencies.

## PR Checklist

## Other Information


jfalkenstein:
@schlichtanders, This addresses what you requested in #886.

@craighurley This addresses what you requested in #491 (well, it makes it possible to setup whatever kind of bucket you want).

sceptre/stack.py
```diff
@@ -116,7 +116,12 @@ class Stack(object):
     parameters = ResolvableContainerProperty("parameters")
     sceptre_user_data = ResolvableContainerProperty("sceptre_user_data")
     notifications = ResolvableContainerProperty("notifications")
-    tags = ResolvableContainerProperty('tags')
+    tags = ResolvableContainerProperty("tags")
+
+    s3_details = ResolvableContainerProperty("s3_details")
```
zaro0508, line 121: why is this a resolvable property?  it's not a key in stack configs is it?

jfalkenstein: So the template_bucket_name is accessed off of this property in the template class. So, in order to make template_bucket_name resolvable in all cases, the s3_details needed to be resolvable too.

sceptre/config/reader.py
```diff
@@ -475,7 +475,9 @@ def _collect_s3_details(stack_name, config):
         :rtype: dict
         """
         s3_details = None
-        if "template_bucket_name" in config:
+        # If the config explicitly sets the template_bucket_name to None, we don't want to enter
+        # this conditional block.
+        if config.get("template_bucket_name") is not None:
```
zxiiro, line 480: I think we can shorten this to `if config.get("template_bucket_name"):`?

I believe config.get() defaults to `None` if it's not found and if template_bucket_name is explicity None then the condition wouldn't enter the block.

sceptre/template.py
```diff
@@ -215,7 +215,8 @@ def get_boto_call_parameter(self):
         :returns: The boto call parameter for the template.
         :rtype: dict
         """
-        if self.s3_details:
+        # If bucket_name is set to None, it should be ignored and not uploaded.
+        if self.s3_details and self.s3_details.get("bucket_name") is not None:
```
zxiiro, line 219: I believe the `is not None` can be shortend to simply `if self.s3_details and self.s3_details.get("bucket_name")` here?

zxiiro (commented):
Really minor comment really about the `is not None` in if conditions. Either way I think this looks fine to me.

sceptre/config/reader.py
jfalkenstein: You're right that None is falsey. However, I like to try to be as explicit in these checks as possible, since other values could be falsey too, like empty strings. I suppose in this case, we might want to treat an empty string the same as `None`. Nonetheless, when one is not explicit about what they're checking for, it can lead future maintainers to wonder about what sorts of values could be expected here.

jfalkenstein:
Alright @zaro0508 , @zxiiro this PR is now fully ready for review. You both had made some comments on this PR previously (and I've responded to those). I've added in proper docs and integration tests on this PR, so it should now be ready for review.

**This PR can be merged and deployed to master without requiring any changes from parts 3 and onward**. However, this PR _is_ pointed at the Part 1 branch because it builds on the changes in that one.

docs/_source/docs/stack_group_config.rst
```diff
@@ -66,6 +66,13 @@ supplies the template to Boto3 via the ``TemplateBody`` argument. Templates
 supplied in this way have a lower maximum length, so using the
 ``template_bucket_name`` parameter is recommended.
 
+.. warning::
+
+   If you resolve ``template_bucket_name`` using the ``!stack_output``
+   resolver on a StackGroup, the stack that outputs that bucket name *cannot* be
+   defined in that StackGroup. Otherwise, a circular dependency will exist and Sceptre
+   will raise an error when attempting any Stack action.
```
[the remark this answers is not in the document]
jfalkenstein: Yeah, I can add another line indicating the correct way to do it.

jfalkenstein:
@zaro0508 , I gave two examples of project structure supporting referencing project dependencies with `!stack_output`.

zaro0508 (approved):

PR #1155, opened 2021-11-07
[Resolve #1114] Project Dependencies part 4: Identifying stack as project dependency

jfalkenstein:

## In this PR:
* **Stack Configs can be marked with `is_project_dependency: True`**. This is important, as it allows us to bypass the irreconcilable issue of multiple project dependency stacks depending on each other. with `is_project_dependency: True`, it will have the effects of:
    - Disabling all dependencies on that stack. They cannot be set on the current stack and they can't be inherited from parent StackGroups. In other words, a "project dependency" stack is automatically at the top of the dependency hierarchy.
    - Disabling the StackOutput resolver. It will "resolve to nothing" on project dependency stacks (see below). This means that it can be safely inherited from a StackGroup, but it won't actually do anything.
* **Resolvers that return None "resolve to nothing",** being removed from their container, just like "AWS::NoValue" does in CloudFormation templates. This prevents issues like when every stack in the project "inherits" `template_bucket_name: !stack_output bucket.yaml::BucketName`, and that would include the bucket.yaml stack, which cannot rely on itself.

## Other Information

### DIFF
--- a/sceptre/stack.py
+++ b/sceptre/stack.py
@@ -108,6 +108,11 @@ class Stack(object):
             will result in no timeout. Supports only positive integer value.
     :type stack_timeout: int
 
+    :param is_project_dependency: Indicates whether or not the the stack is a\
+            special dependency of the stack. If True, disables all dependencies\
+            on the the current stack.
+    :type is_project_dependency: bool
+
     :param stack_group_config: The StackGroup config for the Stack
     :type stack_group_config: dict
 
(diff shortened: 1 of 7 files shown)


sceptre/stack.py
```diff
@@ -149,7 +154,7 @@ def __init__(
         self.external_name = external_name or get_external_stack_name(self.project_code, self.name)
         self.template_path = template_path
         self.template_handler_config = template_handler_config
-        self.s3_details = s3_details
```
jfalkenstein, line 152: Looks like this was being double-assigned, so this just removes it.

Issue #1000, opened 2021-04-09
stack_output resolution fails when used in sceptre_user_data


joe1981al:
## issue
Will get "Stack with id <id> does not exist" when a stack_output resolver is used in the sceptre_user_data section of config and points to another config file.  This only occurs when the stack being referenced is not actually running yet.

When using the same stack_output line in the parameter section it understands that the other stack is a dependency and does not through an error.

## expected behavior
I would expect the same behavior between parameters and sceptre_user_data sections of the template when using stack_output resolver.

craighurley:
Hi @joe1981al , it always helps to include example code and the error you get; can you please add?

joe1981al:
@craighurley  https://github.com/joe1981al/sceptre_error_example

jfalkenstein:
I actually ran into this bug very recently. The issue results from how (unlike parameters), sceptre_user_data is accessed to pass in to actually render the templates for jinja and python based templates. 

I had to implement a custom hook that allows you to replace the sceptre_user_data on the stack before_validate and before_generate. It's a bit of a hacky workaround, but it works. It does seem like we might want a better option. 

I think a viable approach to this would be a toggle on the stack (or maybe a globally available variable?) that causes the StackOutput resolver to catch that error and instead return a default, like "<StackOutput('Vpc::VpcId')>", but ONLY when doing certain non-creation activities that require a template, like validation, generation, etc...

sopeters:
I have this issue with the `validate` command even when the stack exists - getting
`stack for stack in self.stack.dependencies if stack.name == friendly_stack_name`

Will the PR fix this as well?

jfalkenstein:
It's not clear what that error is that you're describing. Could you include more detail?

jfalkenstein:
This issue has been generally resolved (as best as possible) with the use of placeholders now that #1185 has been merged.

PR #1156, opened 2021-11-07
[Resolve #1114,#1000] Project Dependencies part 5: Handling resolvers referencing non-deployed stacks in template commands

jfalkenstein:

## In this PR:
* It is possible to enter a global state (using a context manager) that allows Resolvers that raise errors to resolve to a placeholder value. This is explicitly opt-in and will only be enabled on commands that don't actually deploy stacks, such as "generate", "validate", and "diff". This is required so that these commands don't fail when there are resolvers that reference stacks that aren't deployed yet. This ultimately is the bit that resolves issue #1000. This is important because this series of Pull requests more than doubles the number of stack attributes that are resolvable, which are going to be all the more problematic when using `sceptre validate`, etc...

## PR Checklist

## Other Information


jfalkenstein:
@joe1981al This addresses the concern you raised in #1000

sceptre/cli/template.py
```diff
@@ -66,7 +87,11 @@ def generate_command(ctx, path):
     )
 
     plan = SceptrePlan(context)
-    responses = plan.generate()
+
+    execution_context = null_context() if no_placeholders else use_resolver_placeholders_on_error()
```
zaro0508, line 91: it looks like validate and diff contains a more verbose version for this logic.  just curious, any reason why it's different here?

jfalkenstein: A previous version had parameters passed to use_resolver_placeholders_on_error but then they were removed. I can make them uniform now, though.

jfalkenstein:
I'm going to close this PR in favor of #1185 since I needed to recut that branch properly.

PR #1157, opened 2021-11-07
[Resolve #1114] Project Dependencies WIP: Docs and Integration tests

jfalkenstein:
## This PR is on Hold
In discussion with @zaro0508 and @zxiiro , we decided that what remains of this pull request would be put on hold until the work with resolvers was fully handled.

This is the sixth in a series of pull requests that addresses #1114 , allowing Sceptre to manage its own dependencies.

## In this PR:
* Documentation detailing the changes to resolvers, project dependencies, how to manage project dependencies, etc...
* Integration tests on the project dependencies.

## PR Checklist

## Other Information


docs/_source/conf.py
```diff
@@ -42,6 +42,7 @@
 # ones.
 extensions = [
     'sphinx.ext.autodoc',
+    'sphinx_autodoc_typehints',
```
jfalkenstein, line 45: This allows sphinx to better "understand" type annotations and populate those types in the docs. Without this, sphinx was logging warnings when I'd generate docs and wasn't generating docs that were properly indicating the types.

docs/_source/conf.py
```diff
@@ -199,3 +200,5 @@
     ('py:class', 'json.encoder.JSONEncoder'),
     ('py:class', 'sceptre.config.reader.Attributes')
 ]
+
+set_type_checking_flag = True
```
jfalkenstein, line 204: This allows Sphinx to "understand" type annotations where the import is declared in an if TYPE_CHECKING condition.

docs/_source/docs/resolvers.rst
```diff
@@ -156,15 +162,23 @@ custom_resolver.py
...
+            def __init__(self, argument, stack=None):
+                super(CustomResolver, self).__init__(argument, stack)
+
+            def setup(self):
+                """
+                Setup is invoked after the stack has been set on the resolver, whether or not the
+                resolver is ever resolved.
+
+                Implement this method for any setup behavior you want (such as adding to stack dependencies).
+                """
```
jfalkenstein, line 181: These docs were not correct for resolvers. I have fixed them.

docs/_source/docs/stack_config.rst
```diff
@@ -206,15 +236,25 @@ CI/CD system like Jenkins.
 In order to use this argument, however, the role needs to have an AssumeRolePolicyDocument that
 permits the user to assume that role.
 
+**Important**: If you use set the value of the ``iam_role`` with ``!stack_output``, that `iam_role`
+will not actually be used to obtain the stack_output, but it WILL be used for all subsequent stack
+actions. Therefore, it is important that the user executing the stack have permissions to get stack
+outputs for the stack outputting the ``iam_role``.
```
zaro0508, line 242: I don't use iam_role so maybe it's making it difficult for me to fully understand this.  Is this explaining a scenario for a  bootstrapping step where the admin user is using sceptre to deploy their CFN stacks for the 1st time?

docs/_source/docs/stack_group_config.rst
```diff
@@ -144,6 +157,39 @@ For example, if you wanted the ``dev`` StackGroup to build to a different
...
+``project_code`` and ``region``. Doing so will make it so that every stack in the StackGroup (or
+even in the entire project) will be have those dependencies and get those values from Sceptre-managed
+stacks.
+
+Beyond the above mentioned config keys, it is possible to set the ``dependencies`` config key in a
+StackGroup config to be inherited by all Stack configs in that group. All dependencies in child
+stacks will be added to their inherited StackGroup dependencies, so be careful how you structure
+dependencies.
+
+**Important**: You might have already considered this might cause a circular dependency for those
```
zaro0508, line 185: considered this/considered that this

docs/_source/docs/stack_config.rst
```diff
@@ -248,17 +288,56 @@ referring to is in a different AWS account or region.
...
+**the stack_output resolver will always resolve to nothing on project dependency stacks.**
+
+For example, this allows you to safely set the ``template_bucket_name`` on the top-level StackGroup
+config as the output from a project_dependency stack. This means that every **other** stack not marked
+as a project dependency will use that template_bucket_name, while the template bucket stack will
+have NO template bucket name (i.e. its template will not be uploaded anywhere).
+
+Another example would be defining the role_arn on the top-level stack config with the ``!stack_output``
+of a project dependency stack that outputs the role arn. Every **other** stack will use this stack
+as a CloudFormation service role, but project dependency stacks will have no service role.
```
zaro0508, line 332: It's difficult to get the jist of all of this without some concrete template examples.  would be good to add examples to cover some of the most common use cases.

docs/_source/docs/resolvers.rst
```diff
@@ -156,15 +162,23 @@ custom_resolver.py
...
 
-            def __init__(self, *args, **kwargs):
-                super(CustomResolver, self).__init__(*args, **kwargs)
+            def __init__(self, argument, stack=None):
+                super(CustomResolver, self).__init__(argument, stack)
+
+            def setup(self):
+                """
+                Setup is invoked after the stack has been set on the resolver, whether or not the
+                resolver is ever resolved.
```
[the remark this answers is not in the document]
jfalkenstein: `setup()` is invoked when the value is assigned to the stack. Consider the case where we run `sceptre validate`. In that case, any resolvers in the parameters wouldn't be actually _resolved_, but they will be _assigned_ to the stack's parameters.

docs/_source/docs/stack_config.rst
jfalkenstein: Exactly! That's exactly my own use case, actually.

docs/_source/docs/stack_config.rst
jfalkenstein: I have been thinking that the "sample architectures" page could use an overhaul. The changes from these PRs seem like they really could improve that. I'll work this weekend, I think, on adding to that page's docs with concrete examples of the sort of setup that is now supported. I can link to there from here.

jfalkenstein: I added a detailed example project (something we never really had and probably needed anyway) in the "Architecting Sceptre" page: https://github.com/Sceptre/sceptre/pull/1157/files#diff-498f17109556dd683849428994e46e8731d43ab7fff3688fd78d5fd4591003c2

This should make things a lot clearer, I think.

docs/_source/docs/architecture.rst
```diff
@@ -135,15 +135,222 @@ to demonstrate ways of organising different projects.
...
+    # bucket, preventing conflicts.
+    template_key_prefix: {{ var.environment }}
+
+    # Setting the iam_role on the project like this means that all Sceptre actions on all stacks in
+    # the project will be done by assuming the deployment role whose ARN is output by the deployment
+    # stack (except for actions on the deployment stack... see its comments below).
+    iam_role: !stack_output project/deployment.yaml::DeploymentRoleArn
+
+
+The Deployment Stack
```
zaro0508, line 227: The order of describing these files is a bit confusing since config/config.yaml depends on config/project/deployment.yaml.  Would it make sense to put the description of the deployment stack above config/config.yaml?

docs/_source/docs/architecture.rst
```diff
@@ -135,15 +135,222 @@ to demonstrate ways of organising different projects.
...
+    project_code: indigo-deployment
+
+    # Setting is_project_dependency has a few important effects for this stack
+    #   1. It disables dependencies. In other words, this stack cannot depend on any other stack.
+    #       This is how it bypasses the circular dependency that it would otherwise inherit from
+    #       the project's config (config/config.yaml)
+    #   2. It disables the !stack_output resolver, making it resolve to nothing. In other words, it
+    #       cannot inherit the template_bucket_name or iam_role from the project config. This is
+    #       good because this stack DEFINES the template bucket and iam_role.
+    is_project_dependency: True
```
zaro0508, line 250: having to define this key feels kinda clunky to me.  I'm wondering if this should actually be a template type instead?   Something like a `project` or `base` type so we can identify it with something like this..

```
template:
  type: project    # or a "base" type?
  path: project/deployment.yaml
``` 

Another option is to identify it with a special filename, like `project.yaml` or `base.yaml`?

docs/_source/docs/architecture.rst
```diff
@@ -135,15 +135,222 @@ to demonstrate ways of organising different projects.
...
+    # config/project/deployment.yaml
+
+    # This is a StackConfig for a stack that defines the project's basic requirements, that all
+    # stacks and all environments will depend on for deployment to work. Because the project config
+    # references this stack using !stack_output, all OTHER stacks in the project will depend upon
+    # this one.
+
+    # We override the project_code here so that there will only be ONE stack for deployment that is
+    # shared across all environments.
+    project_code: indigo-deployment
```
zaro0508, line 241: isn't project_code in config/config.yaml overriding this since config/config.yaml depends on this template?

docs/_source/docs/architecture.rst
```diff
@@ -135,15 +135,222 @@ to demonstrate ways of organising different projects.
...
+    # By making the project-code parameterized by environment name, it guarantees that one
+    # environment's stacks do not conflict with another environment's stacks.
+    project_code: indigo-{{ var.environment }}
+    region: {{ var.region }}
+
+    # Setting the template_bucket_name with a !stack_output means that every stack in the project
+    # (except the deployment stack) will have its template uploaded to that bucket. It ALSO means
+    # that every stack in the project will inherit a dependency on the deployment stack (except the
+    # deployment stack... see its comments below)
+    template_bucket_name: !stack_output project/deployment.yaml::BucketName
```
zaro0508, line 214: BucketName isn't defined in the example project/deployment.yaml so how does this dependency work?

docs/_source/docs/architecture.rst
```diff
@@ -135,15 +135,222 @@ to demonstrate ways of organising different projects.
...
+
+    # By making the template_key_prefix parameterized by environment, it means that every
+    # environment's templates will be uploaded to its own directory structure within the template
+    # bucket, preventing conflicts.
+    template_key_prefix: {{ var.environment }}
+
+    # Setting the iam_role on the project like this means that all Sceptre actions on all stacks in
+    # the project will be done by assuming the deployment role whose ARN is output by the deployment
+    # stack (except for actions on the deployment stack... see its comments below).
+    iam_role: !stack_output project/deployment.yaml::DeploymentRoleArn
```
zaro0508, line 224: DeploymentRoleArn isn't defined in the example project/deployment.yaml so how does this dependency work?

docs/_source/docs/architecture.rst
```diff
@@ -135,15 +135,222 @@ to demonstrate ways of organising different projects.
...
+
+The Deployment Stack
+^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: yaml
+
+    # config/project/deployment.yaml
+
+    # This is a StackConfig for a stack that defines the project's basic requirements, that all
+    # stacks and all environments will depend on for deployment to work. Because the project config
```
zaro0508, line 235: Because the project config/Because the stack config

docs/_source/docs/architecture.rst
```diff
@@ -135,15 +135,222 @@ to demonstrate ways of organising different projects.
...
+    #   1. It disables dependencies. In other words, this stack cannot depend on any other stack.
+    #       This is how it bypasses the circular dependency that it would otherwise inherit from
+    #       the project's config (config/config.yaml)
+    #   2. It disables the !stack_output resolver, making it resolve to nothing. In other words, it
+    #       cannot inherit the template_bucket_name or iam_role from the project config. This is
+    #       good because this stack DEFINES the template bucket and iam_role.
+    is_project_dependency: True
+
+    # This template defines an S3 Bucket, outputting the created bucket as "BucketName" and an IAM
+    # role that can be assumed by the CI/CD system's role/user, outputting the ARN as
```
zaro0508, line 253: where's the definition of the S3 bucket and IAM role?

docs/_source/docs/architecture.rst
```diff
@@ -135,15 +135,222 @@ to demonstrate ways of organising different projects.
...
+  conflict with production. These environments will have some different configurations to make them
+  cheaper for the company, but otherwise be identical to production.
+
+Project Structure
+*****************
+
+.. code-block:: text
+
+    config/
+        config.yaml             # top-level project configuration
```
zaro0508, line 175: I think the docs already have a special name for this config, it's the StackGroup config correct, not project config?  IMO, keeping the naming consistent will help users.

docs/_source/docs/architecture.rst
```diff
@@ -135,15 +135,222 @@ to demonstrate ways of organising different projects.
...
+***************
+
+Top-level project configuration
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: yaml
+
+    # config/config.yaml
+
+    # This is the top-level config file for the whole project. Every stack in the project will
```
zaro0508, line 200: top-level config/top-level StackGroup config

zaro0508 (commented):
thanks for the detailed example however I think it's a bit confusing on how it describes the file dependencies.  It's not clear to me whether config/config.yaml depends on config/project/deployment.yaml or the other way around.

docs/_source/docs/architecture.rst
jfalkenstein: I don't know if re-ordering would be easier to understand. There's partially a circular dependency here (broken by the is_project_dependency stack config value). One way or another, we need to "break into the circle". I'll reread through this and try to find a better order/way to explain things.

Basically, here's the scenario I'm trying to indicate:
* All stacks in the project "inherit" values from config/config.yaml. Included in this are the dependencies on the deployment stack.
* The deployment stack (which is in the project) disallows dependencies; Instead, it only _provides_ dependencies for the other stacks in the project.

docs/_source/docs/architecture.rst
jfalkenstein: > having to define this key feels kinda clunky to me.

I've only defined it here in the docs for explanatory purposes, as I've explained _all_ the things. This config has been documented elsewhere on the Stack Config docs. Could you elaborate on what you mean by this?

Regarding a special template type, I don't think that option would be viable; the distinction isn't in the _template_, it's in the relation of this stack to the other stacks. Also, `is_project_dependency` doesn't do anything with the stack template. Instead, it alters the way dependencies are handled and the way the !stack_output resolver operates. 

Regarding the "special filename" approach you suggest, that seems to indicate a completely different class of stack. It also would also likely prohibit multiple stacks that are indicated as project dependencies. I'm not crazy about the idea of the config reader needing to check the stack name for every stack, and if it's "project" or "base", instantiating a different kind of stack or something like that.

Rather than any of those, perhaps we could rename the configuration to `disable_dependencies`. Would you find that more self-explanatory?

docs/_source/docs/architecture.rst
jfalkenstein: I think you're thinking about stack group configs incorrectly. Stack group configs (such as config/config.yaml) don't actually "exist" in Sceptre. They don't depend on anything. Rather they are layers of key/value pairs that stacks in their groups _inherit_. Stack configs layer _on top of_ stack group configs.

In other words, if this Stack didn't override the project_code, it would inherit the project code from stack group in which it resides (config/project). Because config/project doesn't override it, then that value would be inherited from config/config.yaml. But this Stack _does_ override the project-code, so the project code used for this stack is indeed the one here.

docs/_source/docs/architecture.rst
jfalkenstein: You're right that it's a StackGroup. But the top level StackGroup is referred to as the project (thus the "project_code" configuration key). I'll clarify this a little better, though.

docs/_source/docs/architecture.rst
jfalkenstein: It's a `!stack_output` resolver, so it's referencing an output on the _template_ of that stack. Remember, outputs are not defined on stack configs; they're on the templates _referenced_ by stack configs.

docs/_source/docs/architecture.rst
jfalkenstein: Again, this is a `!stack_output` resolver. See my comment above. Outputs don't go on stack configs; they go on templates.

docs/_source/docs/architecture.rst
jfalkenstein: I'll fix this. I'm referring to the project StackGroup config.

docs/_source/docs/architecture.rst
jfalkenstein: It's in the template. I didn't include the template in these docs. Are you suggesting I should include a sample CloudFormation template for these?

docs/_source/conf.py
```diff
@@ -204,5 +205,11 @@
     ('py:class', 'sceptre.config.reader.Attributes'),
     ('py:class', 'sceptre.diffing.stack_differ.DiffType'),
     ('py:obj', 'sceptre.diffing.stack_differ.DiffType'),
-    ('py:class', 'TextIO')
+    ('py:class', 'DiffType'),
+    ('py:class', 'TextIO'),
+    ('py:class', '_io.StringIO'),
+    ('py:class', 'yaml.loader.SafeLoader'),
+    ('py:class', 'yaml.dumper.Dumper')
```
zxiiro, line 212: I'd always end the line with `,` in Python. That way when someone adds another line later in this list it doesn't show up as a 2 line change. For example the TextIO line above would have been unmodified if it already had the trailing comma.

docs/_source/docs/architecture.rst
```diff
@@ -11,7 +11,7 @@ our use of Sceptre at Cloudreach.
 Project layout
 --------------
 
-Sceptre’s nested StackGroups means that it is possible to store an entire
+Sceptre’s nested stack groups means that it is possible to store an entire
```
zxiiro, line 14: Why was this changed here to be 2 separate words but I've seen some new instances below write them as `StackGroups` one word?

docs/_source/docs/architecture.rst
zxiiro: I agree that `is_project_dependency` does feel clunky as it's a way of marking a special template. But I can't think of a better way to handle this at the moment.

As for the naming I think having a name that self explains what it does would be more useful as `is_project_dependency` makes me immediately have to look for docs. `disable_dependencies` is a bit more clear for sure or maybe `disable_stack_dependencies` or `is_root_template` or something along those lines?

docs/_source/docs/architecture.rst
zxiiro: Having a sample template would be useful so that someone could use it as a starting point.

To continue the point about being clunky from above having to define a custom template for this definitely does feel clunky to me, it makes the featuer feel like an advanced feature that an expert has to hand create. I wonder if we could provide some kind of minimal out of the box template that just works without the user having to create anything and they can "override" it by creating a file or something or parameters they can pass to modify the template.

docs/_source/conf.py
jfalkenstein: That can be done.

docs/_source/docs/architecture.rst
jfalkenstein: Yeah, great question. Our docs and docstrings seem to be wholly inconsistent on this point. I think we could benefit from a "vocab guide" or something for contributing. It _does_ appear that "StackGroup" is more common than "stack group", though. I'll fix all occurrances in the files I've touched . 

image
image

docs/_source/docs/architecture.rst
jfalkenstein: @zxiiro One of the things I suggested in #1114 was an auto-generation feature. I figured after all of these PRs got merged, I would start working on that. I know, for example, that `sam bootstrap` will auto-generate a stack that contains all the CloudFormation resources necessary to deploy the SAM template using a CI/CD pipeline. I really like that functionality and I think it could be implemented fairly simply with Sceptre.

Really, at minimum, that "bootstrapped" stack would just need to contain a simple s3 bucket for templates, optionally an IAM Role for deployment. It would actually be pretty simple.

PR #1168, opened 2021-11-25
[Resolve #1167] Add list stacks command

### LINKED ISSUE #1167: Feature request: stack-name command
### Feature request

Please implement a feature `stack-name` that returns the stack's stack name in CloudFormation. This is given by `self.external_name` in the `Stack` object. Where ambiguity would exist in the case of nested stacks, also return a mapping of stack names => command paths.

alexharv074:
Implements a feature `list stacks` that returns the stack's
stack name in CloudFormation.

This is given by self.external_name in the Stack object.

Where ambiguity would exist in the case of nested stacks,
this also return a mapping of stack names => command paths.

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/cli/helpers.py
+++ b/sceptre/cli/helpers.py
@@ -113,27 +113,34 @@ def _generate_json(stream):
 
 
 def _generate_yaml(stream):
-    if isinstance(stream, list):
+    kwargs = {
+        "default_flow_style": False,
+        "explicit_start": True
+    }
+
+    if type(stream) in [list, set]:
         items = []
         for item in stream:
             try:
                 if isinstance(item, dict):
                     items.append(
-                        yaml.safe_dump(item, default_flow_style=False, explicit_start=True)
+                        yaml.safe_dump(item, **kwargs)
                     )
                 else:
                     items.append(
                         yaml.safe_dump(
-                            yaml.load(item, Loader=CfnYamlLoader),
-                            default_flow_style=False, explicit_start=True
+                            yaml.load(item, Loader=CfnYamlLoader), **kwargs
                         )
                     )
             except Exception:
                 print("An error occured whilst writing the YAML object.")
         return yaml.safe_dump(
-            [yaml.load(item, Loader=CfnYamlLoader) for item in items],
-            default_flow_style=False, explicit_start=True
+            [yaml.load(item, Loader=CfnYamlLoader) for item in items], **kwargs
         )
+
+    elif isinstance(stream, dict):
+        return yaml.dump(stream, **kwargs)
+
     else:
         try:
             return yaml.safe_loads(stream)
--- a/sceptre/cli/helpers.py
+++ b/sceptre/cli/helpers.py
@@ -361,6 +361,29 @@ def simplify_change_set_description(response):
     return formatted_response
 
 
+def deserialize_json_properties(value):
+    if isinstance(value, str):
+        is_json = (
+            (value.startswith("{") and value.endswith("}"))
+            or
+            (value.startswith("[") and value.endswith("]"))
+        )
+        if is_json:
+            return json.loads(value)
+        return value
+    if isinstance(value, dict):
+        return {
+            key: deserialize_json_properties(val)
+            for key, val in value.items()
+        }
+    if isinstance(value, list):
+        return [
+            deserialize_json_properties(item)
+            for item in value
+        ]
+    return value
+
+
 class ColouredFormatter(logging.Formatter):
     """
     ColouredFormatter add colours to all stack statuses that appear in log
--- a/sceptre/cli/helpers.py
+++ b/sceptre/cli/helpers.py
@@ -118,7 +118,7 @@ def _generate_yaml(stream):
         "explicit_start": True
     }
 
-    if type(stream) in [list, set]:
+    if isinstance(stream, (list, set)):
         items = []
         for item in stream:
             try:
--- a/sceptre/cli/list.py
+++ b/sceptre/cli/list.py
@@ -134,8 +134,7 @@ def list_change_sets(ctx, path, url):
 @catch_exceptions
 def list_stacks(ctx, path):
     """
-    List stack names and command paths for local stacks in a
-    stack group.
+    List sceptre stack config attributes,
     \f
 
     :param path: Path to execute the command on or path to stack group
--- a/sceptre/cli/list.py
+++ b/sceptre/cli/list.py
@@ -152,7 +152,8 @@ def list_stacks(ctx, path, all):
         user_variables=ctx.obj.get("user_variables"),
         output_format=ctx.obj.get("output_format"),
         options=ctx.obj.get("options"),
-        ignore_dependencies=ctx.obj.get("ignore_dependencies")
+        ignore_dependencies=ctx.obj.get("ignore_dependencies"),
+        full_scan=True
     )
     plan = SceptrePlan(context)
 
--- a/sceptre/cli/list.py
+++ b/sceptre/cli/list.py
@@ -135,7 +135,7 @@ def list_change_sets(ctx, path, url):
 @catch_exceptions
 def list_stack_name(ctx, path):
     """
-    List stack name for stack.
+    List stack names for stacks in a stack group.
     \f
 
     :param path: Path to execute the command on.
@@ -152,12 +152,12 @@ def list_stack_name(ctx, path):
 
     plan = SceptrePlan(context)
 
-    chop_output = len(plan.all_stacks) == 1
+    show_paths = len(plan.all_stacks) > 1
     stack_names = []
 
     for stack in plan.all_stacks:
         output = stack.external_name
-        if not chop_output:
+        if show_paths:
             relative_path = normalise_path(f"{stack.name}.yaml")
             output += f" (from {relative_path})"
         stack_names.append(output)
--- a/sceptre/cli/list.py
+++ b/sceptre/cli/list.py
@@ -128,15 +128,7 @@ def list_change_sets(ctx, path, url):
         write(response, context.output_format)
 
 
-@list_group.group(name="local")
-def local_group():
-    """
-    Commands for listing attributes of local stacks.
-    """
-    pass
-
-
-@local_group.command(name="stacks")
+@list_group.command(name="stacks")
 @click.argument("path")
 @click.pass_context
 @catch_exceptions
--- a/sceptre/cli/list.py
+++ b/sceptre/cli/list.py
@@ -152,10 +152,10 @@ def list_stack_name(ctx, path):
 
     plan = SceptrePlan(context)
 
-    chop_output = len(plan.command_stacks) == 1
+    chop_output = len(plan.all_stacks) == 1
     stack_names = []
 
-    for stack in plan.command_stacks:
+    for stack in plan.all_stacks:
         output = stack.external_name
         if not chop_output:
             relative_path = normalise_path(f"{stack.name}.yaml")
--- a/sceptre/cli/list.py
+++ b/sceptre/cli/list.py
@@ -151,5 +151,10 @@ def list_stacks(ctx, path):
     )
 
     plan = SceptrePlan(context)
-    output = [f"{stack.name}.yaml: {stack.external_name}" for stack in plan.all_stacks]
+
+    if context.output_format == "json":
+        output = {f"{stack.name}.yaml: {stack.external_name}" for stack in plan.all_stacks}
+    else:
+        output = [f"{stack.name}.yaml: {stack.external_name}" for stack in plan.all_stacks]
+
     write(output, context.output_format)
--- a/sceptre/cli/list.py
+++ b/sceptre/cli/list.py
@@ -129,10 +129,13 @@ def list_change_sets(ctx, path, url):
 
 
 @list_group.command(name="stacks")
+@click.option(
+    "--all", is_flag=True, help="List all stacks in stack groups."
+)
 @click.argument("path")
 @click.pass_context
 @catch_exceptions
-def list_stacks(ctx, path):
+def list_stacks(ctx, path, all):
     """
     List stack names and command paths for stacks in a
     stack group.
@@ -140,6 +143,8 @@ def list_stacks(ctx, path):
 
     :param path: Path to execute the command on.
     :type path: str
+    :param all: List all stacks in stack groups.
+    :type all: bool
     """
     context = SceptreContext(
         command_path=path,
@@ -149,12 +154,13 @@ def list_stacks(ctx, path):
         options=ctx.obj.get("options"),
         ignore_dependencies=ctx.obj.get("ignore_dependencies")
     )
-
     plan = SceptrePlan(context)
 
-    if context.output_format == "json":
-        output = {f"{stack.name}.yaml: {stack.external_name}" for stack in plan.all_stacks}
-    else:
-        output = [f"{stack.name}.yaml: {stack.external_name}" for stack in plan.all_stacks]
+    output = {
+        f"{stack.name}.yaml: {stack.external_name}"
+        for stack in plan.all_stacks
+        if all or stack in plan.command_stacks
+    }
 
-    write(output, context.output_format)
+    output_format = "json" if context.output_format == "json" else "yaml"
+    write(output, output_format)
--- a/sceptre/cli/list.py
+++ b/sceptre/cli/list.py
@@ -6,13 +6,13 @@ from sceptre.cli.helpers import (
     write
 )
 from sceptre.plan.plan import SceptrePlan
+from sceptre.helpers import normalise_path
 
 
 @click.group(name="list")
 def list_group():
     """
     Commands for listing attributes of stacks.
-
     """
     pass
 
@@ -64,7 +64,7 @@ def list_outputs(ctx, path, export):
     :type path: str
     :param export: Specify the export formatting.
     :type export: str
-   """
+    """
     context = SceptreContext(
         command_path=path,
         project_path=ctx.obj.get("project_path", None),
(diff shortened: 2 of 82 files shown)


sceptre/cli/template.py
```diff
@@ -137,3 +137,32 @@ def fetch_remote_template_command(ctx, path):
...
+    \f
+
+    :param path: The path to execute the command on.
+    :type path: str
+    """
+    context = SceptreContext(
+        command_path=path,
+        project_path=ctx.obj.get("project_path"),
+        user_variables=ctx.obj.get("user_variables"),
+        options=ctx.obj.get("options"),
```
jfalkenstein, line 158: Any reason why you've neglected the output_format arg?

sceptre/cli/template.py
```diff
@@ -137,3 +137,32 @@ def fetch_remote_template_command(ctx, path):
             output.append(template)
 
     write(output, context.output_format)
+
+
+@click.command(name="stack-name", short_help="Reveal the stack name or names.")
```
jfalkenstein, line 142: This command doesn't really seem like a _template_ command. It more appropriately seems a "list" command. It seems like the syntax would better be:

`sceptre list stack-names`

sceptre/cli/template.py
```diff
@@ -137,3 +137,32 @@ def fetch_remote_template_command(ctx, path):
...
+        options=ctx.obj.get("options"),
+        ignore_dependencies=ctx.obj.get("ignore_dependencies")
+    )
+
+    plan = SceptrePlan(context)
+    responses = plan.stack_name()
+
+    chop_output = len(responses) == 1
+    for response in responses.values():
+        output = response.split(" ")[0] if chop_output else response
```
jfalkenstein, line 167: What are the values going to look like here? I'm a bit confused about why you'd need to split the values by spaces.

sceptre/cli/template.py
```diff
@@ -137,3 +137,32 @@ def fetch_remote_template_command(ctx, path):
...
+        write(output, context.output_format)
```
jfalkenstein, line 168: It seems to me that this command should always output a list, not sequentially output individual values.

sceptre/config/reader.py
```diff
@@ -530,6 +530,7 @@ def _construct_stack(self, rel_path, stack_group_config=None):
         )
         stack = Stack(
             name=stack_name,
+            rel_path=rel_path,
```
jfalkenstein, line 533: What's the point of this change?

sceptre/plan/actions.py
```diff
@@ -41,6 +41,7 @@ class StackActions(object):
     def __init__(self, stack):
         self.stack = stack
         self.name = self.stack.name
+        self.rel_path = self.stack.rel_path
```
jfalkenstein, line 44: It's not good practice to abbreviate variable names. It only saves a few keystrokes and sacrifices clear understanding. A better variable name would be useful for communicating intent.

sceptre/plan/actions.py
```diff
@@ -1000,3 +1001,13 @@ def diff(self, stack_differ):
...
+
+    @add_stack_hooks
+    def stack_name(self):
+        """
+        Returns the Stack's stack name.
+
+        :returns: The Stack's stack name (external_name).
+        :rtype: str
+        """
+        return f"{self.stack.external_name} (from: {self.rel_path})"
```
jfalkenstein, line 1013: The "from" bit seems unnecessary. This appears to be a _front-end_, presentation concern that you've injected into the StackActions. Since the SceptrePlan ends up making a dictionary of Stacks with the action return value as the value anyway, I think that if you really wanted this output, that would be put together by the CLI function, not in the Actions.

sceptre/stack.py
```diff
@@ -120,10 +120,11 @@ class Stack(object):
     hooks = HookProperty("hooks")
 
     def __init__(
-        self, name, project_code, region, template_path=None, template_handler_config=None,
-        template_bucket_name=None, template_key_prefix=None, required_version=None,
-        parameters=None, sceptre_user_data=None, hooks=None, s3_details=None,
-        iam_role=None, dependencies=None, role_arn=None, protected=False, tags=None,
+        self, name, rel_path, project_code, region, template_path=None,
```
jfalkenstein, line 123: Same comment above, `rel_path` is not a good variable name. Let's not abbreviate like that.

Beyond that, though, I'm wondering _why_ this is necessary? What _is_ "rel_path"? What purpose does it serve? There's no documentation on that parameter here. It's hard for me to determine what is going on here.

jfalkenstein (changes requested):
My biggest feedback here would be this:
1. This should be a `list` sub-command, not a `template` command.
2. It should output a list of values, taking into account the output_format
3. The action seems to have presentation logic in it that it shouldn't really be concerned with
4. All the business of "rel_path" needs to be better explained, but I don't think you should do whatever you're trying to do this way.

sceptre/plan/actions.py
jfalkenstein: Honestly, I don't think this needs to be an Action at all.

sceptre/stack.py
```diff
@@ -136,6 +137,7 @@ def __init__(
             raise InvalidConfigFileError("Neither 'template_path' nor 'template' is set")
 
         self.name = sceptreise_path(name)
+        self.rel_path = rel_path
```
jfalkenstein, line 140: The name parameter _is_ the rel_path, just without the ".yaml" on it. I don't really think there's a good reason to accept the rel_path, when the `name` is already available. I think you really should just use the `name` for all the places.

sceptre/cli/template.py
jfalkenstein: Honestly, 90% of this PR could go away if you did something like this:

```python
stack_names = []
for stack in plan.command_stacks:
    relative_path = normalize_path(f'{stack.name}.yaml')
    stack_name = stack.external_name
    entry = f'{stack_name} (from {relative_path})'
    stack_names.append(entry)

write(stack_names, context.output_format)
```

zaro0508 (changes requested):
This needs docs.

sceptre/cli/template.py
alexharv074: Agreed, and done.

jfalkenstein:
Could you give some sample outputs here?

alexharv074:
Example output

Screen Shot 2021-12-30 at 4 33 13 pm

jfalkenstein:
image

This is odd to me. I'm assuming it's because of stack dependencies. It might not be clear why, when requesting the name for one stack, it creates three stack names. That would make sense for a Stack Group...

I think you might want to consider maybe hardcoding "ignore-dependencies" to be set to True for this, since I don't think stack names have any relationship to dependencies.

sceptre/cli/list.py
```diff
@@ -127,3 +126,30 @@ def list_change_sets(ctx, path, url):
...
+        command_path=path,
+        project_path=ctx.obj.get("project_path"),
+        user_variables=ctx.obj.get("user_variables"),
+        output_format=ctx.obj.get("output_format"),
+        options=ctx.obj.get("options"),
+        ignore_dependencies=ctx.obj.get("ignore_dependencies")
+    )
+
+    plan = SceptrePlan(context)
+    output = [f"{stack.name}.yaml: {stack.external_name}" for stack in plan.all_stacks]
```
jfalkenstein, line 154: I'm wondering why not just make it a dictionary? It would be easier to parse that way and read even better.
```suggestion
    output = {f"{stack.name}.yaml: {stack.external_name}" for stack in plan.all_stacks}
```

jfalkenstein: Also, I think you should have a cli flag to activate the "all_stacks" capability. When you use "all_stacks", it completely disregards the command path. 

I'm thinking the way to get all stacks would be like `sceptre list stacks --all`. That would require making the command path optional, but you could just blow up if `--all` is not specified and there's no command path.

sceptre/plan/plan.py
```diff
@@ -34,9 +34,8 @@ def __init__(self, context):
         self.launch_order = None
 
         config_reader = ConfigReader(context)
-        all_stacks, command_stacks = config_reader.construct_stacks()
-        self.graph = StackGraph(all_stacks)
```
jfalkenstein, line 38: I think you can actually iterate on `self.graph` to iterate all stacks, just in order of their dependencies. You might not actually need `self.all_stacks`. You should verify this, but I _think_ you could get away with not actually making any changes here.

tests/test_cli.py
```diff
@@ -625,6 +625,18 @@ def test_list_outputs_with_export(self):
         assert result.exit_code == 0
         assert result.output == "export SCEPTRE_Key='Value'\n"
 
+    @pytest.mark.parametrize("output_format,expected_output", [
+        ("yaml", '---\n- mock-stack.yaml: mock-stack-external\n\n'),
+        ("text", 'mock-stack.yaml: mock-stack-external\n'),
+        ("json", '[\n    {\n        "mock-stack.yaml": "mock-stack-external"\n    }\n]\n')
```
jfalkenstein, line 631: Yeah, it would look a lot better if you just made a single dict rather than a list of single-key dicts.

sceptre/plan/plan.py
alexharv074: @jfalkenstein I can't see how, what would be the benefit?

jfalkenstein: I don't understand your statement. My point is that I think you can iterate through the `graph` attribute instead of changing this code to expose `all_stacks`. In other words, I _think_ this change was unnecessary.

sceptre/cli/list.py
```diff
@@ -127,3 +126,35 @@ def list_change_sets(ctx, path, url):
...
+        options=ctx.obj.get("options"),
+        ignore_dependencies=ctx.obj.get("ignore_dependencies")
+    )
+
+    plan = SceptrePlan(context)
+
+    if context.output_format == "json":
+        output = {f"{stack.name}.yaml: {stack.external_name}" for stack in plan.all_stacks}
+    else:
+        output = [f"{stack.name}.yaml: {stack.external_name}" for stack in plan.all_stacks]
```
jfalkenstein, line 158: I'm not sure why you did this. Why not just always make it a dictionary comprehension? 

This is all I wanted: 
```suggestion
    output = {f"{stack.name}.yaml: {stack.external_name}" for stack in plan.all_stacks
```

alexharv074:
@jfalkenstein This implementation is what I think you proposed, and it doesn't seem to work the way you hoped:

```python
@list_group.command(name="stacks")
@click.option(
    "--all", is_flag=True, help="List all stacks in stack groups."
)
@click.argument("path")
@click.pass_context
@catch_exceptions
def list_stacks(ctx, path, all):
    """
    List stack names and command paths for stacks in a
    stack group.
    \f

    :param path: Path to execute the command on.
    :type path: str
    :param all: List all stacks in stack groups.
    :type all: bool
    """
    context = SceptreContext(
        command_path=path,
        project_path=ctx.obj.get("project_path"),
        user_variables=ctx.obj.get("user_variables"),
        output_format=ctx.obj.get("output_format"),
        options=ctx.obj.get("options"),
        ignore_dependencies=ctx.obj.get("ignore_dependencies"),
        full_scan=all
    )
    plan = SceptrePlan(context)

    output = {
        f"{stack.name}.yaml: {stack.external_name}"
        for stack in plan.command_stacks
    }

    write(output, output_format)
```

alexharv074:
New output

Screen Shot 2022-01-13 at 4 36 04 pm

sceptre/plan/plan.py
alexharv074: @jfalkenstein Yes and from reading the code it's not obvious to me how I can iterate through `StackGraph` without adding a method to that class for the purpose of iterating.

sceptre/cli/helpers.py
```diff
@@ -113,27 +113,34 @@ def _generate_json(stream):
 
 
 def _generate_yaml(stream):
-    if isinstance(stream, list):
+    kwargs = {
+        "default_flow_style": False,
+        "explicit_start": True
+    }
+
+    if type(stream) in [list, set]:
```
jfalkenstein, line 121: The proper way to check for types is with `isinstance()`, as that will handle things like inheritance properly.

```suggestion
    if isinstance(stream, (list, set)):
```

sceptre/cli/list.py
```diff
@@ -127,3 +126,42 @@ def list_change_sets(ctx, path, url):
...
+        options=ctx.obj.get("options"),
+        ignore_dependencies=ctx.obj.get("ignore_dependencies"),
+        full_scan=True
+    )
+    plan = SceptrePlan(context)
+
+    output = {
+        f"{stack.name}.yaml": stack.external_name
+        for stack in plan.all_stacks
+        if all or stack in plan.command_stacks
```
jfalkenstein, line 163: That'll work

sceptre/cli/list.py
```diff
@@ -127,3 +126,42 @@ def list_change_sets(ctx, path, url):
...
+    :type all: bool
+    """
+    context = SceptreContext(
+        command_path=path,
+        project_path=ctx.obj.get("project_path"),
+        user_variables=ctx.obj.get("user_variables"),
+        output_format=ctx.obj.get("output_format"),
+        options=ctx.obj.get("options"),
+        ignore_dependencies=ctx.obj.get("ignore_dependencies"),
+        full_scan=True
```
jfalkenstein, line 156: I thought you said `full_scan` wasn't doing anything for you?

sceptre/cli/list.py
```diff
@@ -127,3 +126,42 @@ def list_change_sets(ctx, path, url):
...
+def list_stacks(ctx, path, all):
+    """
+    List stack names and command paths for stacks in a
+    stack group.
+    \f
+
+    :param path: Path to execute the command on.
+    :type path: str
+    :param all: List all stacks in stack groups.
+    :type all: bool
```
jfalkenstein, line 147: Remember, for click commands, we shouldn't have type docstrings, since the docs are more appropriately rendered by Click.

jfalkenstein (approved):

sceptre/cli/list.py
```diff
@@ -127,3 +126,32 @@ def list_change_sets(ctx, path, url):
...
+
+
+@list_group.command(name="stacks")
+@click.argument("path")
+@click.pass_context
+@catch_exceptions
+def list_stacks(ctx, path):
+    """
+    List stack names and command paths for stacks in a
+    stack group.
```
zaro0508, line 138: The docs for this command is a bit unclear because stack names can be retrieved from local or remote.  It can also be derived by calculating a union of local and remote stacks.  This leads me to wonder about the purpose and intention of this command.  It looks like it currently only shows a list of locally defined stacks in a stack group but is that the only thing it should do?  if so then could you please clarify that in the docs?  

On a side note, it would be nice to exclude from the list the local stacks that have not been deployed to AWS.  Although this can be done in a separate PR.

sceptre/cli/list.py
zaro0508: I don't think you want to set full_scan to True here otherwise sceptre will always generate a list for all stack groups no matter what path you pass

sceptre/cli/list.py
```diff
@@ -127,3 +126,32 @@ def list_change_sets(ctx, path, url):
...
+    \f
+
+    :param path: Path to execute the command on.
```
zaro0508, line 141: Path to execute the command on/Path to stack group

tests/test_cli.py
```diff
@@ -625,6 +625,18 @@ def test_list_outputs_with_export(self):
...
         assert result.output == "export SCEPTRE_Key='Value'\n"
 
+    @pytest.mark.parametrize("output_format,expected_output", [
+        ("yaml", '---\nmock-stack.yaml: mock-stack-external\n\n'),
+        ("text", '---\nmock-stack.yaml: mock-stack-external\n\n'),
+        ("json", '{\n    "mock-stack.yaml": "mock-stack-external"\n}\n')
+    ])
+    def test_list_stacks(self, output_format, expected_output):
+        result = self.runner.invoke(
+            cli, ["--output", output_format, "list", "stacks", "dev/vpc.yaml"]
```
zaro0508, line 635: don't you just want to pass in `dev` instead of `dev/vpc.yaml` because `dev` is the stack group?

zaro0508 (changes requested):

sceptre/cli/list.py
alexharv074: Ah, ok... it seemed to work fine but I'll change it back.

zaro0508: you didn't change it back?

zaro0508 (changes requested):

jfalkenstein:
@alexharv074 , you have some conflicts that need to be resolved, then I'll run integration tests for you.

sceptre/cli/list.py
alexharv074: @zaro0508  I  did now, I must have somehow missed this.

sceptre/cli/list.py
```diff
@@ -127,3 +126,39 @@ def list_change_sets(ctx, path, url):
 
     for response in responses:
         write(response, context.output_format)
+
+
+@list_group.group(name="local")
```
zaro0508, line 131: A `local` command doesn't really make sense to me and seems unnecessary.  I think i prefer just `sceptre list stacks prod/mystack.yaml`.   if we decide to support getting attributes from remote templates in the future then i think we can just add a `--remote` flag?

sceptre/cli/list.py
```diff
@@ -127,3 +126,39 @@ def list_change_sets(ctx, path, url):
...
+
+@local_group.command(name="stacks")
+@click.argument("path")
+@click.pass_context
+@catch_exceptions
+def list_stacks(ctx, path):
+    """
+    List stack names and command paths for local stacks in a
+    stack group.
+    \f
```
zaro0508, line 147: Help is cutting this off. please try to shorten the description slightly..

```
➜  sceptre list local --help
Usage: sceptre list local [OPTIONS] COMMAND [ARGS]...

  Commands for listing attributes of local stacks.

Options:
  --help  Show this message and exit.

Commands:
  stacks  List stack names and command paths for local stacks in a stack...
```

How about just `List sceptre stack config attributes`?

zaro0508 (changes requested):

sceptre/cli/list.py
alexharv074: Reverted.

zaro0508 (approved):

PR #1172, opened 2021-11-28
update setuptools to fix doc deployments

zaro0508:
Doing as instructed here More info at https://github.com/pypa/setuptools_scm/issues/605


zaro0508:
fix with PR #1171

Issue #1180, opened 2021-12-05
Change set display doesn't follow --output flag


dboitnot:
### Change set display doesn't follow --output flag
Despite [being referenced in update.py](https://github.com/Sceptre/sceptre/blob/251d2b6a855b25dd725fc6658b86d57eb83d5edc/sceptre/cli/update.py#L48), it appears that the ChangeSet display does not use the setting from the `--output` flag.

### Your environment
* Sceptre, version 2.7.0
* Python 3.8.2
* Environment: python:3.8.2-slim docker image

### Steps to reproduce
```
sceptre --output=yaml update -c ban/banner-app-image-build.yaml
[2021-12-05 16:45:57] - ban/banner-app-image-build - Successfully initiated creation of Change Set 'change-set-d167800455ea11ec86ff0242ac110002'
{'ChangeSetName': 'change-set-d167800455ea11ec86ff0242ac110002', 'StackName': 's-ban-banner-app-image-build', 'CreationTime': datetime.datetime(2021, 12, 5, 16, 45, 57, 184000, tzinfo=tzlocal()), 'ExecutionStatus': 'AVAILABLE', 'Status': 'CREATE_COMPLETE', 'Changes': [{'ResourceChange': {'Action': 'Modify', 'LogicalResourceId': 'CodeBuildProject', 'PhysicalResourceId': 'banner-app', 'ResourceType': 'AWS::CodeBuild::Project', 'Replacement': 'Conditional', 'Scope': ['Properties']}}]}
```

### Expected behaviour
ChangeSet should be displayed in YAML format.

### Actual behaviour
ChangeSet is displayed in JSON format.

### Notes

The `output` flag is clearly being parsed, because if a malformed value is provided we get an error:

```
$ sceptre --output=yamll update -c ban/banner-app-image-build.yaml
Usage: sceptre [OPTIONS] COMMAND [ARGS]...
Try 'sceptre --help' for help.

Error: Invalid value for '--output': 'yamll' is not one of 'text', 'yaml', 'json'.
```

Further, since this is an interactive process, we don't expect this output to be machine-parsed. I'd therefore think that YAML, being vastly more readable, would be a better default.

Possibly, I'm misunderstanding the purpose of the `--output` flag, because if I force the value to `yaml` in `update.py` I still get JSON output:

```python
            # Describe changes
            descriptions = plan.describe_change_set(change_set_name)
            for description in list(descriptions.values()):
                if not verbose:
[...]

Issue #1169, opened 2021-11-25
Feature request: detect-stack-drift command


alexharv074:
## Feature request

Please implement a feature `detect-stack-drift` that wraps the Boto3 [`detect_stack_drift`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation.html#CloudFormation.Client.detect_stack_drift), [`describe_stack_drift_detection_status`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation.html#CloudFormation.Client.describe_stack_drift_detection_status) and [`describe_stack_resource_drifts`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation.html#CloudFormation.Client.describe_stack_resource_drifts) to provide info about stack drifts in a running Sceptre stack.

PR #1183, opened 2021-12-09
More readable ChangeSet review output

dboitnot:
As proposed in #1180, this PR outputs ChangeSet descriptions as YAML to be more readable.

## PR Checklist

## Other Information


sceptre/cli/update.py
```diff
@@ -67,8 +68,11 @@ def update_command(ctx, path, change_set, verbose, yes):
             for description in list(descriptions.values()):
                 if not verbose:
                     description = simplify_change_set_description(description)
-                write(description, context.output_format)
 
+                write('-'*80,'text')
+                write(yaml.dump(description), 'text')
```
jfalkenstein, line 73: Why is this hard-coded to yaml format instead of using the output_format that has been explicitly requested by the user? Sceptre supports 3 output format. What am I missing here?

dboitnot: Per #1180, the `write()` function does not appear to respect the `output_format`. I posted this patch to the issue and @zaro0508 requested a PR.

Note that this PR will need a new commit (5b4c2c3) which adds the missing yaml library dependency.

dboitnot: As an aside, given that this is an explicitly interactive process, I'd argue that the more readable approach should always be used since we don't anticipate the output being piped to another process.

jfalkenstein: @alexharv074 Is this, by any chance, the same issue you pointed out regarding the `write()` function? Or is this different?

jfalkenstein: @dboitnot, rather than doing this to bypass the output format on `write`, I'd rather you _fix_ the `write` function. That seems like the more appropriate fix. I don't think this is the first time the weaknesses with `write` have come up recently. But by providing a global option for output format, Sceptre makes "contract" with the user. I don't want to override that to force yaml (despite the superiority of yaml over json, IMO).

dboitnot: @jfalkenstein, as mentioned above, this PR was created by request from a workaround I posted to #1180. The original issue actually reports this as an error with `write()`.

I agree 100% with keeping our promise to the user about output formatting. I do wonder if that applies in this case because it's an explicitly interactive process. But then it's possible I don't fully grasp the purpose of the `--output` flag or the `write()` function.

jfalkenstein: What do you mean by "explicitly interactive"? I'm confused about that bit.

zaro0508: Fixing the output format and what the default output format should be are two separate issues.  Let's focus on fixing the output format first.   I agree with @jfalkenstein we should either fix the write function or replace it with a library that can do for us.

dboitnot: @jfalkenstein, Sorry, by "explicitly interactive" I mean that that the `update -c` command presents user-readable output and pauses for user interaction (pressing 'y' or 'n'). To me this suggests we're assuming a user is interacting with the command at a terminal and that we're not running in a pipe or script which might parse the output. It's for that reason I wonder if the `--output` contract applies in this case.

@zaro0508, This PR is just the workaround I posted to #1180 and I have no objection to discarding it in favor of addressing the issue with `write()`.

jfalkenstein: @dboitnot , sorry I lost track of this convo with the holidays and such. #1169 is soon to be merged, which fixes the issue with dumping yaml dicts. I suspect that once that is merged, this bug likely will go away. Not sure if this PR will even be necessary. 

With that said, there has been ongoing conversation around overhauling the `write()` function, as it has lots of inconsistencies and other issues with it and how it operates (especially around the `text` output type). There will likely be an issue around that soon.

jfalkenstein:
Hey @dboitnot, It's been quite a while since this was posted. Do you happen to know if the issue this PR is attempting to fix has already been resolved in later releases (as I suggested it might be in my comments above)?

dboitnot:
Agreed. The current version of Sceptre addresses this issue.

PR #1184, opened 2021-12-11
[Resolve #1114,#426] Project Dependencies part 1: ResolvableProperty refactor + tags resolvable property

jfalkenstein:
This is the first in a series of pull requests that addresses #1114 , allowing Sceptre to manage its own dependencies.

## In this PR:
* **ResolvableProperty is refactored into a base class with a single subclass (for now): ResolvableContainerProperty.** _This refactor will be necessary so we can build on that base class to make a ResolvableValueProperty in a later PR so we can resolve single values like template_bucket_name, etc..._. 
* **The merge strategies list_join and dict_merge now perform deepcopy on the product of those joins,** preventing changes in child stack configs from modifying the same dict/list in memory that has been inherited by other child stack configs. This change is necessary so inheriting mutable values from parent configs becomes a lot safer.
* **A few bugs have been fixed in ResolvableProperty** in this refactor in preparation for changes that are to come in future Pull requests. These all have to do with being able to inherit resolvers and having more resolvers on stack configurations. 
    - Previously, the `_no_recursive_get` method was effectively blocking access to the same property on a DIFFERENT stack. If a resolver accesses the same property on another stack, there shouldn't be a problem with this. This is relevant in cases like the StackOutput resolver that accesses _another_ stack's attributes. This bug has been fixed now. `_no_recursive_get` specifically disallows recursive gets on the _current stack only_.
    - If a resolver was set on a StackGroup config and then "inherited" by multiple stack groups, the _exact same_ object in memory would now be on two stacks. Worse, they would both have the same Stack object assigned to them... which means that Stack A could use a resolver that was referencing Stack B. This has now been solved by resolvers being cloned when they are assigned to a stack, guaranteeing that every stack has an independent resolver in memory.
    - Deferred resolution is properly handled _within_ the property. Previously it wasn't, so we needed a weird double-layer property on the Stack class to handle sceptre_user_data. This is now no longer necessary and is handled where it should have been handled all along: in the property itself.
* **stack_tags is now a ResolvableProperty**, ultimately finishing what #847 attempted but never finished.
* Documentation on which Stack and StackGroup configuration keys are resolvable and which are not has been added. 

## PR Checklist

## Other Information

### DIFF
--- a/docs/_source/docs/stack_config.rst
+++ b/docs/_source/docs/stack_config.rst
@@ -1,8 +1,9 @@
 Stack Config
 ============
 
-Stack config stores config related to a particular Stack, such as the path to
-that Stack’s Template, and any parameters that Stack may require.
+A Stack config stores configurations related to a particular Stack, such as the path to
+that Stack’s Template, and any parameters that Stack may require. Many of these configuration keys
+support resolvers and can be inherited from parent StackGroup configs.
 
 .. _stack_config-structure:
 
@@ -31,6 +32,8 @@ you will receive an error when deploying the stack.
 
 template_path
 ~~~~~~~~~~~~~~~~~~~~~~~~
+* Resolvable: No
+* Can be inherited from StackGroup: No
 
 The path to the CloudFormation, Jinja2 or Python template to build the Stack
 from. The path can either be absolute or relative to the Sceptre Directory.
@@ -44,6 +47,8 @@ from the Stack config filename.
 
 template
 ~~~~~~~~
+* Resolvable: No
+* Can be inherited from StackGroup: No
 
 Configuration for a template handler. Template handlers can take in parameters
 and resolve that to a CloudFormation template. This enables you to not only
@@ -66,6 +71,9 @@ developing your own in the :doc:`template_handlers` section.
 
 dependencies
 ~~~~~~~~~~~~
+* Resolvable: No
+* Can be inherited from StackGroup: Yes
+* Inheritance strategy: Appended to parent's dependencies
 
 A list of other Stacks in the environment that this Stack depends on. Note that
 if a Stack fetches an output value from another Stack using the
@@ -74,12 +82,18 @@ and that Stack need not be added as an explicit dependency.
 
 hooks
 ~~~~~
+* Resolvable: No
+* Can be inherited from StackGroup: Yes
+* Inheritance strategy: Overrides parent if set
 
 A list of arbitrary shell or Python commands or scripts to run. Find out more
 in the :doc:`hooks` section.
 
 notifications
 ~~~~~~~~~~~~~
+* Resolvable: Yes
+* Can be inherited from StackGroup: Yes
+* Inheritance strategy: Overrides parent if set
 
 List of SNS topic ARNs to publish Stack related events to. A maximum of 5 ARNs
 can be specified per Stack. This configuration will be used by the ``create``,
@@ -89,6 +103,9 @@ documentation`_.
 
 on_failure
 ~~~~~~~~~~
+* Resolvable: No
+* Can be inherited from StackGroup: Yes
+* Inheritance strategy: Overrides parent if set
 
 This parameter describes the action taken by CloudFormation when a Stack fails
 to create. For more information and valid values see the `AWS Documentation`_.
@@ -104,6 +121,9 @@ Examples include:
 
 parameters
 ~~~~~~~~~~
+* Resolvable: Yes
+* Can be inherited from StackGroup: Yes
+* Inheritance strategy: Overrides parent if set
 
 .. warning::
 
@@ -166,6 +186,9 @@ Example:
 
 protected
 ~~~~~~~~~
+* Resolvable: No
+* Can be inherited from StackGroup: Yes
+* Inheritance strategy: Overrides parent if set
 
 Stack protection against execution of the following commands:
 
@@ -180,12 +203,18 @@ throw an error.
 
 role_arn
 ~~~~~~~~
+* Resolvable: No
+* Can be inherited from StackGroup: Yes
+* Inheritance strategy: Overrides parent if set
 
 The ARN of a `CloudFormation Service Role`_ that is assumed by CloudFormation
 to create, update or delete resources.
 
 iam_role
 ~~~~~~~~
+* Resolvable: No
+* Can be inherited from StackGroup: Yes
+* Inheritance strategy: Overrides parent if set
 
 This is the IAM Role ARN that **Sceptre** should *assume* using AWS STS when executing any actions
 on the Stack.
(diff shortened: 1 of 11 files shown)


integration-tests/steps/helpers.py
```diff
@@ -73,7 +73,7 @@ def get_cloudformation_stack_name(context, stack_name):
 
 
 def retry_boto_call(func, *args, **kwargs):
-    delay = 2
+    delay = 5
```
jfalkenstein, line 76: This change should help resolve some of the intermittent throttling errors we get in our integration tests.

sceptre/config/strategies.py
```diff
@@ -21,16 +22,17 @@ def list_join(a, b):
     """
     if a and not isinstance(a, list):
         raise TypeError('{} is not a list'.format(a))
+
     if b and not isinstance(b, list):
         raise TypeError('{} is not a list'.format(b))
 
     if a is None:
-        return b
+        return deepcopy(b)
```
jfalkenstein, line 30: This change helps prevent mutable values that are shared across a stack group (like the dependencies list) from being the _same_ object in memory across stacks. This prevents a weird bug that can happen when one stack's value gets modified by another stack.

sceptre/resolvers/__init__.py
```diff
@@ -49,62 +49,231 @@ def resolve(self):
         """
         pass  # pragma: no cover
 
+    def clone(self, stack: 'stack.Stack') -> 'Resolver':
+        """
+        Produces a "fresh" copy of the Resolver, with the specified stack.
+
+        :param stack: The stack to set on the cloned resolver
+        """
+        return type(self)(self.argument, stack)
```
jfalkenstein, line 58: Cloning resolvers _before_ they're assigned to a stack is critical to preventing a single resolver instance being shared across multiple stacks (inherited from a stack group). The reason why that is so bad is because if the single instance is shared across Stack A and Stack B, only one of those stacks will have the right stack assigned to the resolver.

sceptre/resolvers/__init__.py
```diff
@@ -49,62 +49,231 @@ def resolve(self):
...
+
 
-class ResolvableProperty(object):
+class ResolvableProperty(abc.ABC):
```
jfalkenstein, line 61: I've made ResolvableProperty into a base class with (currently) one implementation, ResolvableContainerProperty. This is necessary so in my next PR I can make the ResolvableValueProperty.

sceptre/resolvers/__init__.py
```diff
@@ -49,62 +49,231 @@ def resolve(self):
...
+        # access them. Thus, we set this "get_in_progress" attribute on the stack instance rather
+        # than the descriptor instance.
+        get_status_name = f'_{self.name}_get_in_progress'
+        if getattr(stack, get_status_name, False):
+            raise RecursiveResolve(f"Resolving Stack.{self.name[1:]} required resolving itself")
+        setattr(stack, get_status_name, True)
+        try:
+            yield
+        finally:
+            setattr(stack, get_status_name, False)
```
jfalkenstein, line 114: Previously, this lock would prevent you from accessing the same property on OTHER stacks. This was a bug I had to fix so that Stack A's resolver can access that same property on Stack B. This is what will happen with the StackOutput resolver when it tries to access the `iam_role` of the target stack.

sceptre/resolvers/__init__.py
```diff
@@ -49,62 +49,231 @@ def resolve(self):
...
+                container,
+                self.ResolveLater
+            )
+            # Search the container to see if there are any ResolveLaters left;
+            # Raise a RecursiveResolve if there are.
+            _call_func_on_values(
+                raise_if_not_resolved,
+                container,
+                self.ResolveLater
+            )
```
jfalkenstein, line 272: This functionality replaces the double-resolution we _were_ doing for Sceptre user data on stacks. Honestly, proper resolution of these circumstances belonged in the resolver all along and not as a double-property on the Stack.

sceptre/stack.py
```diff
@@ -263,17 +265,6 @@ def connection_manager(self):
...
-    def sceptre_user_data(self):
-        """Returns sceptre_user_data after ensuring that it is fully resolved.
-
-        :rtype: dict or list or None
-        """
-        if not self._sceptre_user_data_is_resolved:
-            self._sceptre_user_data_is_resolved = True
-            self._resolve_sceptre_user_data()
-        return self._sceptre_user_data
-
```
jfalkenstein, line 276: We don't need this any more, since Resolvers handle this situation internally now.

jfalkenstein:
Alright, @zaro0508 and @zxiiro Here's the first of my Re-tailored PRs. This is mostly a refactor of the existing resolver system (adding the ability to make tags resolvable). There shouldn't need to be additional integration tests in this PR since existing integration tests should cover the refactor quite well.

**This pull request is _complete_ and can be merged to master and deployed on its own without any subsequent PRs.**

@zaro0508, 90% of the content of this PR you have already reviewed and approved. However, I closed the feature branch PR and pointed this one at master as an independently deployable unit. I figured you'd want to re-review in light of our recent conversation.

zaro0508 (approved):

jfalkenstein:
@zaro0508 You think this thing is ready to merge to master?

zaro0508:
One request and i think we are good to merge.

Setting any dependency in Sceptre's root StackGroup config (config/config.yaml) causes a dependency cycle.  Can we add a warning to the docs explicitly telling users that they cannot define dependencies in their root StackGroup config otherwise it will cause a dependency cycle and fail?

jfalkenstein:
@zaro0508 , I'm cool with that. I can add a similar warning on the `dependencies` attribute docs as well as the `stack_output` resolver docs.

jfalkenstein:
@zaro0508, I added two warnings about circular dependencies, as you requested. Take a look at https://github.com/Sceptre/sceptre/pull/1184/commits/824c2336c88048afd3a0c17fcf9d697cb65e7b0e for the specific warnings commit.

docs/_source/docs/stack_config.rst
```diff
@@ -80,6 +80,13 @@ if a Stack fetches an output value from another Stack using the
 ``stack_output`` resolver, that Stack is automatically added as a dependency,
 and that Stack need not be added as an explicit dependency.
 
+.. warning::
+   Be careful about how you structure dependencies. It is possible to create circular
+   dependencies accidentally, where multiple stacks depend on each other. Sceptre
+   will detect this and raise an error, blocking this sort of setup. This will *always*
+   be the case when you set dependencies on a StackGroup Config referencing stacks
+   inside that StackGroup.
```
zaro0508, line 88: I think this description is a bit confusing and misleading because I don't think that applies to every case.

Given the stest.zip sample.  if i set any dependency in the root StackConfig (config/config.yaml) it will cause a cycle error.  if I set a dependency in a config one level down (config/nba/config.yaml) it works fine.   I believe both configs __NOT__  `referencing stacks inside that StackGroup`

jfalkenstein: I'll try clarifying this. I think I see where your confusion comes from.

jfalkenstein:
@zaro0508 , I've clarified the two circular dependency warnings. Hopefully that is more understandible.

zaro0508 (approved):

jfalkenstein:
@zaro0508 this PR has consistently expected the build and unit test check, but it hasn't ever run. Do you have any idea why it hasn't run?

PR #1185, opened 2021-12-11
[Resolve #1114,#1000] Project Dependencies part 4: Handling resolvers referencing non-deployed stacks in template commands

jfalkenstein:

## In this PR:
* It is possible to enter a global state (using a context manager) that allows Resolvers that raise errors to resolve to a placeholder value. This is explicitly opt-in and will only be enabled on commands that don't actually deploy stacks, such as "generate", "validate", and "diff". This is required so that these commands don't fail when there are resolvers that reference stacks that aren't deployed yet. This ultimately is the bit that resolves issue #1000. This is important because this series of Pull requests more than doubles the number of stack attributes that are resolvable, which are going to be all the more problematic when using `sceptre validate`, etc...

## PR Checklist

## Other Information


sceptre/diffing/stack_differ.py
```diff
@@ -9,9 +9,7 @@
 from cfn_tools import ODict
```
jfalkenstein, line 9: All of the code removed in this file is because the StackDiffer implemented a more primitive form of the placeholder system introduced in this PR. Now that placeholders can be enabled globally, there's no longer a need to explicitly handle those situations in the StackDiffer.

integration-tests/steps/stack_groups.py
```diff
@@ -279,13 +303,19 @@ def get_stack_creation_times(context, stacks):
...
-    path = os.path.join(context.sceptre_dir, "config", stack_group_name)
+    config_dir = Path(context.sceptre_dir) / 'config'
+    path = config_dir / stack_group_name
+
     stack_names = []
-    for root, dirs, files in os.walk(path):
-        for filepath in files:
-            filename = os.path.splitext(filepath)[0]
-            if not filename == "config":
-                stack_names.append(os.path.join(stack_group_name, filename))
```
jfalkenstein, line 288: This didn't account for nested directories.

jfalkenstein:
@zaro0508 , @zxiiro This PR is now ready for full review and it wraps up the overhaul of resolvers we discussed.

sceptre/resolvers/no_value.py
```diff
@@ -0,0 +1,14 @@
...
+    """This resolver resolves to nothing, functioning just like the AWS::NoValue special value. When
+    assigned to a resolvable Stack property, it will remove the config key/value from the stack or
+    the container on the stack where it has been assigned, as if this value wasn't assigned at all.
+
+    This is mostly useful for simplifying conditional logic on Stack and StackGroup config files
+    where, if a certain condition is met, a value is passed, otherwise it's not passed at all.
+    """
+
+    def resolve(self) -> None:
+        return None
```
zaro0508, line 14: Would you get the same affect when using this resolver as passing in `None`?  like this..
```
sceptre_user_data:
  MyParam: "None"
```

will the above also remove `MyParam` key? or does that only happen when you set `MyParam: !no_value`?

docs/_source/docs/resolvers.rst
```diff
@@ -52,6 +52,14 @@ file_contents
 
 **deprecated**: Consider using the `file`_ resolver instead.
 
+no_value
+~~~~~~~~
+
+This resolver "resolves to nothing", functioning just as if it was not set at all. This works just
+like the "AWS::NoValue" special variable that you can reference on a CloudFormation template. It
+can help simplify Stack and StackGroup config Jinja logic in cases where, if a condition is met, a
+value is passed, otherwise no value is passed.
```
zaro0508, line 61: I'm having a difficult time understand the general use case for this resolver.   Is it something that's only used for debugging purposes?  I just can't figure out when I would ever set this in my sceptre files.  Can you please help me out?

sceptre/resolvers/no_value.py
jfalkenstein: No, you technically wouldn't. `None` in yaml resolves to the string `"None"` in Python. It's effectively the same, though, as `~` and `null`.

image.

jfalkenstein: >will the above also remove MyParam key? or does that only happen when you set MyParam: !no_value?

No, `"None"` will not result in `MyParam` from being removed. 

```
sceptre_user_data:
    MyParam: "None"

# This becomes...
{
    'sceptre_user_data': {
        'MyParam': 'None'
    }
}

sceptre_user_data:
    MyParam: null

# This becomes... 
{
    'sceptre_user_data': {
        'MyParam': None
    }
}

sceptre_user_data:
    MyParam: !no_value

# This becomes...
{
    'sceptre_user_data': {}
}
```

docs/_source/docs/resolvers.rst
jfalkenstein: So there's two things going on here:
1. Resolvers that resolve to `None` are removed from their container. 
2. `!no_value` is a resolver that exclusively resolves to `None` without any other functionality. 

It works just like `AWS::NoValue`, which is pretty useful in CloudFormation templates where you have logic like:
`!If [ 'MyCondition', !Ref MyVPCId, !Ref AWS::NoValue ]`. In a CloudFormation template, that condition would result in no value being set on that key if 'MyCondition' was equal to False.

In the context of Sceptre, imagine this scenario:

```yaml
parameters:
    my_parameter: {{ var.get('some_value_that_might_not_be_set', '!no_value' }}
```

Imagine the CloudFormation template has a default value for my_parameter. But if you pass any string at all for that parameter, that default value will not actually be used and it will just use the parameter value you've passed. But by using `!no_value` in this way, we won't actually pass "my_parameter" at all to the template if "some_value_that_might_not_be_set" is not set on the `var`.

Could you accomplish the same via other means in Jinja? Sure. Would it be more elegant? It depends on the way the user is setting up their Stack and StackGroups. `!no_value` gives the user an option that they can use. It gets more important when you talk about inheriting things from StackGroups.

zaro0508: BINGO! your example made it crystal clear to me, thanks.  Can we incorporate an example like that into the docs?  

also, would this be an equivalent example?
```
parameters:
    my_parameter: {{ environment_variable.MY_VAR  |  no_value }}  
```

jfalkenstein: > Can we incorporate an example like that into the docs?

Yeah, not a problem. 

> would this be an equivalent example?

Almost, but not quite. Remember, it's a resolver, so it would need to be:
```yaml
parameters:
    my_parameter: {{ environment_variable.MY_VAR  |  "!no_value" }} 
```

jfalkenstein:
@zaro0508, I have made your requested changes and added an example of using `!no_value`.

zaro0508 (approved):

PR #1189, opened 2022-01-01
[Resolve #1114] Resolvable Template Handler configs and the !stack_attr resolver

jfalkenstein:
This pull request adds two related things:
1. Template handler configurations are now fully resolvable.
2. You can now use the `!stack_attr` resolver to reference other configurations on the stack and StackGroup 

The need for these two things came about in my testing on the recent changes to Sceptre's resolvers. 

## Why is this necessary?
Previously, I had this on one of my stacks:

```yaml
template:
    type: sam
    path: beatbox/template.yaml
    artifact_bucket_name: {{ template_bucket_name }}
    artifact_prefix: {{ template_key_prefix }}
```

^^ This has been working just fine until I set `template_bucket_name` in my project to a `!stack_output` resolver. Once I did that, however, the template handler blew up because Jinja was attempting to cast the StackOutput resolver to string.

With the changes in this PR, I can now do this:

```yaml
template:
    type: sam
    path: beatbox/template.yaml
    # This will access and resolve the value of template_bucket_name from the StackGroup 
    # and then access that value as the resolved value for handler.
    artifact_bucket_name: !stack_attr template_bucket_name
    artifact_prefix: {{ template_key_prefix }}
```

## PR Checklist

## Other Information


docs/_source/docs/resolvers.rst
```diff
@@ -79,6 +79,37 @@ A resolver to execute any shell command.
...
+
+.. code-block:: yaml
+
+   sceptre_user_data:
+       key:
+           - "some random value"
+           - "the value we want to select"
+
+   parameters:
+       my_parameter: !stack_attr sceptre_user_data.key.1
```
zaro0508, line 111: does ordering matter?  can you put paramters above sceptre_user_data?

Would it be possible to use more concrete examples for `key` and the values for it?   Also could we add a comment for `my_parameter` telling users what  `!stack_attr sceptre_user_data.key.1` will pass to my_parameter?

docs/_source/docs/resolvers.rst
```diff
@@ -79,6 +79,37 @@ A resolver to execute any shell command.
...
+For example, if you need to reference the ``template_bucket_name`` set in the StackGroup Config,
+you can do that with ``!stack_attr``:
+
+.. code-block:: yaml
+
+   template:
+       type: sam
+       path: path/from/my/cwd/template.yaml
+       # template_bucket_name could be set by a resolver in the StackGroup.
+       artifact_bucket_name: !stack_attr template_bucket_name
```
zaro0508, line 98: I think users would expect that  `artifact_bucket_name: {{ stack_group_config.template_bucket_name }}` would work as well.  if it doesn't then can you please provide some explanation as to why it would not work?

zaro0508 (changes requested):

docs/_source/docs/resolvers.rst
zaro0508: does the template handler doc need updating? https://github.com/Sceptre/sceptre-sam-handler#arguments section.. 
```
artifact_bucket_name (string, required): The bucket name where artifacts should be uploaded to on S3 during the packaging process. If your project has a template_bucket_name, you can set this to {{ template_bucket_name }}.
```

jfalkenstein: > I think users would expect that artifact_bucket_name: {{ stack_group_config.template_bucket_name }} would work as well

The problem here is that template_bucket_name (and many other items that might be in stack_group_config) might be resolvers. Especially with the recent set of merges to master, there are a lot more resolvable properties and "inheriting" a resolver is a lot more likely. Remember, the order of resolution within Sceptre is:

1. Gather raw stack group config variables from higher-level stack group config.yaml files. Since the yaml resolver constructs the resolvers, these variables might not be strings but instead might be raw resolvers.
2. (**"Render time"**) Use those stack group config and user variables to _render_ the config file using Jinja into a string. Thus `{{ stack_group_config.template_bucket_name }}` would cast the resolver to string rather than resolve it.
3. Parse the string into yaml, which will instantiate resolvers on the Stack.
4. Use that yaml to create the Stack instance, which assigns those resolvers to resolvable properties
5. (**"Resolve time"**) Then, upon accessing those resolvable properties, those resolvers are resolved into values.

Because of this, you cannot access resolved values at render time (step 2 above). This is the whole reason why I made the `!stack_attr` resolver, because we need to access other stack attributes (which are set at step 4 and resolved at step 4). Simply accessing the pre-resolved values at render time doesn't work. 

> if it doesn't then can you please provide some explanation as to why it would not work?

I hope the above works. However, I was thinking that one thing that hasn't been documented in Sceptre's docs is the "Stack construction lifecycle", similar to what I just wrote above. I'll add a docs page that explains this, since I find it's frequently confused in the conversations I've had with you and other maintainers lately.

> does the template handler doc need updating? https://github.com/Sceptre/sceptre-sam-handler#arguments section.

Yeah, it will. However, I didn't want to update it until this change has been released as a version. Otherwise, I'd be documenting functionality that doesn't exist yet. The extension repos make more sense if they "chase" Sceptre versions rather than attempt to preempt them.

docs/_source/docs/resolvers.rst
jfalkenstein: > does ordering matter? can you put parameters above sceptre_user_data?

Sure? I'm not sure why it matters. Remember render time is prior to resolve time, so the order they appear in on the template doesn't really matter. In this particular case, I wanted to introduce the data we're selecting first before selecting it. 

> Would it be possible to use more concrete examples for key and the values for it? Also could we add a comment for my_parameter telling users what !stack_attr sceptre_user_data.key.1 will pass to my_parameter?

Yeah, I could definitely do that.

jfalkenstein:
@zaro0508 I've added more examples and a new documentation section on the resolution order of values on a StackGroup config. Hopefully this clears everything up for you.

tests/test_resolvers/test_cache.py
```diff
@@ -1,9 +0,0 @@
-# -*- coding: utf-8 -*-
```
jfalkenstein, line 1: I found these tests. Somebody committed these and I'm not sure why they still exist!

ykhalyavin:
Honestly, it feels like this resolver adds lots of ambiguity and resolving resolvable properties should be a part of sceptre. E.g. at render time sceptre could check if a stack_group_config is resolvable property, then resolve it and substitute the resolved value transparently to the user.

jfalkenstein:
@ykhalyavin What you suggest is not possible. As I documented on resolution order, resolvers require a Stack and the Stack doesn't exist at render time.

docs/_source/docs/resolvers.rst
zaro0508: > The problem here is that template_bucket_name (and many other items that might be in stack_group_config) might be resolvers.

What if `template_bucket_name` is not a resolver?  I have template_bucket_name set to a static value in my StackGroupConfig (config.yaml) files.  Should I not be able to use {{ stack_group_config.template_bucket_name }} in my stack config files in this case?  I'm just saying that is is documenting for a specific case and you should make it clear why `!stack_attr` needs to be used instead of `{{ stack_group_config.template_bucket_name }}`

docs/_source/docs/stack_config.rst
```diff
@@ -363,6 +363,56 @@ Syntax:
 When compiled, ``sceptre_user_data`` would be the dictionary
 ``{"iam_policy_file": "/path/to/policy.json"}``.
 
+Resolution order of values
```
zaro0508, line 366: thanks for adding this info to the doc.

docs/_source/docs/stack_config.rst
```diff
@@ -363,6 +363,56 @@ Syntax:
...
+   both those inherited from StackGroup Configs and those from the present Stack Config, will be
+   connected to the Stack instance.
+6. The first time a resolvable configuration is *accessed* is when the resolver(s) at that
+   configuration will be resolved and replaced with their resolved value. This is normally done at
+   the very last moment, right when it is needed (and not before).
+
+A common point of confusion tends to be around the distinction between "render time" (phase 3, when
+Jinja logic is applied) and "resolve time" (phase 6, when resolvers are resolved). You cannot use
+a resolver during "render time", since the resolver won't exist or be ready to use yet. You can,
+however, use Jinja logic to indicate *whether*, *which*, or *how* a resolver is configured.
```
zaro0508, line 405: i think this is the reason for using `!stack_attr` resolver?  if so maybe tie this info back to that resolver and vice versa?

zaro0508 (changes requested):

docs/_source/docs/resolvers.rst
jfalkenstein: Great point! I'll make that change.

docs/_source/docs/stack_config.rst
jfalkenstein: Yeah, I can do that. Thanks for the input!

docs/_source/conf.py
```diff
@@ -209,7 +209,9 @@
     ('py:class', 'TextIO'),
     ('py:class', '_io.StringIO'),
     ('py:class', 'yaml.loader.SafeLoader'),
-    ('py:class', 'yaml.dumper.Dumper')
+    ('py:class', 'yaml.dumper.Dumper'),
+    ('py:class',  'cfn_tools.odict.ODict'),
+    ('py:class',  'T_Container')
```
jfalkenstein, line 214: There were some warnings being output by Sphinx about it not knowing how to resolve docs for these classes.

docs/_source/docs/permissions.rst
```diff
@@ -125,10 +125,10 @@ Tips for working with Sceptre, IAM, and a CI/CD system
     permissions of the deployment role to only certain paths, preventing the deployment role from
     elevating its own permissions or modifying unrelated roles and policies.
   * Using ``aws:CalledVia`` and ``aws:CalledViaFirst`` conditions matching against
-    ``"cloudformation.amazonaws.com"`` on non-cloudformation actions to ensure that the deployment
```
jfalkenstein, line 128: "on non-cloudformation actions" made no sense in this context. I fixed that.

jfalkenstein:
Alright, @zaro0508, I've refined the docs some, making the changes you've requested.

zxiiro (approved):

zxiiro (commented):
Sorry miss-clicked approved earlier when I meant to leave a comment.

docs/_source/docs/permissions.rst
```diff
@@ -125,10 +125,10 @@ Tips for working with Sceptre, IAM, and a CI/CD system
...
-    role can only execute changes via CloudFormation and not on its own. Note: Some actions are
-    taken by Sceptre directly and not via cloudformation (see the section below on this). Those
-    actions should *not* have a CalledVia condition applied.
+    ``"cloudformation.amazonaws.com"`` to ensure that the deployment role can only execute changes
+    via CloudFormation and not on its own. Note: Some actions are taken by Sceptre directly and not
+    via cloudformation (see the section below on this). Those actions should *not* have a CalledVia
```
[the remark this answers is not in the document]
jfalkenstein: Yeah, I can do that.

docs/_source/docs/resolvers.rst
```diff
@@ -79,6 +79,53 @@ A resolver to execute any shell command.
...
+           - "some random value"
+           - "the value we want to select"
+
+   iam_role: !stack_output roles.yaml::RoleArn
+
+   parameters:
+       # This will pass the value of "the value we want to select" for my_parameter
+       my_parameter: !stack_attr sceptre_user_data.key.1
+       # You can also access the value of another resolvable property like this:
+       use_role_arn: !stack_attr iam_role
```
zaro0508, line 126: is it possible to reference `stack_name`?  something like this?
```
parameters:
  ThisStack: !stack_attr stack_name
```

i'm wondering if it could be another work around for https://github.com/Sceptre/sceptre/discussions/1021

zaro0508 (approved):

docs/_source/docs/resolvers.rst
jfalkenstein: Yes! That would work perfectly.

PR #1204, opened 2022-02-19
[Partially Resolve #1193] Automatically adding Sceptre tags to all stacks

### LINKED ISSUE #1193: Feature request: handling deleted stacks via launch
Right now, `sceptre launch` will create new stacks and update existing stacks automatically. However, Sceptre doesn't currently see stacks that exist remotely but do not currently have a config locally. The only way to delete a stack via Sceptre is to have its Stack Config locally and run `sceptre delete` explicitly. But if the Stack Config doesn't exist, Sceptre can't even do that.

## Requested behavior
When you run `sceptre launch --allow-delete` it would locate existing stacks that are in CloudFormation but do not currently exist in the local project and then delete them. When prompting the user before launch begins (unless -y is passed), it should list the stacks to be deleted separately to make it clear what deletions will occur.

### Why the `--allow-delete` flag?
There are many situations where a user might be on a different branch or otherwise not want to delete stacks that aren't fully represented locally. As a result, it should be opt-in behavior; Otherwise it could have disastrous consequences.

## How would it find remote-only stacks without state management?
It's important that Sceptre continues to avoid using a state management DB or file, since that's one of the "selling points" of Sceptre.

Instead, I propose that Sceptre be able to find remote-only stacks by:
1. Every stack that Sceptre creates or updates should always add a "sceptre_project_code" tag and a "sceptre_name" tag automatically (in addition to any other configured stack tags).
    * _Automatic tagging for state management is pretty common for infra-as-code tools_. CDK, SAM CLI, and others use this mechanism to store metadata on stacks for later reference.
 2. To locate remote stacks for a given command path would involve:
    1. Gathering the project code for that command path 
    2. Listing all non-deleted cloudformation stacks and collecting all those stacks whose name _starts_with_ the project code.
         * This could produce false-positives, though, in case there's a project code that shares the same start as another project code. 
     3. Describing each of the gathered stacks to retrieve their tags
     4. Selecting only those stacks that share the correct project_code tag and whose stack_name starts with the command path.

Doing this should provide a list of stack names that are remote but not locally represented in configs.

jfalkenstein:
This PR adds some basic Sceptre stack metadata to every stack, namely:
* sceptre_name: This is the "sceptre name", that will look like "outer-stack-group/inner-group/stack-name". Regardless of whether the real stack's name has been set via the "stack_name" attribute or automatically by Sceptre by default, this name will be consistent with the actual directory/StackGroup structure of the sceptre project that launched it.
* sceptre_project_code: This is the project_code of the sceptre project that launched the stack.

This is the **first step** toward accomplishing the goals of #1193, namely providing the _means_ for Sceptre to locate stacks that were _previously_ associated with the project but whose stack configs have been deleted. We will need further code changes to fully implement #1193, but they will build on these changes.

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/config/reader.py
+++ b/sceptre/config/reader.py
@@ -13,7 +13,7 @@ import datetime
 import fnmatch
 import logging
 from os import environ, path, walk
-from typing import Set
+from typing import Set, Tuple, Dict
 
 from pkg_resources import iter_entry_points
 from pathlib import Path
@@ -188,7 +188,7 @@ class ConfigReader(object):
         node.tag = loader.resolve(type(node), node.value, (True, False))
         return node
 
-    def construct_stacks(self) -> Set[Stack]:
+    def construct_stacks(self) -> Tuple[Set[Stack], Set[Stack]]:
         """
         Traverses the files under the command path.
         For each file encountered, a Stack is constructed
@@ -524,9 +524,8 @@ class ConfigReader(object):
                     )
                 )
 
-        s3_details = self._collect_s3_details(
-            stack_name, config
-        )
+        s3_details = self._collect_s3_details(stack_name, config)
+        stack_tags = self._collect_stack_tags(stack_name, config)
         stack = Stack(
             name=stack_name,
             project_code=config["project_code"],
@@ -545,7 +544,7 @@ class ConfigReader(object):
             dependencies=config.get("dependencies", []),
             role_arn=config.get("role_arn"),
             protected=config.get("protect", False),
-            tags=config.get("stack_tags", {}),
+            tags=stack_tags,
             external_name=config.get("stack_name"),
             notifications=config.get("notifications"),
             on_failure=config.get("on_failure"),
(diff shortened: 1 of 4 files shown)


sceptre/config/reader.py
```diff
@@ -569,3 +568,18 @@ def _parsed_stack_group_config(self, stack_group_config):
...
+        """
+        Assembles the stack tags dictionary, including default stack tags.
+
+        :param stack_name:
+        :param config:
+        :return:
+        """
+        stack_tags = {
+            'sceptre_name': stack_name,
+            'sceptre_project_code': config['project_code'],
```
zaro0508, line 582: should we follow the cloudformation convention of using a `<prefix>:` ? https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-resource-tags.html

so something like `sceptre:name` and `sceptre:project-code`?

also, aws tags have limitations that we should check for? https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html

zaro0508 (changes requested):

jfalkenstein:
Looks like there are a number of integration tests I'm going to need to address on this one.

jfalkenstein:
I'm going to close this PR. I think the delete flag will be more appropriate of a solution. There's a real problem with the approach this PR was intending to allow; there's a massive amount of boto3 requests required to have certainty about which stacks have been abandoned.

PR #1210, opened 2022-03-06
[Resolve #1080] Added duration_seconds parameter to adopt DurationSeconds in boto

### LINKED ISSUE #1080: role_arn Session Timeout
I guess this is somewhat of a feature request.

An additional option when specifying role_arn to allow the use of a custom session timeout would be really helpful. I have a sceptre project which takes 3 hours to deploy.  My Role in AWS has a session timeout of 4hrs (which works when manually assuming the role with other tools) ex. the AWS CLI allows --session-duration

AWS Linux 2
Python 3.7
sceptre 2.6.2

Run a sceptre project that takes > 1hr

Session token expires.

shixuyue:
Added a new parameter `iam_role_session_duration` in at `stack` level so users can edit the session duration when `connection_manager` assumes `iam_role` that is provided. 

This new added parameter is an optional parameter. If not provided in the stack yamls, `connection_manager` will use default value of 3600s to assume `iam_role`. (same as before.)

Added type annotation for `stacks` class to ensure `iam_role_session_duration` is `int` type.

Also updated the documentation of how to use `iam_role_session_duration`.

## PR Checklist

## Other Information


shixuyue:
Resolve #1080 
User now can set `duration_seconds` in the yamls as they wanted, where the default is still 1 hour unchanged. In order to assume the `iam_role` for a longer session, `MaxSessionDuration` must be set in the cfn when `iam_role` is created (default is 1 hour). Thus there is no security hole.

sceptre/connection_manager.py
```diff
@@ -92,26 +92,27 @@ class ConnectionManager(object):
...
 
     def __repr__(self):
         return (
             "sceptre.connection_manager.ConnectionManager(region='{0}', "
-            "profile='{1}', stack_name='{2}', iam_role='{3}')".format(
-                self.region, self.profile, self.stack_name, self.iam_role
-            )
+            "profile='{1}', stack_name='{2}', iam_role='{3}', duration_seconds='{4}')".format(
+                self.region, self.profile, self.stack_name, self.iam_role, self.duration_seconds
+             )
```
jfalkenstein, line 112: I think this line is over-indented by a space.

sceptre/connection_manager.py
```diff
@@ -176,6 +178,8 @@ def _get_session(self, profile, region, iam_role):
                         )
 
                     self._boto_sessions[key] = session
+                    self.logger.info("session will expire at " +
+                                     credentials['Expiration'].strftime("%Y-%m-%d %H:%M:%S"))
```
jfalkenstein, line 182: I don't think this information is relevant to log.

sceptre/connection_manager.py
```diff
@@ -240,14 +244,15 @@ def call(
...
                 region = self.region
                 profile = self.profile
                 iam_role = self.iam_role
+                duration_seconds = self.duration_seconds
 
         if kwargs is None:  # pragma: no cover
             kwargs = {}
 
-        client = self._get_client(service, region, profile, stack_name, iam_role)
+        client = self._get_client(service, region, profile, stack_name, iam_role, duration_seconds)
```
jfalkenstein, line 257: I don't think anywhere in the Sceptre Codebase will be invoking this method with `duration_seconds`. I think it would be best to just use the `duration_seconds` instance variable.

sceptre/connection_manager.py
```diff
@@ -214,15 +218,15 @@ def _get_client(self, service, region, profile, stack_name, iam_role):
                     "No %s client found, creating one...", service
                 )
                 self._clients[key] = self._get_session(
-                    profile, region, iam_role
+                    profile, region, iam_role, duration_seconds
```
jfalkenstein, line 221: I think you'd want to use `self.duration_seconds` for this rather than needing it passed as a parameter.

sceptre/stack.py
```diff
@@ -116,6 +116,10 @@ class Stack(object):
     :param stack_group_config: The StackGroup config for the Stack
     :type stack_group_config: dict
 
+    :param duration_seconds: The session duration when Scetre assumes a role.\
+           If not supplied, Sceptre uses default value (3600 seconds)
+    :type duration_seconds: int
```
jfalkenstein, line 121: It's best to just use a type annotation to indicate types rather than explicitly documenting them in the docstrings.

sceptre/connection_manager.py
```diff
@@ -92,26 +92,27 @@ class ConnectionManager(object):
...
-            "profile='{1}', stack_name='{2}', iam_role='{3}')".format(
-                self.region, self.profile, self.stack_name, self.iam_role
-            )
+            "profile='{1}', stack_name='{2}', iam_role='{3}', duration_seconds='{4}')".format(
+                self.region, self.profile, self.stack_name, self.iam_role, self.duration_seconds
+             )
         )
 
-    def _get_session(self, profile, region, iam_role):
+    def _get_session(self, profile, region, iam_role, duration_seconds):
```
jfalkenstein, line 115: Same as before, I think it duration_seconds shouldn't be a parameter here; Just use the instance variable.

jfalkenstein (changes requested):
I'm not opposed to this change, but I'd like to see a few things changed here:
1. You need to update the docs on this. Be sure to indicate in the documentation about how the IAM role needs to permit longer session durations in order for longer than the default 1-hour session to be valid.
2. Let's find a better name for the configuration other than "duration_seconds", something that clearly identifies this as corresponding to the `iam_role`'s session.
3. I'm wondering why we would want to pass the duration as a parameter to the private methods on ConnectionManager. I know that this change as you have it will break https://github.com/Sceptre/sceptre-sam-handler/blob/main/sam_handler/handler.py#L115~L118. I don't see much reason to make it a parameter on those private methods if it's already an instance variable.

shixuyue:
> I'm not opposed to this change, but I'd like to see a few things changed here:
> 
> 1. You need to update the docs on this. Be sure to indicate in the documentation about how the IAM role needs to permit longer session durations in order for longer than the default 1-hour session to be valid.
> 2. Let's find a better name for the configuration other than "duration_seconds", something that clearly identifies this as corresponding to the `iam_role`'s session.
> 3. I'm wondering why we would want to pass the duration as a parameter to the private methods on ConnectionManager. I know that this change as you have it will break https://github.com/Sceptre/sceptre-sam-handler/blob/main/sam_handler/handler.py#L115~L118. I don't see much reason to make it a parameter on those private methods if it's already an instance variable.

- I have added the documentation of this new added variable.
- Also I renamed `duration_seconds` to `iam_role_session_duration`.
- And I have removed all `iam_role_session_duration` parameters in `ConnectionManager`, it uses the instance variable when needed.

docs/_source/docs/permissions.rst
```diff
@@ -86,6 +86,18 @@ As a resolvable property, Sceptre allows you to use a resolver to populate the `
 Stack or StackGroup Config. This means you could define that role within your project, output its
 ARN, and then reference it using `!stack_output`.
 
+iam_role_session_duration
```
jfalkenstein, line 89: I don't we need this parameter documented on this page. This page is all around designing Sceptre permissions, but this parameter is really just a detail. The Stack Config docs is probably the only place we need to document it.

docs/_source/docs/stack_config.rst
```diff
@@ -243,6 +244,23 @@ stack. The ``iam_role`` configuration does not configure anything on the stack i
...
+* Can be inherited from StackGroup: Yes
+* Inheritance strategy: Overrides parent if set
+
+This is the session duration when **Sceptre** *assumes* the **iam_role** IAM Role using AWS STS when
+executing any actions on the Stack.
+
+.. warning::
+
+   If you set the value of ``iam_role_session_duration`` to a number that *GREATER* than 3600, you will need
+   to make sure that the ``iam_role`` has a configuration of ``MaxSessionDuration``, and its value is *GREATER* than the value of ``iam_role_session_duration``.
```
jfalkenstein, line 259:
```suggestion
   to make sure that the ``iam_role`` has a configuration of ``MaxSessionDuration``, and its value is *GREATER* than or equal to the value of ``iam_role_session_duration``.
```

jfalkenstein: Also, the length of this line seems pretty wide. We should try to stick to around 100 chars wide.

sceptre/connection_manager.py
```diff
@@ -92,22 +92,23 @@ class ConnectionManager(object):
     _clients = {}
     _stack_keys = {}
 
-    def __init__(self, region, profile=None, stack_name=None, iam_role=None):
+    def __init__(self, region, profile=None, stack_name=None, iam_role=None, iam_role_session_duration=3600):
```
jfalkenstein, line 95: Let's make the 3600 a properly named constant on this module, I think. Something like DEFAULT_IAM_SESSION_DURATION, or something like that.

sceptre/connection_manager.py
```diff
@@ -92,22 +92,23 @@ class ConnectionManager(object):
...
         self.logger = logging.getLogger(__name__)
 
         self.region = region
         self.profile = profile
         self.stack_name = stack_name
         self.iam_role = iam_role
+        self.iam_role_session_duration = iam_role_session_duration if iam_role_session_duration else 3600
```
jfalkenstein, line 102: When would this value be falsey? I'm not sure when this would be applicable, considering your code.

Also, pro tip: You could do this same thing in a less verbose way like:
```python
        self.iam_role_session_duration = iam_role_session_duration or 3600
```

sceptre/connection_manager.py
```diff
@@ -157,7 +164,8 @@ def _get_session(self, profile, region, iam_role):
                     session_name = f'{iam_role.split("/")[-1][:56]}-session'
                     sts_response = sts_client.assume_role(
                         RoleArn=iam_role,
-                        RoleSessionName=session_name
+                        RoleSessionName=session_name,
+                        DurationSeconds=self.iam_role_session_duration
                     )
```
jfalkenstein, line 169: Based upon how it's instantiated, `self.iam_role_session_duration` is going to be `None` sometimes, which you've already said IAM rejects. Therefore, I think you should do something like this:
```python
assume_role_kwargs = {
    'RoleArn': iam_role,
    'RoleSessionName': session_name,
}
if self.iam_role_session_duration:
    assume_role_kwargs['DurationSeconds'] = self.iam_role_session_duration
sts_response = sts_client.assume_role(**assume_role_kwargs)
```

jfalkenstein (changes requested):
I'm a little concerned with how we're handling default values for this thing:
* reader.py will always pass a default of `None` to Stack
* Stack sets the default value to a magic number of 3600
* Connection Manager sets the default to a constant, where the default value is 3600

Rather than trying to set a default value of 3600 in a bunch of places (and it failing), I think the better approach is to let it be None in all the places and simply not to pass DurationSeconds if it's None.

sceptre/config/reader.py
```diff
@@ -537,6 +539,7 @@ def _construct_stack(self, rel_path, stack_group_config=None):
             template_key_prefix=config.get("template_key_prefix"),
             required_version=config.get("required_version"),
             iam_role=config.get("iam_role"),
+            iam_role_session_duration=config.get("iam_role_session_duration"),
```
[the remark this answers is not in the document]
shixuyue: I didn't realized that this returns None for the most of the cases. I think I should update the logic of how to pass `iam_role_session_duration` to boto based on `None` is the most of the cases.

jfalkenstein:
This looks good to me. I'm running integration tests now. Once they pass, I'll approve and then likely merge. @zaro0508 , you don't have an issue with this change, do you? It seems to be pretty straight-forward and I can see the need. Blast radius seems pretty constrained on this change too.

shixuyue:
> This looks good to me. I'm running integration tests now. Once they pass, I'll approve and then likely merge. @zaro0508 , you don't have an issue with this change, do you? It seems to be pretty straight-forward and I can see the need. Blast radius seems pretty constrained on this change too.

Sounds good. I have one integration test fails on my end. But I believe this is not related to my change: 
```
Feature: Stack output external resolver # integration-tests/features/stack-output-external-resolver.feature:1

  Scenario: launch a stack referencing the external output of an existing stack  # integration-tests/features/stack-output-external-resolver.feature:3
    Given stack "9/A" exists using "dependencies/independent_template.json"      # integration-tests/steps/stacks.py:85 7.505s
    And stack "9/B" does not exist                                               # integration-tests/steps/stacks.py:36 0.830s
    When the user launches stack "9/B"                                           # integration-tests/steps/stacks.py:267 0.959s
    Then stack "9/B" exists in "CREATE_COMPLETE" state                           # integration-tests/steps/stacks.py:354 0.857s
        File "/home/xuyue/.local/lib/python3.8/site-packages/behave/model.py", line 1329, in run
          match.run(runner.context)
        File "/home/xuyue/.local/lib/python3.8/site-packages/behave/matchers.py", line 98, in run
          self.func(context, *args, **kwargs)
        File "/mnt/c/sceptre/integration-tests/steps/stacks.py", line 365, in step_impl
          assert (status == desired_status)
      AssertionError

      Captured logging:
      INFO:sceptre.plan.actions:9/B - Launching Stack
      INFO:sceptre.plan.actions:9/B - Stack is in the PENDING state
      INFO:sceptre.plan.actions:9/B - Creating Stack
      INFO:botocore.credentials:Found credentials in shared credentials file: ~/.aws/credentials


Feature: Stack output resolver # integration-tests/features/stack-output-resolver.feature:1
```

shixuyue:
I saw the integration test fails with error with `MaxDurations`
The integration tests set up uses a service role
```
      - run:
          name: 'Integration Testing'
          command: |
            . ./venv/bin/activate
            mkdir -p ~/.aws
            echo -e "[default]\nregion=eu-west-1\nsource_profile=default\nrole_arn=arn:aws:iam::582448526747:role/sceptre-integration-test-ServiceRole-DUSW2V6ES2ZU" > ~/.aws/config
            echo -e "[default]\nregion=eu-west-1\naws_access_key_id=$CI_SERVICE_AWS_ACCESS_KEY_ID\naws_secret_access_key=$CI_SERVICE_AWS_SECRET_ACCESS_KEY" > ~/.aws/credentials
            behave --junit \
                   --junit-directory build/behave \
                   $(circleci tests glob "integration-tests/features/*.feature" | circleci tests split --split-by=timings)
```

which will result a role chaining, and AWS only allows 1 hour session for all role chaining assumes.

jfalkenstein:
Then I suggest you put up a pull request to update our IAM role here: https://github.com/Sceptre/sceptre-aws/blob/master/templates/service-account.yaml

shixuyue:
> Then I suggest you put up a pull request to update our IAM role here: https://github.com/Sceptre/sceptre-aws/blob/master/templates/service-account.yaml

I dont think that will help. The tests use an user to assume the service role then assume the testing role. This resulted a role chaining, and AWS set the max duration of role chaining to 1 hour.
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_terms-and-concepts.html

jfalkenstein:
huh.... interesting. I'm not prepared to alter the existing IAM setup for our integration tests. Perhaps you can make the tests use a MaxDuration of <3600s. It's not like you were planning on making the tests sleep for >1hr to test that we did indeed let the session last long enough. So maybe you can set the iam_role_session_duration to something like 1000 or something less than 1 hour. That way, you can test that it works and doesn't blow up. Short of finding a way to get the current session info and inspect it to see that the duration is what we set it to, I'm not sure how else to test this in integration tests. But at least testing that setting that value doesn't blow up might be a good route to go.

shixuyue:
> huh.... interesting. I'm not prepared to alter the existing IAM setup for our integration tests. Perhaps you can make the tests use a MaxDuration of <3600s. It's not like you were planning on making the tests sleep for >1hr to test that we did indeed let the session last long enough. So maybe you can set the iam_role_session_duration to something like 1000 or something less than 1 hour. That way, you can test that it works and doesn't blow up. Short of finding a way to get the current session info and inspect it to see that the duration is what we set it to, I'm not sure how else to test this in integration tests. But at least testing that setting that value doesn't blow up might be a good route to go.

oh yea, thats very smart. I didnt think of that.

jfalkenstein:
@shixuyue I'm ready to approve and merge this to go out in our next release. Here's what I'd like you to do before I do, however:

1. Please name your PR according to our naming policy (i.e. begin it with `[Resolve #...]`, referencing the issue this corresponds to.
2. Update the PR description with a solid description of this change that you would like included in the commit history.
3. Let me know when you've done numbers 1 and 2 above.

shixuyue:
@jfalkenstein 
I have updated the title and description, let me know if I need to include more details in the description. I am generally pretty bad at writing the descriptions.....

PR #1217, opened 2022-04-26
Proof of concept list sceptre-user-data

alexharv074:
This implements a proof of concept command list sceptre-user-data for
discussion.

[Your PR description here]

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/cli/list.py
+++ b/sceptre/cli/list.py
@@ -1,4 +1,5 @@
 import click
+import yaml
 
 from sceptre.context import SceptreContext
 from sceptre.cli.helpers import (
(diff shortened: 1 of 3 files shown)


alexharv074:
Example output

Screen Shot 2022-04-27 at 1 48 58 am

jfalkenstein:
The thread where you broached this idea in slack didn't say you were _just_ targeting sceptre user data. I thought you wanted the whole StackConfig. 

While I thing there would be real utility in reproducing the resolved StackConfig, I don't really think it would be that useful for the command to JUST report the sceptre_user_data.

sceptre/cli/list.py
```diff
@@ -156,11 +156,11 @@ def list_stacks(ctx, path):
     write(output, output_format)
 
 
-@list_group.command(name="sceptre_user_data")
+@list_group.command(name="config")
 @click.argument("path")
 @click.pass_context
 @catch_exceptions
-def list_stacks(ctx, path):
+def list_config(ctx, path):
```
jfalkenstein, line 163: It doesn't really seem like a `list` command, though. Maybe `render` would be more accurate.

sceptre/plan/plan.py
```diff
@@ -35,6 +35,7 @@ def __init__(self, context):
 
         config_reader = ConfigReader(context)
         all_stacks, command_stacks = config_reader.construct_stacks()
+        self.read = config_reader.read(self.context.command_path)
```
jfalkenstein, line 38: Why not add a `read()` method instead that would do this instead? Also, what happens if the `command_path` is a StackGroup? I suspect that will actually fail.

jfalkenstein: What you _could_ do is `read_config()` method on the Actions class that would be like this:

```python
def read(self, config_reader):
    return config_reader.read(self.stack.name)
```

^^ I think that might work. That way, SceptrePlan would ultimately map out to actions and operate on StackGroups or individual Stacks correctly, like it does for everything else. You could pass the ConfigReader as a method parameter to the Actions (since it won't have access to that otherwise).

alexharv074:
Closing in favour of #1220

PR #1220, opened 2022-04-29
Implement dump config

alexharv074:
This adds a feature `dump config` that dumps the rendered config
from the `ConfigReader`, showing config after Jinja2 interpolation.

The feature is intended to assist with debugging.

## PR Checklist

## Other Information


alexharv074:
Example output

Screen Shot 2022-04-30 at 2 08 23 am

sceptre/cli/list.py
```diff
@@ -153,3 +153,30 @@ def list_stacks(ctx, path):
     output = {f"{stack.name}.yaml": stack.external_name for stack in plan.graph}
     output_format = "json" if context.output_format == "json" else "yaml"
     write(output, output_format)
+
+
+@list_group.command(name="config")
+@click.argument("path")
+@click.pass_context
+@catch_exceptions
+def list_config(ctx, path):
```
alexharv074, line 162: @jtfalkenstein comment from #1217 
> It doesn't really seem like a list command, though. Maybe render would be more accurate.

alexharv074: @jfalkenstein It seems like a list command to me. We are listing the config keys. I guess it could be argued that we rendering too, although I would expect the output from render to be the jinja2 rendered template similar to `generate`. Thoughts?

sceptre/plan/plan.py
```diff
@@ -35,6 +35,7 @@ def __init__(self, context):
 
         config_reader = ConfigReader(context)
         all_stacks, command_stacks = config_reader.construct_stacks()
+        self.read = config_reader.read(self.context.command_path)
```
alexharv074, line 38: @jfalkenstein comments from #1217 

> Why not add a `read()` method instead that would do this instead? Also, what happens if the `command_path` is a StackGroup? I suspect that will actually fail.
>
> What you _could_ do is `read_config()` method on the Actions class that would be like this:
>
> ```python
> def read(self, config_reader):
>    return config_reader.read(self.stack.name)
> ```
>
> ^^ I think that might work. That way, SceptrePlan would ultimately map out to actions and operate on StackGroups or individual Stacks correctly, like it does for everything else. You could pass the ConfigReader as a method parameter to the Actions (since it won't have access to that otherwise).

alexharv074:
Updated output:

Screen Shot 2022-04-30 at 3 56 57 pm

sceptre/plan/actions.py
```diff
@@ -1141,3 +1141,10 @@ def _describe_stack_resource_drifts(self) -> dict:
                 "StackName": self.stack.external_name
             }
         )
+
+    @add_stack_hooks
+    def list_config(self, config_reader, command_path):
+        """
+        Returns the Template for the Stack
+        """
+        return config_reader.read(command_path)
```
jfalkenstein, line 1150: You shouldn't pass the command path here. On StackActions, each method operates on a single stack; The command path might be a reference to a StackGroup. Rather, you need to find a way to invoke the passed config_reader from attributes on `self.stack`. I believe `self.stack.name` will get you most of the way there.

sceptre/plan/actions.py
```diff
@@ -1141,3 +1141,10 @@ def _describe_stack_resource_drifts(self) -> dict:
                 "StackName": self.stack.external_name
             }
         )
+
+    @add_stack_hooks
+    def list_config(self, config_reader):
+        """
+        Returns the Template for the Stack
+        """
+        return config_reader.read(self.stack.name + ".yaml")
```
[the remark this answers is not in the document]
alexharv074: @jfalkenstein I am not following because, as far as I know, `self.stack.name + ".yaml"` is a file name, not a path:

```
ipdb> self.stack.name + ".yaml"
'aurora.yaml'
```

sceptre/plan/plan.py
```diff
@@ -394,3 +394,10 @@ def drift_show(self, *args) -> Dict[Stack, str]:
         """
         self.resolve(command=self.drift_show.__name__)
         return self._execute(*args)
+
+    def dump_config(self, *args):
```
jfalkenstein, line 398: I don't know why all the commands on this module just take arbitrary `*args`. Most of those commands don't actually take _any_ arguments, but in this case, you do require one. I'd suggest you explicitly define the argument you need here, just like you did on actions.py.

sceptre/cli/dump.py
```diff
@@ -0,0 +1,57 @@
...
+    pass
+
+
+@dump_group.command(name="config")
+@click.argument("path")
+@click.pass_context
+@catch_exceptions
+def dump_config(ctx, path):
+    """
+    Dump config.
```
jfalkenstein, line 28: I'd like to see a better description on this.

sceptre/cli/dump.py
```diff
@@ -0,0 +1,57 @@
...
+    pass
+
+
+@dump_group.command(name="config")
+@click.argument("path")
+@click.pass_context
+@catch_exceptions
+def dump_config(ctx, path):
+    """
+    Dump the rendered (post-Jinja) Stack Configs as yaml.
```
jfalkenstein, line 28: It's not just yaml. If output format is set to json, it outputs as json.

alexharv074: Ok, deleted 'as yaml'.

jfalkenstein (approved):

PR #1227, opened 2022-05-26
[Resolves #1225] Support python 3.10

### LINKED ISSUE #1225: Add Python 3.10 support
### Subject of the issue
We would like to update to Python 3.10. Unfortunately, Sceptre does not yet seem compatible. If we try to run it, we get the following error message:

```
pkg_resources.DistributionNotFound: The 'importlib-metadata; python_version < "3.8"' distribution was not found and is required by click, jsonschema
```

This seems to be a bug in pkg_resources and a version upgrade is probably needed.

### Your environment
* version of sceptre (sceptre --version): 3.1.0
* version of python (python --version): 3.10.4
* which OS/distro: alpine

### Steps to reproduce
Just run any sceptre command after installing sceptre with python 3.10

zaro0508:
Run tests against python 3.10

depends on #1240

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -137,7 +137,7 @@ installed on your system.
 Run unit tests and coverage using tox for multiple Python versions:
 
 ```bash
-$ tox -e py36 -e py37
+$ tox -e py38 -e py39
 ```
 
 If you use pyenv to manage Python versions, try `pip install tox-pyenv` to make
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -131,17 +131,18 @@ $ make coverage
 ```
 
 Note: Sceptre aims to be compatible with Python 3, please run unit test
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -137,7 +137,7 @@ installed on your system.
 Run unit tests and coverage using tox for multiple Python versions:
 
 ```bash
-$ tox -e py36 -e py37
+$ tox -e py38 -e py39
```
jfalkenstein, line 140: What about py37?

jfalkenstein: Or py3.10?

zaro0508: We are going to assume devs know about `tox` and how to use it.  I've updated the doc with that in mind.  Let me know what you think

jfalkenstein (approved):
Thanks for all your work here!

PR #1239, opened 2022-06-10
Poetry Take 2

tarkatronic:
After trying to update the pre-existing Poetry PR, and discovering numerous conflicts, I dediced it would be wise to start fresh. Luckily I was able to copy over much of my 
work from previously, and enough things have been fixed now that this transition has been much easier this time around, so far.

I will need to go back through the other PR to find the full list of work that was necessary for this transition, but so far:

* Tox passes for all environments
* Docs build
* Make commands work

So I'd call this a pretty good start so far!

## PR Checklist

## Other Information

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+poetry.lock binary
--- a/Makefile
+++ b/Makefile
@@ -51,66 +51,60 @@ clean-test:
 	rm -f test-results.xml
 
 lint:
-	pre-commit run --all-files --show-diff-on-failure
+	poetry run pre-commit run --all-files --show-diff-on-failure
 
 test:
-	pytest
+	poetry run pytest
 
 test-all:
-	tox --parallel=auto
+	poetry run tox --parallel=auto
 
 test-integration: install
-	behave integration-tests/
+	poetry run behave integration-tests/
 
 docs:
 	rm -f docs/sceptre.rst
 	rm -f docs/modules.rst
-	sphinx-apidoc -o docs/ sceptre
-	$(MAKE) -C docs clean
-	$(MAKE) -C docs html
+	poetry run sphinx-apidoc -o docs/ sceptre
+	poetry run $(MAKE) -C docs clean
+	poetry run $(MAKE) -C docs html
 	$(BROWSER) docs/_build/html/index.html
 
 docs-latest:
-	$(MAKE) -C docs build-latest
+	poetry run $(MAKE) -C docs build-latest
 
 docs-build-tag:
-	$(MAKE) -C docs build-tag
+	poetry run $(MAKE) -C docs build-tag
 
 docs-build-dev:
-	$(MAKE) -C docs build-dev
+	poetry run $(MAKE) -C docs build-dev
 
 docs-build-commit:
-	$(MAKE) -C docs build-commit
+	poetry run $(MAKE) -C docs build-commit
 
 docs-serve-latest:
-	$(MAKE) -C docs serve-latest
+	poetry run $(MAKE) -C docs serve-latest
 
 docs-serve-tag:
-	$(MAKE) -C docs serve-tag
+	poetry run $(MAKE) -C docs serve-tag
 
 docs-serve-dev:
-	$(MAKE) -C docs serve-dev
+	poetry run $(MAKE) -C docs serve-dev
 
 docs-serve-commit: docs-commit
-	$(MAKE) -C docs serve-commit
+	poetry run $(MAKE) -C docs serve-commit
 
 docs-install:
-	$(MAKE) -C docs install
+	poetry run $(MAKE) -C docs install
 
 docs-clean:
-	$(MAKE) -C docs clean
+	poetry run $(MAKE) -C docs clean
 
 dist: clean
-	python setup.py sdist
-	python setup.py bdist_wheel
-	twine check dist/*
-	ls -l dist
+	poetry build
 
 install: clean
-	pip install .
+	poetry install --no-dev --remove-untracked
 
 install-dev: clean
-	pip install -r requirements/prod.txt
-	pip install -r requirements/dev.txt
-	pip install -e .
-	@echo "To install the documentation dependencies, run:\ncd docs\nmake install"
+	poetry install --remove-untracked
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,6 +77,9 @@ pytest-cov = "^3.0.0"
 pytest-sugar = "^0.9.4"
 tox = "^3.25.0"
 Sphinx = "<5.0"
+sphinx-click = "^4.1.0"
+sphinx-rtd-theme = "^1.0.0"
+sphinx-autodoc-typehints = "<1.18"
 
 [tool.poetry.scripts]
 sceptre = "sceptre.cli:cli"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
-build-backend = "setuptools.build_meta"
-requires = ["setuptools", "wheel"]
+requires = ["poetry-core>=1.0.0"]
+build-backend = "poetry.core.masonry.api"
 
 [tool.coverage.run]
 branch = true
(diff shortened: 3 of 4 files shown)


pyproject.toml
```diff
@@ -22,6 +22,85 @@ ignore_errors = true
 fail_under = 90.0
 show_missing = true
 
+[tool.poetry]
+name = "sceptre"
+version = "3.1.0"
```
zaro0508, line 27: how does this work with version in `sceptre/__init__.py`?

pyproject.toml
```diff
@@ -22,6 +22,85 @@ ignore_errors = true
...
+description = "Cloud Provisioning Tool"
+authors = ["Cloudreach <sceptre@cloudreach.com>"]
```
zaro0508, line 29: Maybe we should remove the email?

pyproject.toml
```diff
@@ -22,6 +22,85 @@ ignore_errors = true
...
+  "Programming Language :: Python :: 3.7",
+  "Programming Language :: Python :: 3.8",
+  "Programming Language :: Python :: 3.9",
+]
+packages = [
+  { include = "sceptre/" },
+]
+include = ["CHANGELOG.md", "README.md"]
+
+[tool.poetry.dependencies]
```
zaro0508, line 51: how does this work with the current set of requirements in `requirements` folder?

pyproject.toml
tarkatronic: That will likely go away; this is just a very preliminary first pass to make the basics work, so I haven't gone through everything yet.

pyproject.toml
tarkatronic: Those all go away; they're no longer needed since all dependencies/requirements are handled by Poetry.

zaro0508 (changes requested):
Generally looks good however this PR is failing in CI. Also I would like to see some documentation added in the Contributions  doc to at least tell users that they need to setup poetry to run builds and tests.

pyproject.toml
jfalkenstein: The version in sceptre/__init__.py is actually used with the `require_version` or whatever it's called flag on StackConfigs.

pyproject.toml
```diff
@@ -22,6 +22,85 @@ ignore_errors = true
...
+stack_output = "sceptre.resolvers.stack_output:StackOutput"
+stack_output_external = "sceptre.resolvers.stack_output:StackOutputExternal"
+no_value = "sceptre.resolvers.no_value:NoValue"
+stack_attr = "sceptre.resolvers.stack_attr:StackAttr"
+
+[tool.poetry.plugins."sceptre.template_handlers"]
+file = "sceptre.template_handlers.file:File"
+s3 = "sceptre.template_handlers.s3:S3"
+http = "sceptre.template_handlers.http:Http"
+
```
jfalkenstein, line 103: This looks like it is meant to replace `setup.py`. Is that True? If so, why is setup.py not removed in this PR?

pyproject.toml
```diff
@@ -22,6 +22,85 @@ ignore_errors = true
...
+tox = "^3.25.0"
+Sphinx = "<5.0"
+sphinx-click = "^4.1.0"
+sphinx-rtd-theme = "^1.0.0"
+sphinx-autodoc-typehints = "<1.18"
+
+[tool.poetry.scripts]
+sceptre = "sceptre.cli:cli"
+
+[tool.poetry.plugins."sceptre.hooks"]
```
jfalkenstein, line 87: What's the point of the quotes here?

.gitattributes
```diff
@@ -0,0 +1 @@
+poetry.lock binary
```
jfalkenstein, line 1: What does this do?

Makefile
```diff
@@ -51,66 +51,60 @@ clean-test:
...
-	pre-commit run --all-files --show-diff-on-failure
+	poetry run pre-commit run --all-files --show-diff-on-failure
 
 test:
-	pytest
+	poetry run pytest
 
 test-all:
-	tox --parallel=auto
+	poetry run tox --parallel=auto
```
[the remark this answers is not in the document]
tarkatronic: Whatever config currently exists for tox will work fine for now! I'm trying to limit the scope of this PR as much as possible.

pyproject.toml
tarkatronic: Yes that is true. And I simply hadn't gotten to that just yet. I'm realizing I should have opened this as a draft PR for the time being!

.gitattributes
tarkatronic: This basically makes diffs easier to deal with. It's not absolutely necessary, and I'd be fine removing it, but diffs of this file are essentially useless since it is entirely machine generated.

The main thing this does is to tell git to not even bother trying to show a diff.

pyproject.toml
tarkatronic: You know, I asked myself the very same question when I was copying this over from the original PR I had 😆 

I know I had a reason for them previously, but I couldn't tell you what that is now. My guess would be, something didn't work right without them. But I can do some experimentation to figure that out.

tarkatronic:
After looking this over and reading the comments so far, I realized it makes the most sense to convert this to a draft.

I'm also thinking that it might be prudent to make some other changes before implementing this. I'm happy to take feedback on these proposed changes.

1. Use `importlib.metadata` to remove the need for the version in `sceptre/__init__.py` (which is a pattern that's known to have potential gotchas)
2. Switch from CircleCI to GitHub Actions. This is something that was discussed previously and people seemed to be in favor of it way back when. This isn't an absolutely necessary pre-requisite, but it would make this transition _massively_ easier and prevent having to get this setup working in one environment, only to have to then switch it to the new.
3. I'm certain there's more but nothing is occurring off the top of my head...

.gitattributes
jfalkenstein: I get that... but then the counter-argument is that if somebody updates dependencies but forgets to update the lock file, I want to see on the PR that they haven't modified the lock file. Perhaps we can add a pre-commit check that the lock file is unchanged after an install or something? In cases with lock files, I don't actually care what the changes _are_, only that there _is_ a change when someone has modified dependencies.

jfalkenstein:
> Use importlib.metadata to remove the need for the version in sceptre/__init__.py (which is a pattern that's known to have potential gotchas)

Unless there's a backport to 3.7 we can install, we cannot use that. That library was introduced in 3.8. It might be installable as a backport, in which case I'm fine doing that. But if we can't, then we shouldn't do that until 3.7 reaches end of life.

> Switch from CircleCI to GitHub Actions. This is something that was discussed previously and people seemed to be in favor of it way back when. This isn't an absolutely necessary pre-requisite, but it would make this transition massively easier and prevent having to get this setup working in one environment, only to have to then switch it to the new.

Regarding GH Actions vs. circleci... I don't really have a dog in that fight. I have no real experience in either. I don't care that much... though I don't see why a transition to Poetry would need to require a transition of CI/CD tool at the same time.


I did have one other question: When Poetry creates a lock-file... that's just used for local development, right? It doesn't actually lock dependencies for those that install Sceptre as a library, right? There needs to be some "wiggle room" to install Sceptre into a virtual environment with other packages without entering a dependency hellscape.

tarkatronic:
> > Use importlib.metadata to remove the need for the version in sceptre/**init**.py (which is a pattern that's known to have potential gotchas)
> 
> Unless there's a backport to 3.7 we can install, we cannot use that. That library was introduced in 3.8. It might be installable as a backport, in which case I'm fine doing that. But if we can't, then we shouldn't do that until 3.7 reaches end of life.

As a matter of fact, there is! https://pypi.org/project/importlib-metadata/

> 
> > Switch from CircleCI to GitHub Actions. This is something that was discussed previously and people seemed to be in favor of it way back when. This isn't an absolutely necessary pre-requisite, but it would make this transition massively easier and prevent having to get this setup working in one environment, only to have to then switch it to the new.
> 
> Regarding GH Actions vs. circleci... I don't really have a dog in that fight. I have no real experience in either. I don't care that much... though I don't see why a transition to Poetry would need to require a transition of CI/CD tool at the same time.
> 

The main reasoning is that CircleCI, especially with Sceptre's current environment, is _very hard_ to get right. You can see my [previous attempt](https://github.com/Sceptre/sceptre/pull/1035/files) for some idea on just how much effort that takes.

In contrast, a lot of what is being done manually with CircleCI just sorta "works" with GitHub Actions. For an example of a Poetry-based project that I set up a while back, see here: https://github.com/godaddy/tartufo/blob/main/.github/workflows/ci.yml

And IIRC since then, GitHub Actions has even added some Poetry-specific setup into their `actions/setup-python` workflow, which makes it even easier!

So the tl;dr is, CircleCI is hard, GitHub Actions is easy, and if the transition is going to happen at some point I think it makes sense to do before Poetry, to avoid even more re-work.

> I did have one other question: When Poetry creates a lock-file... that's just used for local development, right? It doesn't actually lock dependencies for those that install Sceptre as a library, right? There needs to be some "wiggle room" to install Sceptre into a virtual environment with other packages without entering a dependency hellscape.

The lock file is used for local development, and not for end-user installations, yes, that is correct. And there is no reason at all that wiggle room can't be present with Poetry! In fact you may notice, I had to do a bit of massaging of allowed versions already. For example: https://github.com/Sceptre/sceptre/pull/1239/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R60

Poetry has got all kinds of ability to [specify those dependencies](https://python-poetry.org/docs/dependency-specification/) with whatever amount of wiggle room is necessary.

zaro0508:
closing this PR in favor of #1323

PR #1240, opened 2022-06-10
Update sceptre to use importlib

zaro0508:
The purpose of this change is to get us to a point where we can add
support for python 3.10.  It partially resolves issue #1225 

* pkg_resources does not work with python 3.10 so need to switch
to importlib.

* importlib entry_points requires keyword arguments in python 3.10

### DIFF
--- a/sceptre/config/reader.py
+++ b/sceptre/config/reader.py
@@ -176,7 +176,7 @@ class ConfigReader(object):
             return class_constructor
 
         for group in entry_point_groups:
-            if sys.version_info > (3,9):
+            if sys.version_info > (3, 9):
                 entry_points = iter_entry_points(group=group)
             else:
                 entry_points = iter_entry_points(group)
--- a/sceptre/config/reader.py
+++ b/sceptre/config/reader.py
@@ -138,8 +138,8 @@ class ConfigReader(object):
 
     def iterate_entry_points(group):
         """
-        Helper to determine whether to use pkg_resources or
-        importlib.metadata.
+        Helper to determine whether to use pkg_resources or importlib.metadata.
+        https://docs.python.org/3/library/importlib.metadata.html
         """
         if sys.version_info < (3, 10):
             from pkg_resources import iter_entry_points
--- a/sceptre/config/reader.py
+++ b/sceptre/config/reader.py
@@ -136,7 +136,7 @@ class ConfigReader(object):
 
         self.templating_vars = {"var": self.context.user_variables}
 
-    def iterate_entry_points(group):
+    def _iterate_entry_points(group):
         """
         Helper to determine whether to use pkg_resources or importlib.metadata.
         https://docs.python.org/3/library/importlib.metadata.html
@@ -183,7 +183,7 @@ class ConfigReader(object):
 
         for group in entry_point_groups:
 
-            for entry_point in iterate_entry_points(group):
+            for entry_point in self._iterate_entry_points(group):
                 # Retrieve name and class from entry point
                 node_tag = u'!' + entry_point.name
                 node_class = entry_point.load()
--- a/sceptre/config/reader.py
+++ b/sceptre/config/reader.py
@@ -15,7 +15,11 @@ import logging
 from os import environ, path, walk
 from typing import Set
 
-from pkg_resources import iter_entry_points
+try:
+    from importlib.metadata import entry_points as iter_entry_points
+except:
+    from pkg_resources import iter_entry_points
+
 from pathlib import Path
 import yaml
 
@@ -171,7 +175,7 @@ class ConfigReader(object):
             return class_constructor
 
         for group in entry_point_groups:
-            for entry_point in iter_entry_points(group):
+            for entry_point in iter_entry_points(group=group):
                 # Retrieve name and class from entry point
                 node_tag = u'!' + entry_point.name
                 node_class = entry_point.load()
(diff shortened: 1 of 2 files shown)


sceptre/config/reader.py
```diff
@@ -12,10 +12,15 @@
...
 import logging
+import sys
 from os import environ, path, walk
 from typing import Set
 
-from pkg_resources import iter_entry_points
+if sys.version_info < (3, 10):
+    from pkg_resources import iter_entry_points
+else:
+    from importlib.metadata import entry_points as iter_entry_points
```
jfalkenstein, line 22: tricky...

sceptre/template.py
```diff
@@ -246,7 +251,12 @@ def _get_handler_of_type(self, type):
         if not self._registry:
             self._registry = {}
 
-            for entry_point in iter_entry_points("sceptre.template_handlers", type):
+            if sys.version_info < (3, 10):
+                entry_points = iter_entry_points("sceptre.template_handlers", type)
+            else:
+                entry_points = iter_entry_points(group="sceptre.template_handlers", name=type)
```
jfalkenstein, line 257: Yeah, all this duplication could be dealt with if you defined the iterate_entry_points function I suggested above.

jfalkenstein (changes requested):
I think we could avoid the duplicate conditional logic and simply things a bit.

sceptre/config/reader.py
```diff
@@ -171,7 +176,12 @@ def class_constructor(loader, node):
             return class_constructor
 
         for group in entry_point_groups:
-            for entry_point in iter_entry_points(group):
+            if sys.version_info < (3, 10):
+                entry_points = iter_entry_points(group)
+            else:
+                entry_points = iter_entry_points(group=group)
```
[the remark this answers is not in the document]
zaro0508: good idea!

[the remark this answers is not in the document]
zaro0508: done

sceptre/template.py
zaro0508: done

sceptre/template.py
```diff
@@ -234,6 +234,19 @@ def _bucket_region(self, bucket_name):
...
+        Helper to determine whether to use pkg_resources or importlib.metadata.
+        https://docs.python.org/3/library/importlib.metadata.html
+        """
+        if sys.version_info < (3, 10):
+            from pkg_resources import iter_entry_points
+            return iter_entry_points(group, name)
+        else:
+            from importlib.metadata import entry_points
+            return entry_points(group=group, name=name)
+
```
jfalkenstein, line 249: The duplication here still makes me a bit sad, though I can see how that name parameter makes it a bit different.

sceptre/template.py
```diff
@@ -234,6 +234,19 @@ def _bucket_region(self, bucket_name):
     def _domain_from_region(region):
         return "com.cn" if region.startswith("cn-") else "com"
 
+    @staticmethod
+    def _iterate_entry_points(group, name):
+        """
+        Helper to determine whether to use pkg_resources or importlib.metadata.
+        https://docs.python.org/3/library/importlib.metadata.html
+        """
+        if sys.version_info < (3, 10):
```
jfalkenstein, line 243: I'm wondering if we should use importlib for 3.8 and above rather than just 3.10, since importlib was added in 3.8. Also, @tarkatronic just found that importlib was backported for earlier versions. We might be able to dispense with the pkg_resources import altogether if we utilized that.

zaro0508: I'm not opposed to your suggestions however I think it's more risky.  I thought it might be safer going this route because we've been running and testing with `pkg_resources` for py37, py38, and py39 for a while now and it's been working.   The idea with changing it for py310 is that it's required and less risky.

sceptre/template.py
zaro0508: i'm open to suggestions.

sceptre/template.py
tarkatronic: I don't think it's inherently any more risky to make this change, given that you've already made this a discrete change, separate from the Python 3.10 support.

This would just become a more holistic change-over from `pkg_resources` to `importlib`.

jfalkenstein: I honestly don't feel all that strongly either way. I see your point about necessity... though if it works for 3.10, I don't see why it wouldn't work for 3.8+. Once 3.7 hits end of life next year, there won't really be a reason to use pkg_resources in the first place.

zaro0508: I tried changing it to use importlib for py38 and higher however I could not get the tests to pass.  It's a bit confusing because the entry_points function seems to work differently in py38 & py39 than in py310.  The other odd thing is that the entry_point function for py38 & py39 does not take any parameters and I can't seem to figure out how to set group and name properties for that class.

This is what I tried..

```
        if sys.version_info < (3, 8):
            from pkg_resources import iter_entry_points
            return iter_entry_points(group, name)
        elif sys.version_info > (3, 10):
            from importlib.metadata import entry_points
            return entry_points(group=group)
        else:
            from importlib.metadata import entry_points
            return entry_points(group, name)
```

however I got this error for py39 ..
```
        else:
            from importlib.metadata import entry_points
>           return entry_points(group, name)
E           TypeError: entry_points() takes 0 positional arguments but 2 were given
```

jfalkenstein: huh. Ok. Yeah, if it doesn't work, it doesn't work. Carry on!

jfalkenstein (approved):

sceptre/template.py
zaro0508: ok, i've already reverted to 5e08bf0 which was the working version

PR #1242, opened 2022-06-12
use circleci cache to speed up builds

zaro0508:
Setup the circleci cache and workspace settings to speed
up sceptre build and tests pipelines.

* reduce the build pipeline time from ~5 min to 1 min
* reduce the lint-and-unit-tests pipeline from ~6 min to ~3 mins

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,6 +65,7 @@ jobs:
           root: /home/circleci
           paths:
             - project
+            - .pyenv
 
   lint-and-unit-tests:
     docker:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -39,19 +39,23 @@ jobs:
           name: 'Creating Virtualenv'
           command: |
             pyenv virtualenv 3.9.4 venv
+      - restore_cache:
+          key:
+            sceptre-{{ .Environment.CACHE_VERSION }}-dependencies-{{ arch }}-{{
+            checksum "requirements/prod.txt" }}-{{ checksum
+            "requirements/dev.txt" }}
```
jfalkenstein, line 46: This spans multiple lines weird... You sure this is valid yaml?

zaro0508: It's been like this before I ever even touched this file.  I'm just going with the flow.  And yes, it works.

PR #1246, opened 2022-07-18
[Resolve #1224] Add --drifted option to drift show command

### LINKED ISSUE #1224: It'd be great if `drift show` had an option to filter out IN_SYNC resources.
Right now, when `drift show` is run, it shows the drift status of all the resources in all stacks. However, this is a bit overload if you're really just trying to find drifted resources. It would be a great addition if we added a `--filter` flag that would only show resources that are not IN_SYNC.

alexharv074:
Adds a feature --drifted to drift show to filter out IN_SYNC
resources.

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/cli/drift.py
+++ b/sceptre/cli/drift.py
@@ -68,9 +68,12 @@ def drift_detect(ctx: Context, path: str):
 
 @drift_group.command(name="show", short_help="Shows stack drift on running stacks.")
 @click.argument("path")
+@click.option(
+    "-D", "--drifted", is_flag=True, default=False, help="Filter out in sync resources."
+)
 @click.pass_context
 @catch_exceptions
-def drift_show(ctx, path):
+def drift_show(ctx, path, drifted):
     """
     Show stack drift on deployed stacks.
 
@@ -92,7 +95,7 @@ def drift_show(ctx, path):
     )
 
     plan = SceptrePlan(context)
-    responses = plan.drift_show()
+    responses = plan.drift_show(drifted)
 
     output_format = "json" if context.output_format == "json" else "yaml"
 
@@ -100,7 +103,6 @@ def drift_show(ctx, path):
     for stack, (status, response) in responses.items():
         if status in BAD_STATUSES:
             exit_status += 1
-        response.pop("ResponseMetadata", None)
         write({stack.external_name: deserialize_json_properties(response)}, output_format)
 
     exit(exit_status)
--- a/sceptre/plan/actions.py
+++ b/sceptre/plan/actions.py
@@ -1039,7 +1039,7 @@ class StackActions(object):
         return response
 
     @add_stack_hooks
-    def drift_show(self, drifted: bool = True) -> Tuple[str, dict]:
+    def drift_show(self, drifted: bool = False) -> Tuple[str, dict]:
         """
         Detect drift status on stacks.
 
--- a/sceptre/plan/actions.py
+++ b/sceptre/plan/actions.py
@@ -1039,10 +1039,11 @@ class StackActions(object):
         return response
 
     @add_stack_hooks
-    def drift_show(self) -> Tuple[str, dict]:
+    def drift_show(self, drifted: bool = True) -> Tuple[str, dict]:
         """
         Detect drift status on stacks.
 
+        :param drifted: Filter out IN_SYNC resources.
         :returns: The detection status and resource drifts.
         """
         response = self.drift_detect()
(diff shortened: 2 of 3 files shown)


alexharv074:
Updated output would be something like this:

Screen Shot 2022-07-19 at 6 12 48 am

sceptre/cli/drift.py
```diff
@@ -68,6 +68,7 @@ def drift_detect(ctx: Context, path: str):
 
 @drift_group.command(name="show", short_help="Shows stack drift on running stacks.")
 @click.argument("path")
+@click.option("--out-filter", is_flag=True, help="Filter out in sync resources.")
```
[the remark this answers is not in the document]
alexharv074: @jfalkenstein  Yes, that name was only a placeholder until I had a better suggestion, such as `--drifted-only`. Is the rest of it ok?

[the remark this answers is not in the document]
jfalkenstein: The more I think about it, I think I like `--drifted` better than `--drifted-only` as an option name.

I think your general approach is good here, though. It's a very useful option, I think.

sceptre/cli/drift.py
```diff
@@ -68,9 +68,12 @@ def drift_detect(ctx: Context, path: str):
 
 @drift_group.command(name="show", short_help="Shows stack drift on running stacks.")
 @click.argument("path")
+@click.option(
+    "-D", "--drifted", is_flag=True, default=False, help="Filter out in sync resources."
```
jfalkenstein, line 72: Why capital D here? Why not just `-d`?

sceptre/cli/drift.py
```diff
@@ -92,15 +95,14 @@ def drift_show(ctx, path):
...
-    responses = plan.drift_show()
+    responses = plan.drift_show(drifted)
 
     output_format = "json" if context.output_format == "json" else "yaml"
 
     exit_status = 0
     for stack, (status, response) in responses.items():
         if status in BAD_STATUSES:
             exit_status += 1
-        response.pop("ResponseMetadata", None)
```
jfalkenstein, line 103: Why are you removing this?

sceptre/plan/actions.py
```diff
@@ -1055,6 +1056,7 @@ def drift_show(self) -> Tuple[str, dict]:
         else:
             raise Exception("Not expected to be reachable")
 
+        response = self._filter_drifts(response, drifted)
```
jfalkenstein, line 1059: Why not only apply this when `drifted`?

```suggestion
        if drifted:
            response = self._filter_drifts(response, drifted)
```

sceptre/plan/actions.py
```diff
@@ -1144,6 +1146,19 @@ def _describe_stack_resource_drifts(self) -> dict:
...
 
+    def _filter_drifts(self, response: dict, show_only_drifted: bool) -> dict:
+        """
+        The filtered response after filtering out StackResourceDriftStatus.
+        :param show_only_drifted: Filter out IN_SYNC resources.
+        """
+        if "StackResourceDrifts" in response:
+            return {"StackResourceDrifts": [
+                drift for drift in response["StackResourceDrifts"]
+                if not show_only_drifted or drift["StackResourceDriftStatus"] != "IN_SYNC"
```
jfalkenstein, line 1157: I think this condition is overly complicated. If you just only filtered when `drifted` is true, you could simplify this. Then you can remove the boolean flag and make this function have less complexity.

sceptre/cli/drift.py
alexharv074: I would have expected `-d` to have been taken by `--debug` although I see we are not using short options at all. Perhaps it should just be `--drifted` with no short option?

sceptre/cli/drift.py
alexharv074: @jfalkenstein Because I had this idea that perhaps all of this 'filtering' should occur inside a single, overcomplicated filtering function!

sceptre/plan/actions.py
jfalkenstein, line 1157: maybe this?
```suggestion
        include_all_drift_statuses = not show_only_drifted
        if "StackResourceDrifts" in response:
            return {"StackResourceDrifts": [
                drift for drift in response["StackResourceDrifts"]
                if include_all_drift_statuses or drift["StackResourceDriftStatus"] != "IN_SYNC"
```

jfalkenstein (approved):
I approve of this, but we need to figure out the test thing before we merge this.

jfalkenstein (changes requested):
Looks like you have some integration test failures around drift detection.

jfalkenstein (approved):

PR #1270, opened 2022-11-28
[Resolves #1268] Modify ConfigFileNotFound Error to Conditionally Include Valid Stack Paths

### LINKED ISSUE #1268: When a non-existent Stack is Launched, the List of Valid Stack Paths is not useful in a Large Workspace
### Subject of the issue

If a non-existent stack is launched, Sceptre tries to be helpful in its error message by outputting a list of valid stack paths. In a large workspace (we have nearly 5000 stacks), this is not useful.

### Your environment

* version of sceptre `3.2.0`
* version of python `3.8.8`
* which OS/distro: `Windows 10`

### Steps to reproduce

Run a sceptre command (launch, generate etc) on a stack name that doesn't exist. i.e. `sceptre launch dev/foo2.yaml`

### Expected behaviour

Annotate the error message with the valid stack paths only when there are a limited number of stacks in the workspace (<10?)

### Actual behaviour

The following error is produced, listing all stack paths in the workspace.
```
"No stacks detected from the given path dev/foo2.yaml. Valid stack paths are: ['dev/foo.yaml', 'dev/bar.yaml']"
```

### Solution

I am happy to raise a PR to improve this error handling.

X-Guardian:
Modifies the Sceptre Plan `ConfigFileNotFound` error to only include the list of valid stack paths from the command path specified and only there are less than 10 stacks.

## Example

```
sceptre launch dev/service1/foo.yaml
```

### New Error Message (when there are less than 10 stacks under the specified parent path)

```
"No stacks detected from the given path 'dev/service1/foo.yaml'. Valid stack paths under 'dev/service1' are: ['dev/service1/foo1.yaml', 'dev/service1/foo2.yaml', 'dev/service1/subservice1/foo3.yaml']"
```

### New Error Message (when there are 10 stacks or more under the specified parent path)

```
"No stacks detected from the given path 'dev/service1/foo.yaml'."
```

- Resolves #1268

## PR Checklist

## Other Information


sceptre/plan/plan.py
```diff
@@ -73,10 +73,10 @@ def _generate_launch_order(self, reverse=False) -> List[Set[Stack]]:
                 graph.remove_stack(stack)
 
         if not launch_order:
-            raise ConfigFileNotFoundError(
-                "No stacks detected from the given path '{}'. Valid stack paths are: {}"
-                .format(sceptreise_path(self.context.command_path), self._valid_stack_paths())
-            )
+            error_text = f'No stacks detected from the given path {sceptreise_path(self.context.command_path)}.'
+            if len(self._valid_stack_paths()) < 10:
```
jfalkenstein, line 77: Let's use a well-named constant here rather than a magic number.

jfalkenstein: I also think we ought to filter the valid stack paths based upon my suggestion here: https://github.com/Sceptre/sceptre/issues/1268#issuecomment-1329332100

X-Guardian: Constant set and refactored to filter the valid stack paths.

sceptre/plan/plan.py
```diff
@@ -55,6 +56,25 @@ def _execute(self, *args):
...
+        command_path = pathlib.PurePath(self.context.command_path)
+        filtered_valid_stack_paths = []
+        for stack_path in self._valid_stack_paths():
+            for parent in pathlib.PurePath(stack_path).parents:
+                if command_path.parent == parent:
+                    filtered_valid_stack_paths.append(stack_path)
+                    break
+            if filtered_valid_stack_paths and len(filtered_valid_stack_paths) > MAX_VALID_STACK_PATH_COUNT:
+                filtered_valid_stack_paths = None
+                break
```
jfalkenstein, line 72: A few thoughts on this:
* This seems rather deeply nested. I'd like to avoid getting so nested, especially in this function. It hurts readability. I also don't think we need to do it this way.
* Rather than crawling every combination of segments and comparing parents, I don't think we need to get that intense about it. Rather, I think a simpler and more straight-forward logic here would be to do this:
    1. Check if there's a "/" in the command path. If not, then the prefix to use is `''` (all strings technically start with `''`). If there IS, do `prefix, _ = command_path.rsplit('/' 1)` to get the prefix. 
    2. Filter the valid paths for those that start with the prefix. You can either use `filter()` or a list comprehension. 

Done this way, it would look more like:

```suggestion
        if '/' in self.context.command_path:
            prefix, _ = self.context.command_path.rsplit('/', 1)
        else:
            prefix = ''
        all_paths = self._valid_stack_paths() 
        # We'll fall back to all_paths if no valid path starts with the prefix
        filtered_valid_stack_paths = [p for p in all_paths if p.startswith(prefix + '/')] or all_paths
        if len(filtered_valid_stack_paths) > MAX_VALID_STACK_PATH_COUNT:
            filtered_valid_stack_paths = []
```

jfalkenstein (dismissed):
I'd also like to see a test around this functionality.

sceptre/plan/plan.py
```diff
@@ -55,6 +56,25 @@ def _execute(self, *args):
         executor = SceptrePlanExecutor(self.command, self.launch_order)
         return executor.execute(*args)
 
+    def _generate_no_launch_order_error(self):
```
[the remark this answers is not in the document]
X-Guardian: Done

sceptre/plan/plan.py
X-Guardian: Refactored function to use this suggestion with `pathlib` rather than `/` processing.

jfalkenstein:
@X-Guardian, what are the status on the tests for this PR?

alexharv074:
@X-Guardian Do you have any plans to finish this PR?

X-Guardian:
From what I remember it works, but I don't use Sceptre any more, so won't be writing any tests, which is what I think was blocking this getting merged.

alexharv074:
@zaro0508  I have had a look at this stale PR and I agree with @X-Guardian  and I think the PR should be merged as-is. The new private method is simply tweaking the error message seen and I do not think that such a trivial change justifies us adding more unit test code. Nor is it obvious to me how you would test this anyway.

Could we merge? I've added it into my integration branch and I'm certainly happy with it.

zaro0508 (approved):

PR #1275, opened 2022-12-11
[Resolve #1273]: Events start from response time

### LINKED ISSUE #1273: Events missing from launch output
### Events missing from launch output
At some point after updating to Sceptre 3 events which appear in the web console are not included in the command output. It seems like if the stack update lasts long enough events will start to appear as they used to. If the stack launch is short then the command simply exits without feedback. This includes launch failures and roll-backs.

### Your environment
* Sceptre, version 3.2.0
* Python 3.9.9
* Official Python 3.9.9 Docker image

### Steps to reproduce
```$ sceptre launch -y <stack>```

This has been consistent across several accounts and versions of Sceptre. I've not yet identified which version the problem appears in.

### Expected behaviour
After the `Updating Stack` message we expect to see a number of resource state-change events and end with the final state of the stack.

### Actual behaviour
The process completes successfully, but events which appear in the web console are not output at the terminal:

```
$ sceptre launch -y prod/alb-internal.yaml
[2022-12-11 15:37:23] - prod/alb-internal - Launching Stack
[2022-12-11 15:37:24] - prod/alb-internal - Stack is in the UPDATE_COMPLETE state
[2022-12-11 15:37:24] - prod/alb-internal - Updating Stack
$
```

### PR DESCRIPTION
**Pull request by dboitnot**
Resolves #1273 by starting event filtering from the timestamp returned in the AWS response headers rather than relying on the workstation clock.

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/helpers.py
+++ b/sceptre/helpers.py
@@ -1,8 +1,10 @@
 # -*- coding: utf-8 -*-
-import dateutil.parser
-
 from contextlib import contextmanager
+from datetime import datetime
 from os import sep
+from typing import Optional
+
+import dateutil.parser
 
 from sceptre.exceptions import PathConversionError
 
@@ -121,11 +123,10 @@ def null_context():
     yield
 
 
-def extract_datetime_from_aws_response_headers(boto_response):
+def extract_datetime_from_aws_response_headers(boto_response: dict) -> Optional[datetime]:
     """Returns a datetime.datetime extracted from the response metadata in a
     boto response or None if it's unable to find or parse one.
     :param boto_response: A dictionary returned from a boto client call
-    :type boto_response: dict
     :returns a datetime.datetime or None
     """
     if boto_response is None:
@@ -133,7 +134,7 @@ def extract_datetime_from_aws_response_headers(boto_response):
     try:
         return dateutil.parser.parse(boto_response["ResponseMetadata"]["HTTPHeaders"]["date"])
     except (KeyError, dateutil.parser.ParserError):
-        # We expect a KeyError if the date isn't present in the response. We expect
-        # a ParserError if it's not well-formatted. Any other error we want to pass
-        # along.
+        # We expect a KeyError if the date isn't present in the response. We
+        # expect a ParserError if it's not well-formed. Any other error we want
+        # to pass along.
         return None
--- a/sceptre/helpers.py
+++ b/sceptre/helpers.py
@@ -1,5 +1,8 @@
 # -*- coding: utf-8 -*-
+import email.utils as eut
+
 from contextlib import contextmanager
+from datetime import datetime
 from os import sep
 
 from sceptre.exceptions import PathConversionError
@@ -117,3 +120,10 @@ def null_context():
     available in py3.6, so providing it here instead.
     """
     yield
+
+
+def get_response_datetime(resp):
+    try:
+        return datetime(*eut.parsedate(resp["ResponseMetadata"]["HTTPHeaders"]["date"])[:6])
+    except (TypeError, KeyError):
+        return None
--- a/sceptre/helpers.py
+++ b/sceptre/helpers.py
@@ -1,8 +1,7 @@
 # -*- coding: utf-8 -*-
-import email.utils as eut
+import dateutil.parser
 
 from contextlib import contextmanager
-from datetime import datetime
 from os import sep
 
 from sceptre.exceptions import PathConversionError
@@ -129,10 +128,12 @@ def extract_datetime_from_aws_response_headers(boto_response):
     :type boto_response: dict
     :returns a datetime.datetime or None
     """
+    if boto_response is None:
+        return None
     try:
-        # email.utils.parsedate returns a 9-tuple that can be passed to
-        # time.mktime. Fields 6 (tm_wday), 7 (tm_yday), and 8 (tm_isdst) are
-        # unneeded so we slice them off when passing to datetime().
-        return datetime(*eut.parsedate(boto_response["ResponseMetadata"]["HTTPHeaders"]["date"])[:6])
-    except (TypeError, KeyError):
+        return dateutil.parser.parse(boto_response["ResponseMetadata"]["HTTPHeaders"]["date"])
+    except (KeyError, dateutil.parser.ParserError):
+        # We expect a KeyError if the date isn't present in the response. We expect
+        # a ParserError if it's not well-formatted. Any other error we want to pass
+        # along.
         return None
--- a/sceptre/helpers.py
+++ b/sceptre/helpers.py
@@ -122,8 +122,17 @@ def null_context():
     yield
 
 
-def get_response_datetime(resp):
+def extract_datetime_from_aws_response_headers(boto_response):
+    """Returns a datetime.datetime extracted from the response metadata in a
+    boto response or None if it's unable to find or parse one.
+    :param boto_response: A dictionary returned from a boto client call
+    :type boto_response: dict
+    :returns a datetime.datetime or None
+    """
     try:
-        return datetime(*eut.parsedate(resp["ResponseMetadata"]["HTTPHeaders"]["date"])[:6])
+        # email.utils.parsedate returns a 9-tuple that can be passed to
+        # time.mktime. Fields 6 (tm_wday), 7 (tm_yday), and 8 (tm_isdst) are
+        # unneeded so we slice them off when passing to datetime().
+        return datetime(*eut.parsedate(boto_response["ResponseMetadata"]["HTTPHeaders"]["date"])[:6])
     except (TypeError, KeyError):
         return None
--- a/sceptre/plan/actions.py
+++ b/sceptre/plan/actions.py
@@ -50,7 +50,6 @@ class StackActions(object):
         self.stack = stack
         self.name = self.stack.name
         self.logger = logging.getLogger(__name__)
-        self.most_recent_event_datetime = None
         self.connection_manager = ConnectionManager(
             self.stack.region, self.stack.profile,
             self.stack.external_name, self.stack.iam_role,
@@ -96,8 +95,7 @@ class StackActions(object):
                 "%s - Create stack response: %s", self.stack.name, response
             )
 
-            self._note_response_time(response)
-            status = self._wait_for_completion()
+            status = self._wait_for_completion(boto_response=response)
         except botocore.exceptions.ClientError as exp:
             if exp.response["Error"]["Code"] == "AlreadyExistsException":
                 self.logger.info(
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by dboitnot** [comment]
I'm not sure I understand yet what's going on with that unit test. As far as I know I haven't messed with the timeout behavior.

**Comment by jfalkenstein** [comment]
Ugh... that test that's failing is exceptionally ugly and patches things in the worst way. Feel free to rewrite it to function and read better.

File: sceptre/helpers.py [hunk 1]
```diff
@@ -117,3 +120,10 @@ def null_context():
     available in py3.6, so providing it here instead.
     """
     yield
+
+
+def get_response_datetime(resp):
+    try:
+        return datetime(*eut.parsedate(resp["ResponseMetadata"]["HTTPHeaders"]["date"])[:6])
```
jfalkenstein, line 127: I think a comment here about why you're only selecting the first 6 characters would be appropriate. What's an example of the format this header is in?

sceptre/helpers.py
```diff
@@ -117,3 +120,10 @@ def null_context():
     available in py3.6, so providing it here instead.
     """
     yield
+
+
+def get_response_datetime(resp):
```
jfalkenstein, line 125: I think this function name and parameter name could definitely use improvement. How about something like:
`extract_datetime_from_aws_response_headers(boto_response)`?

Also, a docstring should definitely be here.

sceptre/plan/actions.py
```diff
@@ -95,6 +96,7 @@ def create(self):
                 "%s - Create stack response: %s", self.stack.name, response
             )
 
+            self._note_response_time(response)
```
jfalkenstein, line 99: Rather than this being on this class... I'm thinking it would make sense for this to be on the ConnectionManager itself. Basically, it records the timestamp of each response and makes it available as a property on it.

sceptre/helpers.py
jfalkenstein: Also type annotations would be great.

sceptre/plan/actions.py
```diff
@@ -767,6 +773,9 @@ def _protect_execution(self):
                 "currently enabled".format(self.stack.name)
             )
 
+    def _note_response_time(self, resp):
+        self.most_recent_event_datetime = get_response_datetime(resp)
```
jfalkenstein, line 777: Technically `get_response_datetime` could return None, in which case, I don't think you want to override the existing value. 
```suggestion
        self.most_recent_event_datetime = get_response_datetime(resp) or self.most_recent_event_datetime
```

sceptre/plan/actions.py
```diff
@@ -784,9 +793,11 @@ def timed_out(elapsed):
 
         status = StackStatus.IN_PROGRESS
 
-        self.most_recent_event_datetime = (
-            datetime.now(tzutc()) - timedelta(seconds=3)
-        )
```
jfalkenstein, line 789: I'm confused here. It looks like you added this property to this class... but then this diff seems to indicate it was already here... were we adding it where it wasn't declared in `__init__` before?

sceptre/helpers.py
dboitnot: Fair enough. Can do.

sceptre/helpers.py
dboitnot: I like it. Will do.

sceptre/plan/actions.py
dboitnot: This, actually, was not my first choice. The class, as written, stores this value as an instance attribute:

https://github.com/Sceptre/sceptre/blob/b73af2f341ff3b7b9ca4192858b69732fb26f195/sceptre/plan/actions.py#L787-L789

This value is "passed" to `_log_new_events` via this instance property. My first approach was to update the signature of `_wait_for_completion`:

```python
    def _wait_for_completion(self, timeout=0, start_time=None):
        # ...
        self.most_recent_event_datetime = start_time or datetime.now(tzutc()) - timedelta(seconds=3)
 ```

This broke several unit tests whose mocks weren't expecting the added parameter in the method calls. What I'm doing here resolves those but I'm not super happy with it either. If we're free to re-work things a bit, I'd argue that rather than an instance attribute, `most_recent_event_datetime` should be a stack variable in `_wait_for_completion` which is passed to `_log_new_events`.

`_log_new_events` would then return either:

- The datetime of the last event; or
- The datetime it was passed if there were no events.

This way we can eliminate some instance state which might cause issues down the line. We'll still need to update the signature for `_wait_for_completion`. Given this is an internal method and always invoked after a boto call, how would y'all feel about just passing the boto response?

```python
    def _wait_for_completion(self, timeout=0, boto_response=None):
```

sceptre/plan/actions.py
dboitnot: I'm resetting it to None here so that if/when `_log_new_events` is called it will compute a default based on the system clock. But, as I mentioned above, I think there's a better way where we can eliminate this instance attribute in favor of a stack variable.

sceptre/plan/actions.py
dboitnot: That's correct. This value wasn't being initialized previously in __init__.

tests/test_helpers.py
```diff
@@ -67,3 +69,25 @@ def test_sceptreise_path_with_trailing_backslash(self):
...
+                "HTTPHeaders": {"date": "garbage"}
+            }
+        }
+        assert get_response_datetime(resp) is None
+
+    def test_get_response_datetime_missing(self):
+        assert get_response_datetime({}) is None
+
+    def test_get_response_datetime_null(self):
+        assert get_response_datetime(None) is None
```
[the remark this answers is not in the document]
dboitnot: Love it. Will do.

sceptre/plan/actions.py
jfalkenstein: I think refactoring these private methods to be easier to test/more maintainable and overall reducing instance state sounds great to me and I'd definitely support that. As far as a stack variable... I don't think that's the appropriate place for them, though. If it's not held via the connection manager, passing parameters makes sense. Alternatively, using generators to maintain state in a really self-contained way can be a handy trick. Not sure if it'd apply here.

As far as the tests, it's not a secret that our tests need overhaul. We have several open Issues just for test improvement. As far as I'm concerned here, follow the "boyscout principle": Leave the code better than you found it. In this case, if the tests (which I recognize are overly rigid and poorly made) are fighting you, let's refactor/fix those tests to improve the quality of the tests and improve code coverage. You'll get 1000 Jon-points if you accomplish that without using `patch`. Don't go overboard and refactor things that are unrelated here. But for the things that touch the code you're introducing, by all means, be a boyscout.

dboitnot: Sorry, what I meant by "stack variable" was "variable on the stack as opposed to the heap." Possibly not the best term in Python. I just mean, "plain old variable scoped within `_wait_for_completion` and passed as an argument to `_log_new_events`.

My approach in this commit was definitely less "boyscout" and more "touch as little as possible." Let me do some actual refactoring and see if I can earn those Jon-points.

sceptre/helpers.py
```diff
@@ -1,5 +1,8 @@
 # -*- coding: utf-8 -*-
+import email.utils as eut
```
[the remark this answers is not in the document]
dboitnot: Whelp, turns out email.utils was a bad choice anyway since it doesn't include timezone info. I'll switch to dateutil.

dboitnot:
Just pushed commits for review which address all of the above EXCEPT the mock tests in TestStackActions. If you guys are happy with the refactor then I'll see about getting those tests back in order.

sceptre/helpers.py
```diff
@@ -117,3 +119,21 @@ def null_context():
...
     yield
+
+
+def extract_datetime_from_aws_response_headers(boto_response):
+    """Returns a datetime.datetime extracted from the response metadata in a
+    boto response or None if it's unable to find or parse one.
+    :param boto_response: A dictionary returned from a boto client call
+    :type boto_response: dict
+    :returns a datetime.datetime or None
+    """
```
jfalkenstein, line 130: Now that we're in the land of Python 3, it doesn't make any sense to indicate types in docstrings. Just use type annotations! They're actually readable by code and work better for many reasons. 

```suggestion
# you'll need to import Optional and datetime.datetime for these annotations to work
def extract_datetime_from_aws_response_headers(boto_response: dict) -> Optional[datetime]:
    """Returns a datetime.datetime extracted from the response metadata in a
    boto response or None if it's unable to find or parse one.
    :param boto_response: A dictionary returned from a boto client call
    :returns a datetime.datetime or None
    """
```

sceptre/plan/actions.py
```diff
@@ -843,15 +844,20 @@ def _get_simplified_status(status):
                 "{0} is unknown".format(status)
             )
 
-    def _log_new_events(self):
+    def _log_new_events(self, after_datetime):
         """
         Log the latest Stack events while the Stack is being built.
+
+        :param after_datetime: Only events after this datetime will be logged.
+        :type after_datetime: datetime.datetime
```
jfalkenstein, line 852: again, use type annotations, don't indicate types in your docstrings.

sceptre/plan/actions.py
```diff
@@ -843,15 +844,20 @@ def _get_simplified_status(status):
                 "{0} is unknown".format(status)
             )
 
-    def _log_new_events(self):
+    def _log_new_events(self, after_datetime):
```
jfalkenstein, line 847: Oh yeah, this is better. I think passing the parameters rather than using instance variables is WAY better here. Thanks for making things better!

jfalkenstein (commented):
I like what I'm seeing. My biggest criticism is the lack of type annotations. Let's use those as much as is possible. They'll help developers as well as type-checkers prevent bugs in code in the future.

sceptre/helpers.py
dboitnot: I'd wondered about that but went for consistency with the other functions. Will do.

dboitnot:
Just pushed update with type annotations. If you guys are happy with where we're at, I'll dig into the mock tests.

jfalkenstein (commented):
Looks good to me. I'm happy these changes. Let me know when you've got tests resolved.

dboitnot:
Tests passing. Note that I'm using `ANY` as the matcher on the calls to `_wait_for_completion` since a) the arg to that call isn't relevant to the tests, and b) it was unclear (to me) how to assert that the call to one mock was the mock returned by the call to the last mock.

jfalkenstein:
Running integration tests on this now. This generally looks pretty good to me and should be regression tested well enough by our integration tests. Assuming those pass, I'll approve.

PR #1282, opened 2022-12-28
[Resolves #1179] cloudformation disable-rollback option

### LINKED ISSUE #1179: Feature: support disable-rollback flag
AWS cloudformation just introduced the `--disable-rollback` flag to allow quickly retrying stack operations from the point of failure.  It would be great if sceptre could create/update/launch stacks with this flag as well.

More info here:
https://aws.amazon.com/blogs/aws/new-for-aws-cloudformation-quickly-retry-stack-operations-from-the-point-of-failure/

zaro0508:
Allow user to disable a cloudformation rollback on a Sceptre deployment.

### DIFF
--- a/docs/_source/docs/stack_config.rst
+++ b/docs/_source/docs/stack_config.rst
@@ -214,12 +214,13 @@ disable_rollback
 
 This parameter describes the action taken by CloudFormation when a Stack fails
 to create or update, default is False. This option can be set from the stack
-config or from the Sceptre CLI commands to deploy stacks. Enabling disable-rollback
-using the CLI option disables rollback globally for all stacks. This option
-overrides on_failure since Cloudformation does not allow setting both on deployment.
-For more information and valid values see the `AWS Documentation`_.
+config or from the Sceptre CLI commands to deploy stacks. The disable_rollback
+CLI option (i.e. sceptre launch --disable-rollback) disables cloudformation
+rollback globally for all stacks. This option overrides on_failure since
+Cloudformation does not allow setting both on deployment. For more information
+and valid values see the `AWS Documentation`_.
 
-Examples include:
+Examples:
 
 ``disable_rollback: "True"``
 
--- a/docs/_source/docs/stack_config.rst
+++ b/docs/_source/docs/stack_config.rst
@@ -20,6 +20,7 @@ particular Stack. The available keys are listed below.
 -  `notifications`_ *(optional)*
 -  `obsolete`_ *(optional)*
 -  `on_failure`_ *(optional)*
+-  `disable_rollback`_ *(optional)*
 -  `parameters`_ *(optional)*
 -  `protected`_ *(optional)*
 -  `role_arn`_ *(optional)*
@@ -205,6 +206,22 @@ Examples include:
 
 ``on_failure: "DELETE"``
 
+disable_rollback
+~~~~~~~~~~~~~~~~
+* Resolvable: No
+* Can be inherited from StackGroup: Yes
+* Inheritance strategy: Overrides parent if set
+
+This parameter describes the action taken by CloudFormation when a Stack fails
+to create or update, default is False. This option can be set from the stack
+config or from the Sceptre CLI commands to deploy stacks. Enabling disable-rollback
+using the CLI option disables rollback globally for all stacks. This option
+overrides on_failure since Cloudformation does not allow setting both on deployment.
+For more information and valid values see the `AWS Documentation`_.
+
+Examples include:
+
+``disable_rollback: "True"``
 
 parameters
 ~~~~~~~~~~
@@ -579,6 +596,7 @@ Examples
 .. _hooks: #hooks
 .. _notifications: #notifications
 .. _on_failure: #on-failure
+.. _disable_rollback: #disable-rollback
 .. _parameters: #parameters
 .. _protected: #protected
 .. _role_arn: #role-arn
--- a/sceptre/cli/create.py
+++ b/sceptre/cli/create.py
@@ -10,9 +10,15 @@ from sceptre.cli.helpers import stack_status_exit_code
 @click.argument("path")
 @click.argument("change-set-name", required=False)
 @click.option("-y", "--yes", is_flag=True, help="Assume yes to all questions.")
+@click.option(
+    "-dr",
+    "--disable-rollback",
+    is_flag=True,
+    help="Disable the auto rollback and keep resources successfully created or updated",
+)
 @click.pass_context
 @catch_exceptions
-def create_command(ctx, path, change_set_name, yes):
+def create_command(ctx, path, change_set_name, yes, disable_rollback: bool):
     """
     Creates a stack for a given config PATH. Or if CHANGE_SET_NAME is specified
     creates a change set for stack in PATH.
@@ -24,9 +30,12 @@ def create_command(ctx, path, change_set_name, yes):
     :type change_set_name: str
     :param yes: A flag to assume yes to all questions.
     :type yes: bool
+    :param disable_rollback: A flag to disable cloudformation rollback.
+    :type disable_rollback: bool
     """
     context = SceptreContext(
         command_path=path,
+        command_params=ctx.params,
         project_path=ctx.obj.get("project_path"),
         user_variables=ctx.obj.get("user_variables"),
         options=ctx.obj.get("options"),
--- a/sceptre/cli/create.py
+++ b/sceptre/cli/create.py
@@ -14,7 +14,7 @@ from sceptre.cli.helpers import stack_status_exit_code
 @click.option(
     "--disable-rollback/--enable-rollback",
     default=None,
-    help="Disable the auto rollback and keep resources successfully created or updated",
+    help="Disable or enable the cloudformation automatic rollback",
 )
 @click.pass_context
 @catch_exceptions
--- a/sceptre/cli/create.py
+++ b/sceptre/cli/create.py
@@ -1,5 +1,6 @@
 import click
 
+from typing import Optional
 from sceptre.context import SceptreContext
 from sceptre.cli.helpers import catch_exceptions, confirmation
 from sceptre.plan.plan import SceptrePlan
@@ -30,7 +31,6 @@ def create_command(ctx, path, change_set_name, yes, disable_rollback: Optional[b
     :param yes: A flag to assume yes to all questions.
     :type yes: bool
     :param disable_rollback: A flag to disable cloudformation rollback.
-    :type disable_rollback: bool
     """
     context = SceptreContext(
         command_path=path,
--- a/sceptre/cli/create.py
+++ b/sceptre/cli/create.py
@@ -12,9 +12,9 @@ from sceptre.cli.helpers import stack_status_exit_code
 @click.argument("change-set-name", required=False)
 @click.option("-y", "--yes", is_flag=True, help="Assume yes to all questions.")
 @click.option(
-  "--disable-rollback/--enable-rollback",
-  default=None,
-  help="Disable the auto rollback and keep resources successfully created or updated",
+    "--disable-rollback/--enable-rollback",
+    default=None,
+    help="Disable the auto rollback and keep resources successfully created or updated",
 )
 @click.pass_context
 @catch_exceptions
--- a/sceptre/cli/create.py
+++ b/sceptre/cli/create.py
@@ -11,14 +11,13 @@ from sceptre.cli.helpers import stack_status_exit_code
 @click.argument("change-set-name", required=False)
 @click.option("-y", "--yes", is_flag=True, help="Assume yes to all questions.")
 @click.option(
-    "-dr",
-    "--disable-rollback",
-    is_flag=True,
-    help="Disable the auto rollback and keep resources successfully created or updated",
+  "--disable-rollback/--enable-rollback",
+  default=None,
+  help="Disable the auto rollback and keep resources successfully created or updated",
 )
 @click.pass_context
 @catch_exceptions
-def create_command(ctx, path, change_set_name, yes, disable_rollback: bool):
+def create_command(ctx, path, change_set_name, yes, disable_rollback: Optional[bool]):
     """
     Creates a stack for a given config PATH. Or if CHANGE_SET_NAME is specified
     creates a change set for stack in PATH.
--- a/sceptre/cli/delete.py
+++ b/sceptre/cli/delete.py
@@ -30,6 +30,7 @@ def delete_command(ctx, path, change_set_name, yes):
     """
     context = SceptreContext(
         command_path=path,
+        command_params=ctx.params,
         project_path=ctx.obj.get("project_path"),
         user_variables=ctx.obj.get("user_variables"),
         options=ctx.obj.get("options"),
--- a/sceptre/cli/launch.py
+++ b/sceptre/cli/launch.py
@@ -24,9 +24,17 @@ logger = logging.getLogger(__name__)
     is_flag=True,
     help="If set, will delete all stacks in the command path marked as obsolete.",
 )
+@click.option(
+    "-dr",
+    "--disable-rollback",
+    is_flag=True,
+    help="Disable the auto rollback and keep resources successfully created or updated",
+)
 @click.pass_context
 @catch_exceptions
-def launch_command(ctx: Context, path: str, yes: bool, prune: bool):
+def launch_command(
+    ctx: Context, path: str, yes: bool, prune: bool, disable_rollback: bool
+):
     """
     Launch a Stack or StackGroup for a given config PATH. This command is intended as a catch-all
     command that will apply any changes from Stack Configs indicated via the path.
@@ -36,11 +44,13 @@ def launch_command(ctx: Context, path: str, yes: bool, prune: bool):
     * Any stacks that already exist will be updated (if there are any changes)
     * If any stacks are marked with "ignore: True", those stacks will neither be created nor updated
     * If any stacks are marked with "obsolete: True", those stacks will neither be created nor updated.
-      Furthermore, if the "-p"/"--prune" flag is used, these stacks will be deleted prior to any
+    * If the "-p"/"--prune" flag is used, these stacks will be deleted prior to any
       other launch commands
+    * If the "-dr"/"--disable-rollback" flag is used, automatic cloudformation rollback will be disabled
     """
     context = SceptreContext(
         command_path=path,
+        command_params=ctx.params,
         project_path=ctx.obj.get("project_path"),
         user_variables=ctx.obj.get("user_variables"),
         options=ctx.obj.get("options"),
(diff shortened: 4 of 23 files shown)


docs/_source/docs/stack_config.rst
```diff
@@ -205,6 +206,22 @@ Examples include:
...
+This parameter describes the action taken by CloudFormation when a Stack fails
+to create or update, default is False. This option can be set from the stack
+config or from the Sceptre CLI commands to deploy stacks. Enabling disable-rollback
+using the CLI option disables rollback globally for all stacks. This option
+overrides on_failure since Cloudformation does not allow setting both on deployment.
+For more information and valid values see the `AWS Documentation`_.
+
+Examples include:
+
+``disable_rollback: "True"``
```
jfalkenstein, line 224: Why make this a stack config?

sceptre/cli/create.py
```diff
@@ -10,9 +10,15 @@
 @click.argument("path")
 @click.argument("change-set-name", required=False)
 @click.option("-y", "--yes", is_flag=True, help="Assume yes to all questions.")
+@click.option(
+    "-dr",
```
jfalkenstein, line 14: Usually with a CLI command, `-dr` is equivalent to `-d -r`. I don't think we ought to shorten one command to two letters in the short form. Considering disabling the rollback on an action really ought to be an edge case anyway, we can just take out the short form and have the only way to do it via cli as `--disable-rollback`.

sceptre/cli/create.py
```diff
@@ -24,9 +30,12 @@ def create_command(ctx, path, change_set_name, yes):
     :type change_set_name: str
     :param yes: A flag to assume yes to all questions.
     :type yes: bool
+    :param disable_rollback: A flag to disable cloudformation rollback.
+    :type disable_rollback: bool
```
jfalkenstein, line 34: When you're using type annotations, specifying the `:type` isn't necessary and is probably redundant.

docs/_source/docs/stack_config.rst
jfalkenstein: I kind of feel like this should _only_ be a cli flag and not a stack config at all. What's the use case of having this as a stack configuration?

sceptre/context.py
```diff
@@ -71,6 +72,9 @@ def __init__(
 
         self.normal_command_path = normalise_path(command_path)
 
+        # the sceptre command parameters (e.g. sceptre launch <command params>)
+        self.command_params = command_params if command_params else {}
```
jfalkenstein, line 76:
```suggestion
        self.command_params = command_params or {}
```

sceptre/plan/actions.py
```diff
@@ -82,8 +82,12 @@ def create(self):
             ],
         }
 
-        if self.stack.on_failure:
+        # can specify either DisableRollback or OnFailure , but not both
+        if self.stack.disable_rollback:
+            create_stack_kwargs.update({"DisableRollback": self.stack.disable_rollback})
+        elif self.stack.on_failure:
             create_stack_kwargs.update({"OnFailure": self.stack.on_failure})
```
jfalkenstein, line 89: Where are you going to validate that we don't use both?

sceptre/cli/delete.py
```diff
@@ -30,6 +30,7 @@ def delete_command(ctx, path, change_set_name, yes):
     """
     context = SceptreContext(
         command_path=path,
+        command_params=ctx.params,
```
jfalkenstein, line 33: Is disable rollback an option on delete as well?

tests/test_config_reader.py
```diff
@@ -293,6 +300,7 @@ def test_construct_stacks_constructs_stack(
             external_name=None,
             notifications=None,
             on_failure=None,
+            disable_rollback=False,
             stack_timeout=0,
             required_version=">1.0",
             template_bucket_name="stack_group_template_bucket_name",
```
jfalkenstein, line 306: You should also add two tests that shows disable_rollback being set to True when it's present on the config and in the command params.

jfalkenstein (changes requested):
We also should add integration tests that test something along the lines that:
When a stack that creates two resources, one properly configured and one improperly configured gets deployed with disable rollback, once it's launched with disable-rollback, the properly configured resource still exists.

docs/_source/docs/stack_config.rst
zaro0508: good question,  it’s because i’ve followed the pattern that’s been setup in the existing code base.  it’s actually what you suggested in [your comment on the issue](https://github.com/Sceptre/sceptre/issues/1179#issuecomment-997427832).  We implement CLI flag only in the code or just leave this PR as-is and document CLI flag without documenting the `disable-rollback` stack config.

sceptre/cli/create.py
zaro0508: sounds good, will do

sceptre/cli/delete.py
zaro0508: no, it's not.

sceptre/plan/actions.py
zaro0508: have not done it yet but will add a test.

tests/test_config_reader.py
zaro0508: yes sir.

sceptre/cli/create.py
zaro0508: just following the existing pattern but can do as you suggest.

sceptre/cli/create.py
```diff
@@ -10,9 +10,15 @@
 @click.argument("path")
 @click.argument("change-set-name", required=False)
 @click.option("-y", "--yes", is_flag=True, help="Assume yes to all questions.")
+@click.option(
+    "-dr",
+    "--disable-rollback",
+    is_flag=True,
+    help="Disable the auto rollback and keep resources successfully created or updated",
+)
```
dboitnot, line 18: Since we're also doing a StackConfig for this option, we'll probably want to provide a CLI override. Rather than using `is_flag`, recommend providing both options:
```suggestion
@click.option(
  "--disable-rollback/--enable-rollback",
  default=None,
  help="Disable the auto rollback and keep resources successfully created or updated",
)
```

See: https://click.palletsprojects.com/en/8.1.x/options/#boolean-flags

sceptre/cli/create.py
```diff
@@ -10,9 +10,15 @@
...
 @click.pass_context
 @catch_exceptions
-def create_command(ctx, path, change_set_name, yes):
+def create_command(ctx, path, change_set_name, yes, disable_rollback: bool):
```
dboitnot, line 21: If we adopt the enable/disable suggestion above, we'll need to update this function signature:

```suggestion
def create_command(ctx, path, change_set_name, yes, disable_rollback: Optional[bool]):
```

sceptre/cli/launch.py
```diff
@@ -24,9 +24,17 @@
     is_flag=True,
     help="If set, will delete all stacks in the command path marked as obsolete.",
 )
+@click.option(
+    "-dr",
+    "--disable-rollback",
+    is_flag=True,
+    help="Disable the auto rollback and keep resources successfully created or updated",
+)
```
dboitnot, line 32: Same as above.

```suggestion
@click.option(
  "--disable-rollback/--enable-rollback",
  default=None,
  help="Disable the auto rollback and keep resources successfully created or updated",
)
```

sceptre/cli/launch.py
```diff
@@ -24,9 +24,17 @@
...
 @click.pass_context
 @catch_exceptions
-def launch_command(ctx: Context, path: str, yes: bool, prune: bool):
+def launch_command(
+    ctx: Context, path: str, yes: bool, prune: bool, disable_rollback: bool
+):
```
dboitnot, line 37: Same as above.

```suggestion
def launch_command(
    ctx: Context, path: str, yes: bool, prune: bool, disable_rollback: Optional[bool]
):
```

sceptre/cli/update.py
```diff
@@ -17,9 +17,15 @@
 )
 @click.option("-v", "--verbose", is_flag=True, help="Display verbose output.")
 @click.option("-y", "--yes", is_flag=True, help="Assume yes to all questions.")
+@click.option(
+    "-dr",
+    "--disable-rollback",
+    is_flag=True,
+    help="Disable the auto rollback and keep resources successfully created or updated",
+)
```
dboitnot, line 25: Same as above.

```suggestion
@click.option(
  "--disable-rollback/--enable-rollback",
  default=None,
  help="Disable the auto rollback and keep resources successfully created or updated",
)
```

sceptre/cli/update.py
```diff
@@ -17,9 +17,15 @@
...
 @click.pass_context
 @catch_exceptions
-def update_command(ctx, path, change_set, verbose, yes):
+def update_command(ctx, path, change_set, verbose, yes, disable_rollback: bool):
```
dboitnot, line 28:
```suggestion
def update_command(ctx, path, change_set, verbose, yes, disable_rollback: Optional[bool]):
```

sceptre/config/reader.py
```diff
@@ -552,6 +552,11 @@ def _construct_stack(self, rel_path, stack_group_config=None):
                 )
 
         s3_details = self._collect_s3_details(stack_name, config)
+        disable_rollback = (
+            True
+            if self.context.command_params.get("disable_rollback")
+            else config.get("disable_rollback", False)
+        )
```
dboitnot, line 559:
```suggestion
        # If disable/enable rollback was specified on the command line, use that. Otherwise,
        # fall back to the stack config.
        disable_rollback = self.context.command_params.get("disable_rollback")
        disable_rollback = config.get("disable_rollback", False) if disable_rollback is None else disable_rollback
```

dboitnot (changes requested):

sceptre/cli/create.py
jfalkenstein: Great feedback! You're absolutely right, @dboitnot. I didn't even think of that.

sceptre/cli/create.py
jfalkenstein: We only started adding type annotations like a year and a half ago. Most of our existing patterns were around long before that. However, in the era of Python type annotations (py 3.5+, I believe), there's no longer any need to indicate types in docstrings. Sphinx understands type annotations just fine as well and interprets them the same way.

sceptre/cli/launch.py
```diff
@@ -36,11 +44,13 @@ def launch_command(ctx: Context, path: str, yes: bool, prune: bool):
     * Any stacks that already exist will be updated (if there are any changes)
     * If any stacks are marked with "ignore: True", those stacks will neither be created nor updated
     * If any stacks are marked with "obsolete: True", those stacks will neither be created nor updated.
-      Furthermore, if the "-p"/"--prune" flag is used, these stacks will be deleted prior to any
+    * If the "-p"/"--prune" flag is used, these stacks will be deleted prior to any
```
jfalkenstein, line 47: No this line is a continuation of the previous point, not a new point.

sceptre/config/reader.py
dboitnot: Just realized this might be cleaner:
```python
        disable_rollback = self.context.command_params.get("disable_rollback")
        if disable_rollback is None:
          disable_rollback = config.get("disable_rollback", False)
```

docs/_source/docs/stack_config.rst
```diff
@@ -205,6 +206,22 @@ Examples include:
...
 
+disable_rollback
+~~~~~~~~~~~~~~~~
+* Resolvable: No
+* Can be inherited from StackGroup: Yes
+* Inheritance strategy: Overrides parent if set
+
+This parameter describes the action taken by CloudFormation when a Stack fails
+to create or update, default is False. This option can be set from the stack
+config or from the Sceptre CLI commands to deploy stacks. Enabling disable-rollback
```
jfalkenstein, line 217: Saying "enabling disable-rollback" sounds kind of weird. How about just "Disabling rollback..." intead?

docs/_source/docs/stack_config.rst
```diff
@@ -205,6 +206,22 @@ Examples include:
...
+using the CLI option disables rollback globally for all stacks. This option
+overrides on_failure since Cloudformation does not allow setting both on deployment.
+For more information and valid values see the `AWS Documentation`_.
+
+Examples include:
```
jfalkenstein, line 222: How about just "Example:"... ? 

"Examples include" doesn't really make sense for such a simple example. It's not like there are lots of possible configurations for this.

docs/_source/docs/stack_config.rst
```diff
@@ -205,6 +206,22 @@ Examples include:
...
+disable_rollback
+~~~~~~~~~~~~~~~~
+* Resolvable: No
+* Can be inherited from StackGroup: Yes
+* Inheritance strategy: Overrides parent if set
+
+This parameter describes the action taken by CloudFormation when a Stack fails
+to create or update, default is False. This option can be set from the stack
+config or from the Sceptre CLI commands to deploy stacks. Enabling disable-rollback
+using the CLI option disables rollback globally for all stacks. This option
```
jfalkenstein, line 218: Maybe clarify here what you mean by "using the CLI option"?

sceptre/cli/create.py
```diff
@@ -10,9 +11,14 @@
 @click.argument("path")
 @click.argument("change-set-name", required=False)
 @click.option("-y", "--yes", is_flag=True, help="Assume yes to all questions.")
+@click.option(
+    "--disable-rollback/--enable-rollback",
+    default=None,
+    help="Disable the auto rollback and keep resources successfully created or updated",
```
jfalkenstein, line 17: This `help` isn't completely accurate any more, since this can also be `--enable-rollback`, which doesn't disable it. I think all these CLI options here and below for disable/enable need to be updated along these lines.

docs/_source/docs/stack_config.rst
zaro0508: kept as-is since there seems to be some valid use cases for setting disable_rollback in stack config

sceptre/cli/create.py
zaro0508: implemented switch `--disable-rollback/--enable-rollback`

sceptre/cli/create.py
zaro0508: done

sceptre/plan/actions.py
zaro0508: tests have been added

tests/test_config_reader.py
zaro0508: tests added

sceptre/cli/launch.py
zaro0508: fixed

docs/_source/docs/stack_config.rst
zaro0508: fixed

docs/_source/docs/stack_config.rst
zaro0508: fixed

docs/_source/docs/stack_config.rst
zaro0508: fixed

sceptre/cli/create.py
zaro0508: fixed

tests/test_config_reader.py
```diff
@@ -387,6 +395,42 @@ def test_construct_stacks_with_valid_config(
...
+            "template_path": rel_path,
+        }
+
+        abs_path = os.path.join(config_dir, rel_path)
+        self.write_config(abs_path, config)
+        self.context.project_path = project_path
+        self.context.command_params["disable_rollback"] = True
+        config_reader = ConfigReader(self.context)
+        all_stacks, command_stacks = config_reader.construct_stacks()
+        assert "disable_rollback=True" in str(all_stacks)
```
jfalkenstein, line 414: This strikes me as a bit odd. Why are you asserting off the string of the stacks? Why not loop the stacks and check that they all have `disable_rollback` equal to True? Technically this test is only testing that just one Stack in the list of stacks has `disable_rollback`.

tests/test_config_reader.py
```diff
@@ -387,6 +395,42 @@ def test_construct_stacks_with_valid_config(
...
+            "template_path": rel_path,
+            "disable_rollback": True,
+        }
+
+        abs_path = os.path.join(config_dir, rel_path)
+        self.write_config(abs_path, config)
+        self.context.project_path = project_path
+        config_reader = ConfigReader(self.context)
+        all_stacks, command_stacks = config_reader.construct_stacks()
+        assert "disable_rollback=True" in str(all_stacks)
```
jfalkenstein, line 432: same comment here.

tests/test_config_reader.py
zaro0508: this test doesn't require testing with multiple stacks.  It's is deliberately testing with just one stack therefore no need to loop?

jfalkenstein: It still isn't really proper to test off the `str` output of a thing unless your test is on the `__str__` method itself. Rather, you should assert off the property you're _actually_ testing for. Like this:

```suggestion
        assert list(all_stacks)[0].disable_rollback
```

zaro0508: updated the assert

tests/test_config_reader.py
zaro0508: fixed

jfalkenstein (approved):
Great feature we've needed for a while now! Thanks for putting this together!

sceptre/cli/launch.py
```diff
@@ -24,9 +24,16 @@
     is_flag=True,
     help="If set, will delete all stacks in the command path marked as obsolete.",
 )
+@click.option(
+    "--disable-rollback/--enable-rollback",
+    default=False,
```
dboitnot, line 29: Unless I'm missing something, I think we might want `default=None` here. That way if the user doesn't specify `--disable...` or `--enable...` it will fall back to whatever's in the StackConfig (or False if nothing's there).

dboitnot (commented):
Sorry I only noticed this after the merge.

sceptre/cli/launch.py
jfalkenstein: Great catch!

dboitnot: Is this something for @zaro0508 to deal with or should I open another PR?

jfalkenstein: If you have opportunity and time to get it done, I say go for it. I'm not sure what @zaro0508 has going on over the next few days and if he'll have time. (I'll let him speak for himself, though).

zaro0508: yes, please. would love to see a PR from you @dboitnot 😄

dboitnot: Added #1288. Since it was a one-word update I tried doing it through the GitHub interface.

PR #1285, opened 2023-01-08
[Resolve #1264] Updating the CDK docs to point to the new sceptre-cdk-handler

### LINKED ISSUE #1264: The Sceptre Template AWS CDK Example is out of date and does not support CDK Assets
### Subject of the issue

The Sceptre template AWS CDK example in https://github.com/Sceptre/sceptre/blob/master/docs/_source/docs/templates.rst is out of date, as it is for AWS CDK v1 rather than the current v2 and it does not support CDK stacks that require assets (constructs that create CloudFormation custom resources for example). 

### Your environment
* version of sceptre `3.2.0`
* version of python `3.8.8`
* which OS/distro: `Windows 10`
* AWS CDK version: `2.51.1`

### Steps to reproduce

Use the following code with the current AWS CDK example. The `BucketDeployment` construct creates a custom CloudFormation resource:

```python
        s3_bucket = s3.Bucket(self,
                              "S3Bucket",
                              bucket_name=sceptre_user_data['BucketName']
                              )
 
        cdk.aws_s3_deployment.BucketDeployment(self,
                                               "S3Deployment",
                                               sources=[cdk.aws_s3_deployment.Source.data(
                                                   'object-key.txt', 'hello, world!')],
                                               destination_bucket=s3_bucket
                                               )
```
 
### Expected behaviour

The CDK stack should be deployed successfully,

### Actual behaviour

`sceptre launch` fails with the following error:
```
ImportError: cannot import name 'core' from 'aws_cdk'
```

Once the example code has been refactored for CDK v2.x `sceptre launch` fails with the following error:
```
[2022-11-23 16:52:08] - dev/cdk/s3/s3-original S3DeploymentAwsCliLayer8AAFE44F AWS::Lambda::LayerVersion CREATE_FAILED Error occurred while GetObject. S3 Error Code: NoSuchKey. S3 Error Message: The specified key does not exist. (Service: AWSLambdaInternal; Status Code: 400; Error Code: InvalidParameterValueException; Request ID: 370f2978-891f-428d-921e-1f7116bacc19; Proxy: null)
```

This is due to the CDK assets for the CloudFormation custom resource not having been published.

### Solution

I have an AWS CDK v2 example for Sceptre that published the CDK assets and I am happy to raise a PR to update the current example with it.

jfalkenstein:
This updates our docs to no longer reference the old CDK approach (which didn't work with CDK assets). In its place, it references the new `sceptre-cdk-handler` package that covers that functionality.

## PR Checklist

## Other Information

### DIFF
--- a/docs/_source/docs/faq.rst
+++ b/docs/_source/docs/faq.rst
@@ -83,15 +83,6 @@ used.
 
 .. _AWS documentation: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html
 
-How do I build a Serverless application using Sceptre and SAM?
---------------------------------------------------------------
-
-There is now a SAM Template Handler that lets you incorporate SAM applications into environments that
-are managed and deployed using Sceptre. For more information on how to install and use SAM in your
-Sceptre project, see the `sceptre-sam-handler`_ page on PyPI.
-
-
-.. _sceptre-sam-handler: https://pypi.org/project/sceptre-sam-handler/
 
 My CI/CD process uses ``sceptre launch``. How do I delete stacks that aren't needed anymore?
 ---------------------------------------------------------------------------------------------
--- a/docs/_source/docs/templates.rst
+++ b/docs/_source/docs/templates.rst
@@ -190,53 +190,21 @@ to generate CloudFormation Template as a `json` string.
 AWS CDK
 ^^^^^^^
 
-This example generates a cloudformation template from `AWS CDK`_ code.
+AWS CDK can be used to programmatically generate templates for your stacks and then Sceptre can
+deploy those stacks. Taking this approach leverages the best capabilities of CDK (simple template
+generation with sane defaults) and the best capabilities of Sceptre ("wiring together" stacks and
+deploying them as consistent environments, leveraging hooks and resolvers to dynamically connect them).
 
-.. code-block:: yaml
+In order to use CDK with Sceptre, you need to install the `sceptre_cdk_handler`_ package. You can find
+documentation and examples on how to use it on the `github repository`_.
 
-  sceptre_user_data:
-    BucketName: my-bucket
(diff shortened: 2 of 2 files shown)


docs/_source/docs/faq.rst
```diff
@@ -83,15 +83,6 @@ used.
...
 
-How do I build a Serverless application using Sceptre and SAM?
---------------------------------------------------------------
-
-There is now a SAM Template Handler that lets you incorporate SAM applications into environments that
-are managed and deployed using Sceptre. For more information on how to install and use SAM in your
-Sceptre project, see the `sceptre-sam-handler`_ page on PyPI.
-
-
-.. _sceptre-sam-handler: https://pypi.org/project/sceptre-sam-handler/
```
jfalkenstein, line 94: I moved this to the section on templates. It really makes more sense there.

docs/_source/docs/templates.rst
```diff
@@ -190,53 +190,21 @@ to generate CloudFormation Template as a `json` string.
 AWS CDK
 ^^^^^^^
 
-This example generates a cloudformation template from `AWS CDK`_ code.
+AWS CDK can be used to programmatically generate templates for your stacks and then Sceptre can
```
zaro0508, line 193: hmm, looks like we have two locations where we are referencing template handlers.  what do you think about consolidating that info.  Maybe move [list of available template handlers](https://docs.sceptre-project.org/3.2.0/docs/template_handlers.html#available-template-handlers) to this page?

zaro0508 (approved):
In general LGTM.  Just had a question that you might want to think about.

docs/_source/docs/templates.rst
jfalkenstein: I think doing that might need a larger-scale rewrite of those two pages. I don't want to do that in this PR, though.

zaro0508: sounds good

PR #1300, opened 2023-02-05
[Resolve #1299] Making ConnectionManager.call() work the way we all want it to

### LINKED ISSUE #1299: ConnectionManager's call() method is too easy to misuse by not passing enough variables.
### Subject of the issue
The ConnectionManager is used by many hooks, plugins, and template managers. It is essential for interacting with AWS using the StackConfig's values. However, as a result of how it operates, it can (and often is) misused accidentally. 

The source of the issue is here: 
image

The problem is that this code assumes that if someone wants to use the Stack default configurations, one would pass `None` for `region`, `profile`, and `sceptre_role`. If all three are None, it is assumed that the stack values ought to be utilized. But if any _one_ of those is not None, then it is assumed that all three have been specified and all three are used.

The result of this is we end up with outdated hooks and resolvers [(such as the SSM Resolver)](https://github.com/Sceptre/sceptre-ssm-resolver/issues/13) that are passing the profile and region, but not the sceptre_role (aka iam_role). Since `region` is always specified (by necessity), it will mean the `iam_role` parameter is overridden and set to `None`.

Fundamentally, the issue is that there isn't here a distinction between setting a value to `None` to _nullify_ a setting (meaning "don't use any iam_role") and setting a value to `None` to retain the stack's configuration. 

The solution to this is set the default values for `call()` to a third constant that is interpreted as the Stack default, similar to how the new `get_session()` method operates: 

image

jfalkenstein:
For a complete description on the situation that we need to resolve, see issue #1299.

This PR updates the logic on the `call()` method (the primary public method on ConnectionManager) to make it operate with fewer "gotchyas" involved. See the docstring and extensive comments within the body of the function for more explicit details, but here's the general TLDR of this change:

* `profile`, `region`, and `sceptre_role` parameters all default now to a special STACK_DEFAULT constant, which means "use the default configuration of the targeted stack". _When `stack_name` is `None`_, this means it will use the default configuration for these values for the CURRENT Stack; _When the `stack_name` is NOT `None`_, this means it will use the default configuration for the _named_ stack.
* When any one or two of those is set to `None`, it means "I explicitly want to nullify this value". For example, if the target stack has an `sceptre_role` set, you can set `sceptre_role=None` and it will mean "Don't assume any role when performing this call". (Any remaining values that are set to STACK_DEFAULT will continue to default to the target stack's configurations).
* When any (or all) of these values are explicitly set, it means "Use these values instead of the stack defaults".
* For historical/legacy support reasons: If all three values are set to "None", it is interpreted as meaning the same thing as all three values being set to STACK_DEFAULT. This should be relatively safe, because that would otherwise mean overriding the region to set it to `None` and that should blow up with boto3 in every circumstance and thus is unlikely to ever be the desired behavior. 

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/connection_manager.py
+++ b/sceptre/connection_manager.py
@@ -72,11 +72,6 @@ def _retry_boto_call(func):
     return decorated
 
 
-# STACK_DEFAULT is a sentinel value meaning "default to the stack's configuration". This is in
-# contrast with passing None, which would mean "use no value".
-STACK_DEFAULT = "[STACK DEFAULT]"
-
-
 class ConnectionManager(object):
     """
     The Connection Manager is used to create boto3 clients for
@@ -89,7 +84,9 @@ class ConnectionManager(object):
     :param sceptre_role_session_duration: The duration to assume the specified sceptre_role per session.
     """
 
-    STACK_DEFAULT = STACK_DEFAULT
+    # STACK_DEFAULT is a sentinel value meaning "default to the stack's configuration". This is in
+    # contrast with passing None, which would mean "use no value".
+    STACK_DEFAULT = "[STACK DEFAULT]"
 
     _session_lock = threading.Lock()
     _client_lock = threading.Lock()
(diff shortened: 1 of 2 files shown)


ericabrauer (commented):
Still getting familiar with the code so this review is more for my own SA, but looks great to me @jfalkenstein 👍🏻

zaro0508 (approved):

PR #1309, opened 2023-02-18
[Resolve #1304] Adding StackLoggerAdapter to hooks, resolvers, and template handlers

### LINKED ISSUE #1304: Logs out of Template Handlers, Hooks, and Resolvers don't indicate what stack they're operating on
### Subject of the issue
This is really a copy of the issue raised here: https://github.com/Sceptre/sceptre-cdk-handler/issues/9

The basic issue here is that Sceptre creates a logger in the base class of Template Handlers, Hooks, and Resolvers, but that logger doesn't indicate by default the stack it is acting upon, which can cause ambiguous logs to be emitted. Rather than having a patchwork approach to this across all resolvers, hooks, and template handlers, it'd be better to have a consistent system where loggers can be set up to prefix all log messages with the stack name.

A solution has been suggested by @X-Guardian that is likely to be the most viable approach, using a LoggingAdapter. There was an approach introduced [here](https://github.com/Sceptre/sceptre-cdk-handler/commit/25b4d31736824c0763a1d746f4bccaffc3d3783c#diff-f9f02cf5496a224e547fb8115ef2a2f782afebe34385546a1db96e45f4e09daaR28) and since removed from the CDK Handler code. We should investigate implementing a standardized logger adapter for all Hooks, Resolvers, and Template Handlers that add that prefix and any other relevant contextual information).

### Your environment
* version of sceptre (sceptre --version): 4.0.0

jfalkenstein:
This converts the loggers for Hooks, Resolvers, and Template Handlers to be LoggerAdapters that prefix all log messages with the Stack name

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/hooks/__init__.py
+++ b/sceptre/hooks/__init__.py
@@ -3,21 +3,23 @@ import logging
 from functools import wraps
 
 from sceptre.helpers import _call_func_on_values
+from sceptre.logging import StackLoggerAdapter
+
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from sceptre.stack import Stack
 
 
-class Hook(object):
+class Hook(abc.ABC):
     """
     Hook is an abstract base class that should be inherited by all hooks.
 
     :param argument: The argument of the hook.
-    :type argument: str
     :param stack: The associated stack of the hook.
-    :type stack: sceptre.stack.Stack
     """
 
-    __metaclass__ = abc.ABCMeta
-
-    def __init__(self, argument=None, stack=None):
+    def __init__(self, argument: Any = None, stack: "Stack" = None):
         self.logger = logging.getLogger(__name__)
         self.argument = argument
         self.stack = stack
@@ -61,15 +63,16 @@ class HookProperty(object):
         """
         return getattr(instance, self.name)
 
-    def __set__(self, instance, value):
+    def __set__(self, instance: "Stack", value):
         """
         Attribute setter which adds a stack reference to any hooks in the
         data structure `value` and calls the setup method.
 
         """
 
-        def setup(attr, key, value):
+        def setup(attr, key, value: Hook):
             value.stack = instance
+            value.logger = StackLoggerAdapter(value.logger, instance.name)
             value.setup()
 
         _call_func_on_values(setup, value, Hook)
--- a/sceptre/hooks/__init__.py
+++ b/sceptre/hooks/__init__.py
@@ -21,6 +21,10 @@ class Hook(abc.ABC):
 
     def __init__(self, argument: Any = None, stack: "Stack" = None):
         self.logger = logging.getLogger(__name__)
+
+        if stack is not None:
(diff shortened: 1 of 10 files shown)


sceptre/hooks/__init__.py
```diff
@@ -37,6 +43,15 @@ def run(self):
...
 
+    def clone(self, stack: "Stack") -> "Hook":
+        """
+        Produces a "fresh" copy of the Hook, with the specified stack.
+
+        :param stack: The stack to set on the cloned resolver
+        """
+        clone = type(self)(self.argument, stack)
+        return clone
+
```
jfalkenstein, line 55: This makes hooks operate more like Resolvers so we have more consistent behavior all around.

sceptre/hooks/__init__.py
```diff
@@ -61,16 +76,16 @@ def __get__(self, instance, type):
...
         data structure `value` and calls the setup method.
 
         """
 
-        def setup(attr, key, value):
-            value.stack = instance
-            value.setup()
+        def setup(attr, key, value: Hook):
+            attr[key] = clone = value.clone(instance)
+            clone.setup()
```
jfalkenstein, line 88: This change actually patches a bug that we've had for a long time where the Stack instance on hooks inherited from parent StackGroups might not actually be the same as the current Stack. This ensures a fresh Hook is set up on each stack, even if its parameters have been inherited.

sceptre/logging.py
```diff
@@ -0,0 +1,20 @@
...
+class StackLoggerAdapter(LoggerAdapter):
+    def __init__(self, logger: Logger, stack_name: str, extra: dict = None):
+        """A small wrapper around a Logger that prefixes log messages with the stack name.
+
+        :param logger: The logger to wrap
+        :param stack_name: The name of the stack to every log message
+        :param extra: Extra kwargs to add to the log context (if any)
+        """
+        super().__init__(logger, extra or {})
+        self.stack_name = stack_name
```
dboitnot, line 14: I like this idea. My only question is whether the situation might arise when a StackLoggingAdapter is wrapping another StackLoggingAdapter. If this is a possibility should we detect it in `__init__` and maybe wrap the inner logger instead of the outer one to avoid chaining the stack names?

dboitnot (approved):
Looks good.

sceptre/logging.py
jfalkenstein: So, _technically_ (so long as you're using the public interface of a logger), wrapping a logger adapter with a logger adapter shouldn't actually break anything. You'd just end up with functionality from Both.

However, if you look at the way that this logger adapter is being used (and the _only_ way it's being used), it's being used on pretty much the first line of the `__init__` method of classes that don't have the logger injected. So in practice, the only way this scenario could happen would be if someone in their custom hook, resolver, or handler did something like: `self.log = MyOtherLoggerAdapter(self.log, self.stack.name)`. Even if they did, it wouldn't really break anything.

dboitnot: Sounds good to me.

PR #1312, opened 2023-02-25
[Resolve #1263] Stack output caching

### LINKED ISSUE #1263: Sceptre shouldn't need to make redundant !stack_output calls
### Subject of the issue
Right now, if you have the same `!stack_output` resolver argument spread across many stacks, sceptre will get the stack outputs for that stack to resolve the value for every single one. The leads to a longer deployment process with MANY more CloudFormation API calls than it needs.

I think Sceptre should implement some sort of stack_output resolver lazy-loaded cache where identical arguments should always resolve to the same value without future api calls needed. This should be fairly trivial to implement, though I think the question of just how to maintain a global state between stacks would be a relevant implementation concern.

jfalkenstein:
This PR caches the stack outputs of each cache so that we're not making redundant requests for the same stack outputs over and over again (which is what we currently do now in every instance we use !stack_output or !stack_output_external).

Thanks to @dboitnot for the very simple solution suggestion!

(I've not written any tests because I believe that if existing unit and integration tests pass, this should be considered very "safe".)

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/resolvers/stack_output.py
+++ b/sceptre/resolvers/stack_output.py
@@ -1,21 +1,18 @@
 # -*- coding: utf-8 -*-
 
-import abc
-import six
+import functools
 import logging
 import shlex
 
 from botocore.exceptions import ClientError
 
+from sceptre.exceptions import DependencyStackMissingOutputError, StackDoesNotExistError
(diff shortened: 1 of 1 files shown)


dboitnot (approved):
Looks good.

PR #1313, opened 2023-02-25
[Resolve #1252] Supporting resolvers in Hook and Resolver arguments, with new !substitute, !join, !split, and !select resolvers!

### LINKED ISSUE #1252: Question: Join Resolver output with additional strings possible?
I'm wondering how you can use resolver outputs with some additional strings - e.g. use the stack_output and use some additional strings to attach / join them with. Can Jinja2 be use for this?

Usecase for example is to build a connection string (host:hostname:port?option=A), e.g.

parameters:
  paramA: `host:!stack_output <stack_name>::output:1234?option=A`

Dose the new stack_attr help here, e.g.

host_endpoint: !stack_output <stack_name>::output

parameters:
  paramA: `host:{{ stack_attr.host_endpoint }}:1234?option=A`

jfalkenstein:
This pull request brings the ability for Hooks and Resolvers to have resolvers in their arguments and have those resolvers automatically resolve when `self.argument` is accessed. When a resolver or hook is `setup`, it will also setup all nested resolvers in its argument. And when it is attached to a Stack, it and its argument will be recursively cloned so that no resolver is associated with a different stack. This will let resolvers be "inherited" from parent stack configurations.

## PR Checklist

## Other Information

### DIFF
--- a/docs/_source/docs/hooks.rst
+++ b/docs/_source/docs/hooks.rst
@@ -207,7 +207,9 @@ This hook can be used in a Stack config file with the following syntax:
 
 hook arguments
 ^^^^^^^^^^^^^^
-Hook arguments can be a simple string or a complex data structure.
+Hook arguments can be a simple string or a complex data structure. You can even use resolvers in
+hook arguments, so long as they're nested in a list or a dict.
+
 Assume a Sceptre `copy` hook that calls the `cp command`_:
 
 .. code-block:: yaml
@@ -224,7 +226,7 @@ Assume a Sceptre `copy` hook that calls the `cp command`_:
        - !copy
            options: "-r"
            source: "from_dir"
-           destination: "to_dir"
+           destination: !stack_output my/other/stack::CopyDestination
 
 .. _Custom Hooks: #custom-hooks
 .. _subprocess documentation: https://docs.python.org/3/library/subprocess.html
--- a/docs/_source/docs/resolvers.rst
+++ b/docs/_source/docs/resolvers.rst
@@ -253,7 +253,7 @@ sub
 ~~~~~~~~~~
 
 This resolver allows you to create a string using Python string format syntax. This functions as a
-great way to combine together a number of resolver outputs into a single string. This functions very
+great way to combine together a number of resolver outputs into a single string. This functions
 similarly to Cloudformation's ``!Sub`` intrinsic function.
 
 The argument to this resolver should be a two-element list: (1) Is the format string, using
--- a/docs/_source/docs/resolvers.rst
+++ b/docs/_source/docs/resolvers.rst
@@ -5,6 +5,9 @@ Sceptre implements resolvers, which can be used to resolve a value of a
 CloudFormation ``parameter`` or ``sceptre_user_data`` value at runtime. This is
 most commonly used to chain the outputs of one Stack to the inputs of another.
 
+You can use resolvers with any resolvable property on a StackConfig, as well as in the arguments
+of hooks and other resolvers.
+
 If required, users can create their own resolvers, as described in the section
 on `Custom Resolvers`_.
 
@@ -52,6 +55,27 @@ file_contents
 
 **deprecated**: Consider using the `file`_ resolver instead.
 
+join
+~~~~
+
+This resolver allows you to join multiple strings together to form a single string. This is great
+for combining the outputs of multiple resolvers. This resolver works just like CloudFormation's
+``!Join`` intrinsic function.
+
+The argument for this resolver should be a list with two elements: (1) A string to join the elements
+on and (2) a list of items to join.
+
+Example:
+
+.. code-block:: yaml
+
+   parameters:
+     BaseUrl: !join
+       - ":"
+       - - !stack_output my/app/stack.yaml::HostName
+         - !stack_output my/other/stack.yaml::Port
+
+
 no_value
 ~~~~~~~~
 
@@ -80,6 +104,49 @@ A resolver to execute any shell command.
 Refer to `sceptre-resolver-cmd <https://github.com/Sceptre/sceptre-resolver-cmd/>`_ for documentation.
 
 .. _stack_attr_resolver:
+
+select
+~~~~~~
+
+This resolver allows you to select a specific index of a list of items. This is great for combining
+with the ``!split`` resolver to obtain part of a string. This function works almost the same as
+CloudFormation's ``!Select`` intrinsic function, **except you can use this with negative indexes to
+select with a reverse index**.
+
+The argument for this resolver should be a list with two elements: (1) A numerical index and (2) a
+list of items to select out of. If the index is negative, it will select from the end of the list.
+For example, "-1" would select the last element and "-2" would select the second-to-last element.
+
+Example:
+
+.. code-block:: yaml
+
+   parameters:
+     # This selects the last element after you split the connection string on "/"
+     DatabaseName: !select
+       - -1
+       - !split ["/", !stack_output my/database/stack.yaml::ConnectionString]
+
+split
+~~~~~
+
+This resolver will split a value on a given delimiter string. This is great when combining with the
+``!select`` resolver. This function works the same as CloudFormation's ``!Split`` intrinsic function.
+
+Note: The return value of this resolver is a *list*, not a string. This will not work to set Stack
+configurations that expect strings, but it WILL work to set Stack configurations that expect lists.
+
+The argument for this resolver should be a list with two elements: (1) The delimiter to split on and
+(2) a string to split.
+
+Example:
+
+.. code-block:: yaml
+
+   notifications: !split
+     - ";"
+     - !stack_output my/sns/topics.yaml::SemicolonDelimitedArns
+
 
 stack_attr
 ~~~~~~~~~~
@@ -180,6 +247,32 @@ Example:
 
    parameters:
      VpcIdParameter: !stack_output_external prj-network-vpc::VpcIdOutput prod
+
+
+substitute
+~~~~~~~~~~
+
+This resolver allows you to create a string using Python string format syntax. This functions as a
+great way to combine together a number of resolver outputs into a single string. This functions very
+similarly to Cloudformation's ``!Sub`` intrinsic function.
+
+The argument to this resolver should be a two-element list: (1) Is the format string, using
+curly-brace templates to indicate variables, and (2) a dictionary where the keys are the format
+string's variable names and the values are the variable values.
+
+Example:
+
+.. code-block:: yaml
+
+   parameters:
+     ConnectionString: !substitute
+       - "postgres://{username}:{password}@{hostname}:{port}/{database}"
+       - username: {{ var.username }}
+         password: !ssm /my/ssm/password
+         hostname: !stack_output my/database/stack.yaml::HostName
+         port: !stack_output my/database/stack.yaml::Port
+         database: {{var.database}}
+
 
 Custom Resolvers
 ----------------
@@ -306,18 +399,22 @@ For details on calling AWS services or invoking AWS-related third party tools in
 
 Resolver arguments
 ^^^^^^^^^^^^^^^^^^
-Resolver arguments can be a simple string or a complex data structure.
+Resolver arguments can be a simple string or a complex data structure. You can even use
+other resolvers in the arguments to resolvers! (Note: Other resolvers can only be passed in
+arguments when they're passed in lists and dicts.)
 
 .. code-block:: yaml
 
    template:
      path: <...>
      type: <...>
-    parameters:
-      Param1: !ssm "/dev/DbPassword"
-      Param2: !ssm {"name": "/dev/DbPassword"}
-      Param3: !ssm
-        name: "/dev/DbPassword"
+   parameters:
+     Param1: !ssm "/dev/DbPassword"
+     Param2: !ssm {"name": "/dev/DbPassword"}
+     Param3: !ssm
+       name: "/dev/DbPassword"
+     Param4:
+       name: !stack_output my/other/stack.yaml::MyOutputName
 
 .. _Custom Resolvers: #custom-resolvers
 .. _this is great place to start: https://docs.python.org/3/distributing/
--- a/docs/_source/docs/resolvers.rst
+++ b/docs/_source/docs/resolvers.rst
@@ -121,7 +121,7 @@ Example:
 
 .. code-block:: yaml
 
-   parameters:
+   sceptre_user_data:
      # This selects the last element after you split the connection string on "/"
      DatabaseName: !select
        - -1
--- a/docs/_source/docs/resolvers.rst
+++ b/docs/_source/docs/resolvers.rst
@@ -414,7 +414,7 @@ arguments when they're passed in lists and dicts.)
      Param3: !ssm
        name: "/dev/DbPassword"
      Param4: !ssm
-       name: !stack_output my/other/stack.yaml::MySsmName
+       name: !stack_output my/other/stack.yaml::MySsmParameterName
 
 .. _Custom Resolvers: #custom-resolvers
 .. _this is great place to start: https://docs.python.org/3/distributing/
--- a/docs/_source/docs/resolvers.rst
+++ b/docs/_source/docs/resolvers.rst
@@ -249,7 +249,7 @@ Example:
      VpcIdParameter: !stack_output_external prj-network-vpc::VpcIdOutput prod
 
 
-substitute
+sub
 ~~~~~~~~~~
 
 This resolver allows you to create a string using Python string format syntax. This functions as a
@@ -265,7 +265,7 @@ Example:
 .. code-block:: yaml
 
    parameters:
-     ConnectionString: !substitute
+     ConnectionString: !sub
        - "postgres://{username}:{password}@{hostname}:{port}/{database}"
        - username: {{ var.username }}
          password: !ssm /my/ssm/password
--- a/docs/_source/docs/resolvers.rst
+++ b/docs/_source/docs/resolvers.rst
@@ -249,7 +249,7 @@ Example:
 
 
 sub
-~~~~~~~~~~
+~~~
 
 This resolver allows you to create a string using Python string format syntax. This functions as a
 great way to combine together a number of resolver outputs into a single string. This functions
--- a/docs/_source/docs/resolvers.rst
+++ b/docs/_source/docs/resolvers.rst
@@ -110,8 +110,8 @@ select
 
 This resolver allows you to select a specific index of a list of items. This is great for combining
 with the ``!split`` resolver to obtain part of a string. This function works almost the same as
-CloudFormation's ``!Select`` intrinsic function, **except you can use this with negative indexes to
-select with a reverse index**.
+CloudFormation's ``!Select`` intrinsic function, **except you can use this with negative indices to
+select from the end of a list**.
 
 The argument for this resolver should be a list with two elements: (1) A numerical index and (2) a
 list of items to select out of. If the index is negative, it will select from the end of the list.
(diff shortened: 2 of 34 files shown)


sceptre/hooks/__init__.py
```diff
@@ -20,21 +22,12 @@ class Hook(abc.ABC):
...
-        self.argument = argument
-        self.stack = stack
-
-    def setup(self):
-        """
-        setup is a method that may be overwritten by inheriting classes. Allows
-        hooks to run so initalisation steps when config is first read.
-        """
-        pass  # pragma: no cover
-
```
jfalkenstein, line 37: This is now defined in the base class

sceptre/hooks/__init__.py
```diff
@@ -43,15 +36,6 @@ def run(self):
...
 
-    def clone(self, stack: "Stack") -> "Hook":
-        """
-        Produces a "fresh" copy of the Hook, with the specified stack.
-
-        :param stack: The stack to set on the cloned resolver
-        """
-        clone = type(self)(self.argument, stack)
-        return clone
-
```
jfalkenstein, line 54: This is now defined on the base class

sceptre/hooks/__init__.py
```diff
@@ -1,39 +1,21 @@
...
-from typing import TYPE_CHECKING, Any
+from sceptre.resolvers import CustomYamlTagBase
 
 if TYPE_CHECKING:
     from sceptre.stack import Stack
 
 
-class Hook(abc.ABC):
+class Hook(CustomYamlTagBase, metaclass=abc.ABCMeta):
```
jfalkenstein, line 13: The big change here is that, since there's very little that differentiates a hook from a resolver, other than the "resolve" vs "run" public method, they now share a common base class. A lot of the functionality on Hooks has now been moved up into that base class since it was shared with Resolvers.

sceptre/hooks/__init__.py
```diff
@@ -84,7 +57,7 @@ def __set__(self, instance: "Stack", value):
         """
 
         def setup(attr, key, value: Hook):
-            attr[key] = clone = value.clone(instance)
+            attr[key] = clone = value.clone_for_stack(instance)
```
jfalkenstein, line 60: `clone_for_stack` (as you'll see below) is a recursive clone operation that allows for cloning resolvers inside of the arguments.

sceptre/resolvers/__init__.py
```diff
@@ -17,37 +17,158 @@
...
         initial setup of the object.
         """
         pass  # pragma: no cover
 
+    def __str__(self):
+        as_str = f"!{self.__class__.__name__}"
+        if self._argument is not None:
+            as_str += f"({self._argument})"
+
+        return as_str
```
jfalkenstein, line 164: This was mostly moved from the placeholders.py file. Previously, placeholders would access `resolver.argument` to represent the resolver when it couldn't be resolved. But if we're going to allow nested resolvers, that won't work. Instead, we need to represent the raw resolvers _all the way down_. Defining a `__str__` method on this base class and using the stringification of `self._argument` ensures that if any resolver in the argument cannot be resolved, it's placeholders end up being used all the way down into its arguments.

sceptre/resolvers/__init__.py
```diff
@@ -17,37 +17,169 @@
...
+        be converted to useful placeholder values.
+        """
+        as_str = f"!{self.__class__.__name__}"
+        if self._argument is not None:
+            as_str += f"({self._argument})"
+
+        return as_str
+
+
+class Resolver(CustomYamlTagBase, metaclass=abc.ABCMeta):
```
jfalkenstein, line 178: I've moved most of the functionality that _was_ in the Resolver class into the CustomYamlTagBase, which is now shared by both the Resolver and Hook classes.

sceptre/resolvers/__init__.py
```diff
@@ -127,23 +251,6 @@ def _no_recursive_get(self, stack: "stack.Stack"):
...
-        :param resolver: The Resolver to clone and set up
-        :return: The cloned resolver.
-        """
-        # We clone the resolver when we assign the value so that every stack gets its own resolver
-        # rather than potentially having one resolver instance shared in memory across multiple
-        # stacks.
-        clone = resolver.clone(stack)
-        clone.setup()
-        return clone
-
```
jfalkenstein, line 146: This method is no longer necessary. It's all been moved into the CustomYamlTagBase's `clone_for_stack` method.

sceptre/helpers.py
```diff
@@ -68,6 +68,28 @@ def func_on_instance(key):
     return attr
 
 
+Container = Union[list, dict]
+Key = Union[str, int]
+
+
+def delete_keys_from_containers(keys_to_delete: List[Tuple[Container, Key]]):
```
jfalkenstein, line 75: This is mostly just extracted out of the ResolvableProperty class and put here. I needed to also use this in the CustomYamlTagBase class.

sceptre/resolvers/placeholders.py
```diff
@@ -65,10 +65,7 @@ def _create_explicit_resolver_placeholder(resolver: "resolvers.Resolver") -> str
     :param resolver: The resolver to create a placeholder for
     :return: The placeholder value
     """
-    base = f"!{type(resolver).__name__}"
-    suffix = f"({resolver.argument})" if resolver.argument is not None else ""
-    # double-braces in an f-string is just an escaped single brace
-    return f"{{ {base}{suffix} }}"
```
jfalkenstein, line 71: This functionality has been moved to the CustomYamlTagBase class's `__repr__` method, where it accounts for arguments properly.

docs/_source/docs/resolvers.rst
```diff
@@ -81,6 +105,49 @@ Refer to `sceptre-resolver-cmd <https://github.com/Sceptre/sceptre-resolver-cmd/
 
 .. _stack_attr_resolver:
 
+select
+~~~~~~
+
+This resolver allows you to select a specific index of a list of items. This is great for combining
+with the ``!split`` resolver to obtain part of a string. This function works almost the same as
+CloudFormation's ``!Select`` intrinsic function, **except you can use this with negative indexes to
+select with a reverse index**.
```
dboitnot, line 114: What about "except you can use this with negative indices to select from the end of the list." ?

docs/_source/docs/resolvers.rst
```diff
@@ -181,6 +248,32 @@ Example:
    parameters:
      VpcIdParameter: !stack_output_external prj-network-vpc::VpcIdOutput prod
 
+
+substitute
```
dboitnot, line 252: I don't have super-strong feelings about this, but why "substitute" and not "sub" as in CF? I might argue in favor of "sub" for consistency with CF.

sceptre/helpers.py
dboitnot: I'm glad that this has been extracted but I'm a bit queasy about this function mutating collections it doesn't own. I got side-tracked by work but I feel this could be simplified (in the future) using list comprehensions and/or `filter`. But then I wasn't able to dig enough into how it's used to propose a refactor.

dboitnot (commented):
I think this is great and will be a boon to users!

Looking at how resolvers work, though, I'm wondering if a more functional approach might drastically improve the code. There's a lot of mutation and behavior that, while correct, is not obvious and might be harder to reason about or troubleshoot.

It feels like resolvers (and maybe hooks?) could be evaluated as a series of transformations that act on data and their invocation is basically an abstract syntax tree.

The correct reply to the above is, of course, "Put your PR where your mouth is Dan." :)

docs/_source/docs/resolvers.rst
jfalkenstein, line 114:
```suggestion
CloudFormation's ``!Select`` intrinsic function, **except you can use this with negative indices to
select from the end of a list**.
```

docs/_source/docs/resolvers.rst
jfalkenstein: I don't know. I guess I just don't like the abbreviation. It doesn't seem as clear as `substitute`. @zaro0508 , thoughts on this?

sceptre/helpers.py
jfalkenstein: A couple responses:
1. It _does_ own them. That is why there is the cloning functionality of resolvers. The only collections that are mutated are ones that are effectively deep-copied when they are assigned to the stack.
2. This is generally the way resolvers have always worked. The basic idea is that it prevents us from ever needing to resolve the same resolver twice for the same stack. This can be important, since the resolvable attribute might be accessed a _number_ of times over the course of execution.
3. Once we've resolved the resolver, there's no sense in keeping it around; its purpose has been fulfilled.

jfalkenstein:
On the subject of in-place mutation...
1. Because of the cloning employed when a resolver is associated with a stack, it effectively owns the data at that point. The only collections that are mutated are ones that are effectively deep-copied when they are assigned to the stack.
2. This is generally the way resolvers have always worked. The basic idea is that it prevents us from ever needing to resolve the same resolver twice for the same stack. This can be important, since the resolvable attribute might be accessed a _number_ of times over the course of execution.
3. Once we've resolved the resolver, there's no sense in keeping it around; its purpose has been fulfilled.

docs/_source/docs/resolvers.rst
dboitnot: I agree "substitute" has more intrinsic clarity, but since most of our users are going to be pretty familiar with CF and `sub` I think we could avoid the mental dichotomy by sticking to CF's nomenclature. But I would like @zaro0508's thoughts as well.

dboitnot (approved):

docs/_source/docs/resolvers.rst
jfalkenstein: I've changed this to `sub`

quinnjr:
I'm really looking forward to this PR being merged

tests/test_resolvers/test_sub.py
```diff
@@ -0,0 +1,31 @@
...
+class TestSub:
+    def test_resolve__combines_resolvers_into_single_string(self):
+        argument = [
+            "{first} is {first_value}; {second} is {second_value}",
+            {
+                "first": FirstResolver(),
+                "second": SecondResolver(),
+                "first_value": 123,
+                "second_value": 456,
+            },
```
zaro0508, line 26: how about a test when one of the resolvers returns a non-string (list or dict) object?

how about a test with a missing replacement parameter? 
```
        argument = [
            "{first} and {second}",
            {
                "second": "B"
            },
``` 

How about a test where resolver returns `None`?
```
        argument = [
            "{first} and {second}",
            {
                "first": ResolverReturnsNone(),
                "second": "B"
            },
```

tests/test_resolvers/test_join.py
```diff
@@ -0,0 +1,30 @@
...
+    def resolve(self):
+        return "first"
+
+
+class SecondResolver(Resolver):
+    def resolve(self):
+        return "last"
+
+
+class TestJoin:
```
zaro0508, line 17: would it make sense to test when a resolver returns non-string values to join? also a test when resolver returns None?

setup.py
```diff
@@ -71,7 +71,11 @@ def get_version(rel_path):
             "stack_output_external ="
             "sceptre.resolvers.stack_output:StackOutputExternal",
             "no_value = sceptre.resolvers.no_value:NoValue",
+            "select = sceptre.resolvers.select:Select",
             "stack_attr = sceptre.resolvers.stack_attr:StackAttr",
+            "sub = sceptre.resolvers.substitute:Sub",
```
zaro0508, line 76: shouldn't this be `sub = sceptre.resolvers.sub:Sub`?

zaro0508 (changes requested):

docs/_source/docs/resolvers.rst
```diff
@@ -81,6 +105,49 @@ Refer to `sceptre-resolver-cmd <https://github.com/Sceptre/sceptre-resolver-cmd/
...
+The argument for this resolver should be a list with two elements: (1) The delimiter to split on and
+(2) a string to split.
+
+Example:
+
+.. code-block:: yaml
+
+   notifications: !split
+     - ";"
+     - !stack_output my/sns/topics.yaml::SemicolonDelimitedArns
```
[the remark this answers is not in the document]
jfalkenstein: Actually, I don't really. !sub, !join, !select, and !split are ONLY useful when you have resolvers to string together; They're clumsy if you don't. It's better to use Jinja to do these things, if you can. But if you have resolvers whose resolved values you need to manipulate/interpolate/select from, they're really the only way to do that. 

As such, If we don't show how these resolvers can be used with other resolvers, we're not really showing how to use them well.

docs/_source/docs/resolvers.rst
```diff
@@ -181,6 +248,32 @@ Example:
...
+curly-brace templates to indicate variables, and (2) a dictionary where the keys are the format
+string's variable names and the values are the variable values.
+
+Example:
+
+.. code-block:: yaml
+
+   parameters:
+     ConnectionString: !sub
+       - "postgres://{username}:{password}@{hostname}:{port}/{database}"
```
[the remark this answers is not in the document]
jfalkenstein: `!sub` _needs_ a second argument that defines the variable values. It's not the same as CloudFormation's sub that lets you just resolve values without defining them. It's just a glorified `str.format()` invocation. It doesn't have a "scope" outside its own arguments. If you want to resolve something from the Stack scope, you could combine `!sub` and Jinja syntax together, like this:

```python
!sub
  # Note the double-curly braces is JINJA syntax to access variables within Jinja scope
  - "postgres://{{ var.username }}:{password}@{hostname}:{port}/{{ database }}"
  - password: !ssm /my/ssm/password
     hostname: !stack_output my/database/stack.yaml::HostName
     port: !stack_output my/database/stack.yaml::Port
```

Notice how, if it's accessible within Jinja, you don't actually need to provide that to `!sub` as a variable, because Jinja interpolation will happen BEFORE the yaml is rendered. I'm going to look into clarifying this in the docs, since I doubt you'll be the first one confused about this.

docs/_source/docs/stack_config.rst
```diff
@@ -99,7 +99,7 @@ and that Stack need not be added as an explicit dependency.
 
 hooks
 ~~~~~
-* Resolvable: No
+* Resolvable: No (but you can use resolvers _in_ hook arguments!)
```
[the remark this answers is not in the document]
jfalkenstein: Yeah, I'll look into that.

sceptre/resolvers/placeholders.py
```diff
@@ -65,10 +65,7 @@ def _create_explicit_resolver_placeholder(resolver: "resolvers.Resolver") -> str
...
+    return f"{{ {resolver} }}"
```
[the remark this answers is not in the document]
jfalkenstein: Yes! Notice how it puts the resolver's repr in curly braces. It's actually doing something. (The double-curly braces are escaped curly braces in f-strings).

setup.py
jfalkenstein: Yikes! Yes! That's probably why my tests are failing.

tests/test_resolvers/test_join.py
jfalkenstein: Great suggestions. I suppose in the first case, I'll want it to coerce the values to strings. In the second case, they'll just be removed from the list. I'll add those tests!

jfalkenstein: Done!

tests/test_resolvers/test_select.py
```diff
@@ -0,0 +1,33 @@
...
+    def resolve(self):
+        return ["first", "second", "third"]
+
+
+class ItemResolver(Resolver):
+    def resolve(self):
+        return self.argument
+
+
+class TestSplit:
```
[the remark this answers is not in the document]
jfalkenstein: Yes!

tests/test_resolvers/test_split.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class MyResolver(Resolver):
+    def resolve(self):
+        return "first,second,third"
+
+
+class TestSplit:
+    def test_resolve__splits_resolver_value_into_list(self):
+        argument = [",", MyResolver()]
```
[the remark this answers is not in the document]
jfalkenstein: Well, if it's not a string, it'll just blow up. However... it's hard to imagine a scenario where somebody would want to split a value that isn't a string. What sort of behavior would you expect for non-string arguments?

tests/test_resolvers/test_select.py
```diff
@@ -0,0 +1,47 @@
...
+    def resolve(self):
+        return ["first", "second", "third"]
+
+
+class ItemResolver(Resolver):
+    def resolve(self):
+        return self.argument
+
+
+class TestSelect:
```
jfalkenstein, line 17: I need to implement better error handling around missing indexes/keys.

docs/_source/docs/resolvers.rst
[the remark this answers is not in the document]
dboitnot: ¿Por qué no los dos? :)

```yaml
   notifications: !split
     - ";"
     - "note1; note2; note3"
```

Resolves to:

```yaml
   notifications: [ "note1", "note2", "note3" ]
```

As a more practical example, the following will split a semicolon-delimited list exported by another stack:

```yaml
   notifications: !split
     - ";"
     - !stack_output my/sns/topics.yaml::SemicolonDelimitedArns
```

tests/test_resolvers/test_select.py
dboitnot: Pydantic has changed my life:

https://docs.pydantic.dev/

tests/test_resolvers/test_split.py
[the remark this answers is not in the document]
zaro0508: i'm thinking in a case where the resolver returned something that was unexpected, like a list instead of a string?

docs/_source/docs/resolvers.rst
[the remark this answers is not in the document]
jfalkenstein: @dboitnot, sure, that _works_. But If you wanted to split a simple string, you'd just go like this:

```yaml
notifications: {{ "note1; note2; note3".split(";") }}
```

I fear that if we show people examples of using these resolvers in cases where you don't actually _need_ to use them (and where you really _shouldn't_ use them), it will be more confusing. The only times you need to use these resolvers is when you need to combine/manipulate the outputs of other resolvers. I actually added documentation to that effect to make it absolutely clear.

docs/_source/docs/stack_config.rst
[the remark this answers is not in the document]
jfalkenstein: Docs have been added.

tests/test_resolvers/test_split.py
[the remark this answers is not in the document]
jfalkenstein: Additional validation and tests have been added.

tests/test_resolvers/test_sub.py
jfalkenstein: Additional validation and tests have been added.

zaro0508 (changes requested):
generally LGTM.  just a few minor fixes and should be good to go.

docs/_source/docs/hooks.rst
```diff
@@ -224,7 +226,7 @@ Assume a Sceptre `copy` hook that calls the `cp command`_:
        - !copy
            options: "-r"
            source: "from_dir"
-           destination: "to_dir"
+           destination: !stack_output my/other/stack::CopyDestination
```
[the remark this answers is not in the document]
jfalkenstein: No. `!copy` in this imaginary example takes a dict as an argument, not a list. However, `after_update` and `before_update` should be lists. So it's correct in the docs, not in your suggestion.

zaro0508 (approved):

Issue #1318, opened 2023-03-14
Feature to save generate cloudformation templates files


zaro0508:
### Subject of the issue
When debugging we typically use the `generate` command to output the generated cloudformation template files to standard out.    This is fine when generating one or a few templates however it's not as useful when generating a group of templates which can consist of a bunch of templates.  It would be nice to have an option to store all of the generated templates into a folder locally so that further inspection can be applied to each template.  The AWS CDK does this and outputs all templates to a `cdk.out` folder.  Each generated cloudformation template filename would be the name of the stack.

### Expected behaviour
```
sceptre generate --output template-output
```
generated cloudformations template files save in a `template-output` folder.  each filename is the stack name:

template-output
  |
  -- my-vpc.yaml
  -- my-ec2.yaml

jfalkenstein:
That seems like a cool option.

dboitnot:
I have an immediate use for this.

PR #1321, opened 2023-04-10
Write rendered template to a temp file

alexharv074:
In the event that the reader class is unable to parse the rendered config, this adds logic to write the rendered config to a file to assist the caller in understand what went wrong.

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/config/reader.py
+++ b/sceptre/config/reader.py
@@ -14,6 +14,7 @@ import fnmatch
 import logging
 import sys
 import yaml
+import tempfile
 
 from os import environ, path, walk
 from typing import Set, Tuple
@@ -413,6 +414,16 @@ class ConfigReader(object):
 
         return config
 
+    def _write_debug_file(self, content):
+        with tempfile.NamedTemporaryFile(
+            mode="w", delete=False, prefix="rendered_"
+        ) as temp_file:
+            temp_file.write(content)
+            temp_file.flush()
+
+        print(f"Debug file location: {temp_file.name}")
+        return temp_file.name
+
     def _render(self, directory_path, basename, stack_group_config):
         """
         Reads a configuration file, loads the config file as a template
@@ -467,8 +478,11 @@ class ConfigReader(object):
             try:
                 config = yaml.safe_load(rendered_template)
             except Exception as err:
+                debug_file_path = self._write_debug_file(rendered_template)
+
                 raise ValueError(
-                    "Error parsing {}:\n{}".format(abs_directory_path, err)
+                    f"Error parsing {abs_directory_path}:\n{err}\n"
+                    f"Rendered template saved to: {debug_file_path}"
                 )
 
             return config


sceptre/config/reader.py
```diff
@@ -413,6 +414,16 @@ def _recursive_read(
...
         return config
 
+    def _write_debug_file(self, content):
+        with tempfile.NamedTemporaryFile(
+            mode="w", delete=False, prefix="rendered_"
+        ) as temp_file:
+            temp_file.write(content)
+            temp_file.flush()
+
+        print(f"Debug file location: {temp_file.name}")
```
jfalkenstein, line 424: I think that...
1. We should only do this if the `--debug` flag has been passed to the CLI
2. We should log this at the debug level rather than using `print`

alexharv074: @jfalkenstein I have a vague memory of you asking me to do something similar in the past, and I think I did. However, my thinking here is:

1. This message can only ever be printed when Sceptre already bombed out, dumped its stack trace, and gave up. As the caller, I would expect to always be told as much as I need to know here about what went wrong, and how we ended up here! The idea that you could get more info by adding --debug seems a bit counter-intuitive therefore, although I suppose many desperate users will try --debug anyway at this point.
2. We could log to debug level but that's again not really where a stack trace goes? So, it seems a bit unclean to me to have the stack trace message going to the screen and other import info going elsewhere to the logger.

However, if you really want it done this way, I'm happy to—just let me know how!

zaro0508 (changes requested):
write to file only when a template cannot be rendered?  Why not just implement issue #1318  I would much rather be able to save all templates (with or without error).

alexharv074:
> write to file only when a template cannot be rendered? Why not just implement issue #1318 I would much rather be able to save all templates (with or without error).

@zaro0508 Why not both? Why leave Sceptre in a mess like it currently is where it blows up embarrassingly and the user has no idea why and no idea how to fix it, when this small patch I've added allows them to know immediately and exactly what went wrong? Even if we implement your proposed feature, the user may not know about it when it blows up like in the example above. I think all failures should be as helpful as possible.

dboitnot (commented):
I like this idea but I'm not in love with new files being created without me explicitly asking for it. I'd argue for a CLI flag which activates this dumping behavior and, when it's not set, to tell the user something like `Add --dump-template to output the failed template`.

alexharv074:
@dboitnot , @zaro0508 how about I change the behaviour to save the template somewhere if `--output` is set, and if not, print a message, `use --output to save the rendered template`? Then if and when we implement #1318 the two features will work together. (And I'd be happy to implement #1318 too if someone can tell me how to implement it.)

zaro0508:
you have compelling arguments @alexharv074.  This PR is focused on helping debug issues however I'm just not sure this is the right solution.  Why not spend the time to make error messages more clear in general? there are a host of existing sceptre issues regarding sceptre errors not being clear.  Anyways i'll take another look at this PR and give it a fair review.

sceptre/config/reader.py
```diff
@@ -467,8 +478,11 @@ def _render(self, directory_path, basename, stack_group_config):
             try:
                 config = yaml.safe_load(rendered_template)
             except Exception as err:
+                debug_file_path = self._write_debug_file(rendered_template)
+
                 raise ValueError(
-                    "Error parsing {}:\n{}".format(abs_directory_path, err)
+                    f"Error parsing {abs_directory_path}:\n{err}\n"
+                    f"Rendered template saved to: {debug_file_path}"
```
zaro0508, line 485: if the user is processing a config group does sceptre exit on the 1st failure or does it continue to process all templates thus printing this error message multiple times? if it's the latter then we might want to capture all errors.

zaro0508 (changes requested):

sceptre/config/reader.py
```diff
@@ -413,6 +414,16 @@ def _recursive_read(
 
         return config
 
+    def _write_debug_file(self, content):
+        with tempfile.NamedTemporaryFile(
+            mode="w", delete=False, prefix="rendered_"
+        ) as temp_file:
+            temp_file.write(content)
```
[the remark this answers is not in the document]
dboitnot: Just my two cents, but I think writing to a temp file made some sense when the idea is that this would happen automatically. But rather than make ANY default I'd personally prefer if the user specified a path. I propose a new switch: `--dump-config <path>` dedicated to this purpose. It's usage is clear to any user and it's behavior is easy to understand.

Since it's possible we'll be rendering multiple configs, I think @zaro0508 is right about the need for a directory. So rendered configs would go to `<path>/the-stack-name.yaml`.

sceptre/config/reader.py
dboitnot: I think [my suggestion above](https://github.com/Sceptre/sceptre/pull/1321#discussion_r1166827324) would simplify this.

alexharv074: @zaro0508 An exception is raised so it exits on the first failure.

alexharv074:
My thoughts here are still that this feature "should" be merged as is.

I don't personally think it's a good design or user friendly to have an option that only works during an exception. The 'Principle of Least Surprise' would say here that the user who found that Sceptre gave up and raised an exception would expect to either raise a bug next or investigate Sceptre's source code. They won't expect there to be a special option to write the rendered template to a file, so the special option will not be helpful if its existence is surprising. If I implement this, even I will forget that this feature exists in 6 months, and so it probably won't even help me.

The discussion above about relating this to #1318 also missed the point that the implementation _there_—regardless of what it finally looks like—won't resolve the problem _here_ of a rendered template that can't be parsed causing Sceptre to raise an exception.

There is also a suggestion above to write the temporary file to a `sceptre.errors` directory. However, the proposed implementation in #1318 does not support a `sceptre.errors` directory and if I had to change it then it seems to me that this would muck up the otherwise clean folder hierarchy I have proposed. It also results in the temp file ending up in a bit of a funny location so again I don't think this leads to an improvement over the proposal as it is.

Could I humbly request @zaro0508 @dboitnot and @jfalkenstein  have another look.

sceptre/config/reader.py
alexharv074: @dboitnot A path to a temp file is being printed during after the stack trace and just before the program exits.

sceptre/config/reader.py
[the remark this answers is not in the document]
zaro0508: i think `--dump-config <path>` can be done in a follow on PR.

@alexharv074 says errors exit on first failure so i'm fine with writing to the system temp directory

sceptre/config/reader.py
```diff
@@ -467,8 +483,11 @@ def _render(self, directory_path, basename, stack_group_config):
             try:
                 config = yaml.safe_load(rendered_template)
             except Exception as err:
+                debug_file_path = self._write_debug_file(rendered_template)
+
                 raise ValueError(
-                    "Error parsing {}:\n{}".format(abs_directory_path, err)
+                    f"Error parsing {abs_directory_path}:\n{err}\n"
```
zaro0508, line 489: the `abs_directory_path` only provides the path to the folder containing the sceptre configs.  If there are many config files in the folder how would the user know which config and stack failed to render?  Is there a way to tell users the exact config file and stack that failed to render?

zaro0508 (dismissed):
Sorry, i found another small thing that can be improved.

Now i'm wondering about whether we need this functionality at all if we have the dump feature?  if we have #1318 wouldn't the typical workflow be..
1. run some sceptre command
2. template fails to render -> user is told exactly which config and stack failed
3. user uses dump feature to print out the failed rendered template and inspect why it failed

sceptre/config/reader.py
alexharv074: @zaro0508 you can see that above in the screenshot I already provided. The exception already tells us which file it couldn't render; it just doesn't give us any clue about "why". After merging this feature, it will become very easy for the user to figure what went wrong. Compared to now, where it's basically impossible to know.

alexharv074:
@zaro0508  

> Sorry, i found another small thing that can be improved.
> 
> Now i'm wondering about whether we need this functionality at all if we have the dump feature? if we have #1318 wouldn't the typical workflow be..

Nope. As I have said already, an exception raised here (in the `_render` method) prevents any config or template being generated, so there will be nothing to dump. So neither `dump config` nor `dump template` in #1318 will help.

Here's a demo.

1. Check out my 1318 branch:
```
% git symbolic-ref --short HEAD 
alexharvey/1318-part-1
```
2. Break something in the config so that it won't render:
```
# vim s3-bucket-generic/config/s3-bucket-generic.yaml 
```yaml
...
hooks: after_create:
    - !stack_termination_protection 'enabled'
  before_delete:
    - !stack_termination_protection 'disabled'
```
3. see for yourself

Screen Shot 2023-04-22 at 9 27 58 pm

sceptre/config/reader.py
```diff
@@ -467,8 +478,11 @@ def _render(self, directory_path, basename, stack_group_config):
             try:
                 config = yaml.safe_load(rendered_template)
             except Exception as err:
```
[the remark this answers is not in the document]
alexharv074: I definitely think that should be looked at in a separate PR.

sceptre/config/reader.py
```diff
@@ -413,6 +414,16 @@ def _recursive_read(
 
         return config
 
+    def _write_debug_file(self, content):
```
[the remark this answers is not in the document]
alexharv074: done

sceptre/config/reader.py
alexharv074: Oops, now I see your point. That's actually legacy behaviour you are referring to, and not part of my change, however I've now updated it anyway to add the basename. So now you'll see this:
Screen Shot 2023-04-22 at 9 41 40 pm

sceptre/config/reader.py
alexharv074: @jfalkenstein  I belatedly agreed that making this work in debug mode only is correct. I forgot that Sceptre is normally hiding stack traces from us unless `--debug` is specified. So that's done.

sceptre/config/reader.py
```diff
@@ -429,49 +446,64 @@ def _render(self, directory_path, basename, stack_group_config):
...
-                ),
-                "loader": FileSystemLoader(abs_directory_path),
-                "undefined": StrictUndefined,
-            }
-            j2_environment_config = strategies.dict_merge(
-                default_j2_environment_config,
-                stack_group_config.get("j2_environment", {}),
+
+        if not path.isfile(path.join(abs_directory_path, basename)):
+            return
```
alexharv074, line 451: I don't understand what the point of this if statement is or whether this is reachable but it seems clearer to rewrite it this way than have the main logic inside an if branch.

sceptre/config/reader.py
```diff
@@ -429,49 +446,64 @@ def _render(self, directory_path, basename, stack_group_config):
...
-                raise SceptreException(
-                    f"{Path(directory_path, basename).as_posix()} - {err}"
-                ) from err
-
-            try:
-                config = yaml.safe_load(rendered_template)
-            except Exception as err:
-                raise ValueError(
-                    "Error parsing {}:\n{}".format(abs_directory_path, err)
+                message += f"\nTemplating vars saved to: {debug_file_path}"
```
alexharv074, line 489: I spent some time testing the rest of this method, and found that exceptions raised here are just as confusing, therefore the code should also emit debug information in debug mode at this point.

jfalkenstein (changes requested):
This looks good. I'd like to have unit and integration tests on this, though.

alexharv074:
> This looks good. I'd like to have unit and integration tests on this, though.

Ok... I have written some unit tests for `_render` which did not appear to be covered by unit tests before this change. I am thinking, surely we are not going to spend compute and build time on integration testing a line of code that can only be reached during an exception?

tests/test_config_reader.py
```diff
@@ -580,3 +582,82 @@ def test_resolve_node_tag(self):
...
+
+            self.context.project_path = project_path
+            config_reader = ConfigReader(self.context)
+
+            result = config_reader._render("configs", basename, stack_group_config)
+
+            assert result == {"key": "value"}
+
+    def test_render_invalid_jinja_template(self):
+        with self.runner.isolated_filesystem():
```
jfalkenstein, line 620: Huh... I've never seen this `isolated_filesystem` business. Is that a PyTest thing?

tests/test_config_reader.py
```diff
@@ -580,3 +582,82 @@ def test_resolve_node_tag(self):
         new_node = config_reader.resolve_node_tag(mock_loader, mock_node)
 
         assert new_node.tag == "new_tag"
+
+    def test_render_missing_config_file(self):
```
jfalkenstein, line 586: I've found it's more helpful in test names if you follow the format of `test_{method under test}__{situation under test}__{expected behavior or result}`. 

So, for example, in this case: `test_render__missing_config_file__returns_none`. That way when you read the test name, you know what the test is about. And when the test fails, it's absolutely clear what case and result _failed_.

tests/test_config_reader.py
```diff
@@ -580,3 +582,82 @@ def test_resolve_node_tag(self):
...
+            test_config_path = os.path.join(directory_path, basename)
+            test_config_content = "{ key: value"
+            with open(test_config_path, "w") as file:
+                file.write(test_config_content)
+
+            self.context.project_path = project_path
+            config_reader = ConfigReader(self.context)
+
+            with pytest.raises(ValueError, match="Error parsing .*"):
+                config_reader._render("configs", basename, stack_group_config)
```
jfalkenstein, line 663: Why isn't there a test that the rendered file is dumped when it cannot load/render? That seems to be the whole point of this PR, but I don't see a test for that?

jfalkenstein (dismissed):
I'm not seeing a test for the actual functionality you added. I'm grateful for the extra tests... but I think you should add a test that covers the NEW functionality.

dboitnot (commented):
I don't see any issues that others haven't raised.

tests/test_config_reader.py
alexharv074: It's quite cool isn't it? I didn't know about it either but it's just the pattern that the other tests use.

tests/test_config_reader.py
alexharv074: I just feel it is sufficient to prove that the code goes into the exception block, however I've imported glob and a test to prove that debug files are created.

tests/test_config_reader.py
alexharv074: ok done

jfalkenstein (changes requested):
I still want to see integration tests on this to match our coverage on `generate`.

jfalkenstein:
Running integration tests on this now. Assuming they pass, I'll merge.

PR #1323, opened 2023-04-11
[Resolves #894] Switch to using poetry

zaro0508:
The intent here is to change the tooling over to use [poetry](https://python-poetry.org) instead of `setuptools`.
This is building on the work that was started by @tarkatronic in PR #1035

Some important things to note:

* Poetry will be used for packaging, dependencies, and publishing. 
* Poetry is smart enough to automatically include data files such as the `*.json` that had to be manually included by setuptools.
* `twine` is no longer necessary as we can simply use `poetry publish`.
* Poetry handles all virtualenv matters automatically. So we no longer need to instruct users to construct virtualenvs for development purposes, nor do we need to manually create and manage them for CI.
* Dependency specifications are a bit different with Poetry, and a bit more "modern", [as seen here](https://python-poetry.org/docs/dependency-specification/).
* I have told git (via the `.gitattributes`) file to treat the `poetry.lock` file as binary. This is a fairly standard practice, also followed in the JS community with the `package-lock.json` file. This does two primary things:
  * Tells GitHub to not attempt to display a diff for this file
  * Prevents frequent merge conflicts, since this tends to be a frequently updated file

PR #1325, opened 2023-04-15
[Resolve #1318] Implement dump template and write output files

### LINKED ISSUE #1318: Feature to save generate cloudformation templates files

alexharv074:
# Description

Proposed implementation of #1318 

This introduces a `dump template` command that behaves similarly to `generate` but also writes output docs to files.

The behaviour of `dump config` is changed here not to print multi-doc output to screen similar to `dump config`.

The `write` helper function is modified to optionally write to a `file_path`.

Writes files in a naming convention for `dump config` and `dump template` respectively:

```
f".dump/{stack_name}/config.yaml"
f".dump/{stack_name}/template.yaml"
```

# Usage example

## Dump config

```
% sceptre --merge-vars --dir=emr --var-file=/app/sceptre-environment/nonprod/datalake-nonprod/common-env.yaml --var-file=/app/sceptre-environment/common-env.yaml --var-file=/app/sceptre-environment/nonprod/datalake-nonprod/emr/streaming-nonprod-6100/cluster.yaml dump config emr.yaml > /dev/null
```

## Dump template

```
% sceptre --merge-vars --dir=emr --var-file=/app/sceptre-environment/nonprod/datalake-nonprod/common-env.yaml --var-file=/app/sceptre-environment/common-env.yaml --var-file=/app/sceptre-environment/nonprod/datalake-nonprod/emr/streaming-nonprod-6100/cluster.yaml dump template emr.yaml > /dev/null
```

## File structure created

```
% tree .dump/
.dump/
├── streaming-nonprod-6100-efs-pvt
│   ├── config.yaml
│   └── template.yaml
├── streaming-nonprod-6100-emr
│   ├── config.yaml
│   └── template.yaml
├── streaming-nonprod-6100-emr-profile
│   ├── config.yaml
│   └── template.yaml
└── streaming-nonprod-6100-emr-sg-pvt
    ├── config.yaml
    └── template.yaml

4 directories, 8 files
```

## File content

```yaml
% head .dump/streaming-nonprod-6100-emr/config.yaml 
---
dependencies:
- emr-profile.yaml
hooks:
  after_create:
  - !!python/object:hook.stack_termination_protection.StackTerminationProtection
    _argument: enabled
    _argument_is_resolved: false
    logger: &id001 !!python/object/apply:logging.getLogger
    - hook.stack_termination_protection
```

```yaml
% head .dump/streaming-nonprod-6100-emr/template.yaml 
---
AWSTemplateFormatVersion: '2010-09-09'

Parameters:
  MasterInstanceType:
    Type: String
  CoreInstantType:
    Type: String
  CoreInstantCount:
    Type: String
```

## PR Checklist

## Other Information


sceptre/cli/dump.py
```diff
@@ -53,3 +55,55 @@ def dump_config(ctx, path):
...
+@click.argument("path")
+@click.option(
+    "-n",
+    "--no-placeholders",
+    is_flag=True,
+    help="If True, no placeholder values will be supplied for resolvers that cannot be resolved.",
+)
+@click.pass_context
+@catch_exceptions
+def dump_template(ctx, no_placeholders, path):
```
jfalkenstein, line 70: How is this different from `generate`?

jfalkenstein:
I feel like there must be a better way. Is there a way we can fix the `generate` command?

sceptre/cli/dump.py
alexharv074: @jfalkenstein This was what @zaro0508 and I discussed in the long thread you may not have read! The problem with `generate` always has been that it had no way to associate templates with stacks in its output for multiple stacks in a stack group, but rather just randomly orders a bunch of yaml docs that the caller has to figure out. In order to implement #1318 however we need to know which stack goes with which doc. So we agreed to deprecate `generate` and create `dump template` which behaves similarly to `dump config`.

jfalkenstein: Hmmm... I'm not crazy about the idea of deprecating `generate`. Can't we just fix it to make it better?

CONTRIBUTING.md
```diff
@@ -77,11 +77,10 @@ $ git clone git@github.org:<github_username>/sceptre.git
 3. Install Sceptre for development (we recommend you use a
    [virtual environment](http://docs.python-guide.org/en/latest/dev/virtualenvs/))
 
+A convenient script is also provided:
```
zaro0508, line 80: no, we are going to use poetry and the poetry CLI to manage dependencies very soon, check PR #1323

zaro0508 (changes requested):

sceptre/cli/dump.py
dboitnot: Yeah, I use the heck out of generate.

dboitnot:
> I feel like there must be a better way. Is there a way we can fix the `generate` command?

Maybe a flag like `--dump` or something to the `generate` sub-command?

alexharv074:
> I feel like there must be a better way. Is there a way we can fix the `generate` command?

@jfalkenstein Well if we `fixed` the generate command (we could copy this functionality there instead), we are then left with:
1. A breaking change
2. A seriously ugly inconsistency in the naming between `dump config` and `generate`. 

I think this is actually quite clean and a big improvement myself.

CONTRIBUTING.md
alexharv074: Will remove if the other discussions are resolved.

alexharv074:
> > I feel like there must be a better way. Is there a way we can fix the `generate` command?
> 
> Maybe a flag like `--dump` or something to the `generate` sub-command?

Yes, I assume we will probably make this behaviour optional via some sort of flag in the final version.

sceptre/cli/dump.py
alexharv074: Be sure to consider all this in the context of how the new naming allows such a sane and easy to understand structure for the output templates to emerge https://github.com/alexharv074/sceptre/pull/3 

I think it is unavoidable that if we keep the existing naming (`dump config` which was named by Jon and I after a _lot_ of discussion(!) vis-a-vis `generate`) the output templates naming is going to be a dog's breakfast.

sceptre/cli/dump.py
```diff
@@ -53,3 +55,55 @@ def dump_config(ctx, path):
...
+        else:
+            output.append({stack.external_name: template})
+
+    output_format = "json" if context.output_format == "json" else "yaml"
+
+    if len(output) == 1:
+        write(output[0][stack.external_name], output_format)
+    else:
+        for template in output:
+            write(template, output_format)
```
jfalkenstein, line 109: It strikes me as odd that the output format is fundamentally different if it's just one stack vs. if it's numerous stacks.

sceptre/cli/dump.py
```diff
@@ -53,3 +55,55 @@ def dump_config(ctx, path):
...
+    )
+    with execution_context:
+        responses = plan.dump_template()
+
+    output = []
+    for stack, template in responses.items():
+        if template is None:
+            logger.warning(f"{stack.external_name} does not exist")
+        else:
+            output.append({stack.external_name: template})
```
jfalkenstein, line 101: Why a list of dicts and not just one big dict?

sceptre/plan/actions.py
```diff
@@ -1152,3 +1152,10 @@ def dump_config(self, config_reader: ConfigReader):
         """
         stack_path = normalise_path(self.stack.name + ".yaml")
         return config_reader.read(stack_path)
+
+    @add_stack_hooks
+    def dump_template(self):
+        """
+        Returns the Template for the Stack
+        """
+        return self.stack.template.body
```
jfalkenstein, line 1161: This is identical to `generate()`.

venv.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
```
jfalkenstein, line 1: This file doesn't belong in this PR.

jfalkenstein (commented):
I think I'm just very concerned about adding a completely new command that is ALMOST identical to the old command except it outputs in a slightly different format. I really don't think this is the right way to proceed.

sceptre/cli/dump.py
alexharv074: It makes sense if you use it, and, again, is what you and I agreed on already after lengthy discussions for `dump config`. In the case of a single stack—most of the time—you don't want to see the stack names in this output; you just want the yaml output. In the case of a stack group however, you need to know which stack goes with which doc.

HOWEVER, please note that in the implementation of Khai's feature, I have changed this back to multi-doc yaml output without the stack names. That's because the .dump structure that is created allows us to know which stack goes with which doc.

sceptre/plan/actions.py
alexharv074: Yes, it needs to be identical to `generate()` since the latter is deprecated and would be removed. Also, the naming standard requires that CLI commands have a method named after the CLI command (at least I hope we preserve and continue that naming standard!).

sceptre/cli/dump.py
alexharv074: Again. For consistency with `dump config`. I can't remember to be honest why we did it this way.

alexharv074:
> I think I'm just very concerned about adding a completely new command that is ALMOST identical to the old command except it outputs in a slightly different format. I really don't think this is the right way to proceed.

I guess you would need to say then what the right way to proceed is then, because I can't personally see any other option. The legacy `sceptre generate` feature is a mess though. It always has been very broken that it doesn't tell you the stack name and dumps a randomly ordered list of yaml docs.

sceptre/cli/helpers.py
```diff
@@ -78,6 +78,8 @@ def write(var, output_format="json", no_colour=True):
     :type output_format: str
     :param no_colour: Whether to colour stack statuses
     :type no_colour: bool
+    :param file_path: Optional path to a file to save the output
+    :type file_path: str, optional
```
alexharv074, line 82: Yes these are Pathlib paths not strings. Will fix.

sceptre/plan/actions.py
alexharv074: If we agree to alias `generate` and `dump template` of course then perhaps we only need one method here.

sceptre/cli/dump.py
alexharv074: Closing this because it's not relevant after I merged Part 2 into here.

zaro0508 (dismissed):

sceptre/cli/dump.py
```diff
@@ -48,8 +52,70 @@ def dump_config(ctx, path):
...
+        file_path = Path(f".dump/{stack_name}/config.yaml")
+        write(config[stack_name], output_format, no_colour=True, file_path=file_path)
+
     if len(output) == 1:
-        write(output[0][stack.external_name], output_format)
+        process_config(output[0])
     else:
         for config in output:
-            write(config, output_format)
+            process_config(config)
```
jfalkenstein, line 64: I don't see a reason any more to have this `if/else` here. I think you could just have the for loop.

sceptre/cli/dump.py
```diff
@@ -48,8 +52,70 @@ def dump_config(ctx, path):
...
+
+
+@dump_group.command(name="template")
```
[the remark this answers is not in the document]
jfalkenstein: I think I disagree. Sometimes (maybe even most of the time) we'd want to dump the template. But sometimes we'll want to dump the stack-config. 

But I do agree that we shouldn't write (at least by default) both to files and console. I think (in line with every other Sceptre command), we should default to console output. However, you could use `--file` to output to file.

sceptre/cli/dump.py
```diff
@@ -48,8 +52,70 @@ def dump_config(ctx, path):
...
+    execution_context = (
+        null_context() if no_placeholders else use_resolver_placeholders_on_error()
+    )
+    with execution_context:
+        responses = plan.dump_template()
+
+    output = []
+    for stack, template in responses.items():
+        if template is None:
+            logger.warning(f"{stack.external_name} does not exist")
```
jfalkenstein, line 106: I don't understand the case where `template` would be `None`...

sceptre/cli/dump.py
jfalkenstein: That feels a bit odd to me

sceptre/cli/dump.py
```diff
@@ -48,8 +52,70 @@ def dump_config(ctx, path):
...
+
+
+@dump_group.command(name="template")
+@click.argument("path")
+@click.option(
+    "-n",
+    "--no-placeholders",
+    is_flag=True,
+    help="If True, no placeholder values will be supplied for resolvers that cannot be resolved.",
+)
```
jfalkenstein, line 74: I think there ought to be an option, like `--only` that will exclusively output the template at the indicated path and not include any other configs passed along. Probably the same for `dump config` as well.

sceptre/cli/dump.py
```diff
@@ -48,8 +52,70 @@ def dump_config(ctx, path):
...
+    def process_template(template):
+        stack_name = list(template.keys())[0]
+        file_path = Path(f".dump/{stack_name}/template.yaml")
+        write(template[stack_name], output_format, no_colour=True, file_path=file_path)
+
+    if len(output) == 1:
+        process_template(output[0])
+    else:
+        for template in output:
+            process_template(template)
```
jfalkenstein, line 121: Same comment here... I think we could just have the for loop.

sceptre/cli/helpers.py
```diff
@@ -93,6 +95,13 @@ def write(var, output_format="json", no_colour=True):
 
     click.echo(output)
 
+    if file_path:
+        dir_path = file_path.parent
```
jfalkenstein, line 99: You're still treating a string like a pathlib.Path object here.

sceptre/cli/helpers.py
```diff
@@ -93,6 +95,13 @@ def write(var, output_format="json", no_colour=True):
 
     click.echo(output)
```
jfalkenstein, line 96: Echo has a `file` argument that you can use to write the it to a file. https://click.palletsprojects.com/en/8.1.x/api/#click.echo. It defaults to stdout. So you could basically do:
* If `file_path` is passed, make the `file` argument a file object at that path
* If it is not passed, make the `file` argument `sys.stdout`

sceptre/plan/actions.py
jfalkenstein: Yeah, I would say that you should make `generate` here call this method and return its result. That should be a pretty harmless way to start aliasing.

sceptre/plan/plan.py
```diff
@@ -440,3 +440,13 @@ def dump_config(self, *args):
...
         return self._execute(self.config_reader, *args)
+
+    def dump_template(self, *args):
+        """
+        Returns a generated Template for a given Stack
+
+        :returns: A dictionary of Stacks and their template body.
+        :rtype: dict
+        """
+        self.resolve(command=self.generate.__name__)
```
jfalkenstein, line 451: Why is this calling `generate`? I think that might be a mistake.

jfalkenstein (changes requested):
I've left some comments. We'll still need to update docs and properly deprecate `generate`.

sceptre/cli/dump.py
alexharv074: I _think_ it may be to work around or hack the limitations of the `write` helper. If we need to change `dump config`, perhaps this should be looked at in a separate issue/PR?

sceptre/cli/helpers.py
alexharv074: Sorry, can you elaborate? I am not seeing the problem there.

alexharv074: I've updated the type hints to clarify that a pathlib.Path is expected here.

sceptre/plan/actions.py
alexharv074: ok done

sceptre/plan/plan.py
alexharv074: fixed.

sceptre/cli/dump.py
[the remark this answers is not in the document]
alexharv074: I disagree strongly. I never, ever want both, and a change like this would seriously break all our tooling too.

sceptre/cli/dump.py
alexharv074: Yes possibly, but could we do that in a separate change later?

sceptre/cli/dump.py
alexharv074: As hinted in the following line, a `None` would be returned in the case of the stack which hasn't been created or for some reason doesn't exist. I have a vague memory but probably can't remember how to actually reproduce the scenario that led to the if statement.

sceptre/cli/dump.py
alexharv074: Same response here. Unless I'm dreaming these two branches do different things:

```python
    if len(output) == 1:
        process_template(output[0])
    else:
        for template in output:
            process_template(template)                                                                                                                                                                                            
```

sceptre/cli/dump.py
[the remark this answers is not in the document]
alexharv074: So, on thinking about the second part, the problem I see is this. If the command does not always dump to the screen _and_ a file, then we arrive back at the problem that the default behaviour of printing to the screen should also by defaulting tell the user about the stack name. If I do that, then @jfalkenstein 's wish that we can simply redirect a multi-doc yaml to a file is not satisfied. I also like the output better this way, and the dumped files can tell me which template goes with which stack. 

Thoughts?

sceptre/cli/dump.py
jfalkenstein: My only concern is this: Let's say I need to pipe a template over to a linter or awscli for some reason because it has jinja and other stuff in it. I want to do something like: `sceptre dump template my/sceptre/config.yaml | cfn-lint -`. When you're outputting the templates inside a larger data structure like this, that's not a valid template. 

I understand your concern of wanting to be able to associate the template with a stack, so there are a few options I can think of:

Option 1: Rather than this, you could output the stack name via a logger rather than echoing it. That will result it going to stderr rather than stdout. That way, if you pipe output to another command or use `>`, it won't actually get included. So you could alternate by logging the stack name and then writing the output.

Option 2: Have a `--stack-names` flag that optionally makes it output this way but by default it does not.

sceptre/cli/dump.py
```diff
@@ -48,8 +52,70 @@ def dump_config(ctx, path):
...
+        if template is None:
+            logger.warning(f"{stack.external_name} does not exist")
+        else:
+            output.append({stack.external_name: template})
+
+    output_format = "json" if context.output_format == "json" else "yaml"
+
+    def process_template(template):
+        stack_name = list(template.keys())[0]
+        file_path = Path(f".dump/{stack_name}/template.yaml")
```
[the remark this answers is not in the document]
jfalkenstein: @alexharv074 , to my point elsewhere, it seems you might not realize that logs are configured to go to stderr in Sceptre (it's actually the default for Python), so logs wouldn't interfere with using `>` because that sends stdout. They would only get sucked up if you used `$>`.

sceptre/cli/dump.py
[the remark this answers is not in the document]
jfalkenstein: I don't think we ought to dump to a file by default. If there's a `--file` option, though... I don't really care if it goes to the console. I think it would be cleaner if what goes to the console was just "Dumped contents were output to {file path}", though.

sceptre/cli/dump.py
jfalkenstein: `process_config(output[0])` is the same as doing this: 

```python
for config in output[:1]:
    process_config(output)
```

And if output is a list of only one item, it's the same as doing this, because it'll only loop once:

```python
for config in output:
    process_config(output)
```

Because remember, looping through a list like this is essentially the same as doing:

```python
for i in range(len(output)):
    process_config(output[i])
```

But if that's the flow you want for >1 as well, you can just use that for output, no matter if there are 0 items (it will loop zero times), 1 items (it will loop once), or 100 items. 

Anyway, it's all the same. You don't need the if/else.

sceptre/cli/dump.py
jfalkenstein: That's not true. You can absolutely generate a template when a stack isn't created. That's how the diff command works when there's no stack deployed yet.

sceptre/cli/dump.py
jfalkenstein: That's a pretty simple change to make, IMO. But yes, it could be done later.

sceptre/cli/dump.py
jfalkenstein: You're indeed dreaming!

sceptre/cli/dump.py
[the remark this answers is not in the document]
jfalkenstein: Also, I might not be right about this, but I suspect this path isn't Windows-safe. You could instead do one of the following to guarantee it works with Windows:
```python
Path(".dump", stack_name, "template.yaml")
# or
Path(".dump") / stack_name / "template.yaml")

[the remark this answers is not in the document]
Comment (jfalkenstein) [reply sceptre/cli/dump.py:114]: Also, the template might be json, not yaml.

File: sceptre/plan/actions.py (hunk 5 above)
Comment (jfalkenstein) [reply sceptre/plan/actions.py:1161]: If you're deprecating `generate`, you definitely should do that correctly, both in docs and using the `deprecated` decorator.

File: sceptre/cli/dump.py (hunk 4 above)
Comment (alexharv074) [reply sceptre/cli/dump.py:108]: @jfalkenstein you are misunderstanding the implementation and behaviour. Here's a demo, showing a yaml stream that can be parsed by external tools is always emitted, even in the case of multiple stacks:

Screen Shot 2023-04-21 at 8 20 03 pm

File: sceptre/plan/actions.py (hunk 5 above)
Comment (alexharv074) [reply sceptre/plan/actions.py:1166]: ok done

File: sceptre/cli/dump.py (hunk 8 above)
Comment (alexharv074) [reply sceptre/cli/dump.py:64]: @jfalkenstein ah! I see your point. They used to do different things and now they don't. Updated.

File: sceptre/cli/dump.py (hunk 10 above)
Comment (alexharv074) [reply sceptre/cli/dump.py:106]: Ah you know what, it came from the `fetch_remote_template` command. I may have copy/pasted that into the dump config command in error. Ok, I've removed it in both `dump config` and `dump template`.

File: sceptre/cli/helpers.py (hunk 14 above)
Comment (alexharv074) [reply sceptre/cli/helpers.py:101]: Ok, I see the improvement in that, done.

File: sceptre/cli/helpers.py (hunk 7 above)
Comment (alexharv074) [reply sceptre/cli/helpers.py]: fixed

File: sceptre/cli/dump.py (hunk 16 above)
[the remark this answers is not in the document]
Comment (alexharv074) [reply sceptre/cli/dump.py]: Ok fixed fixed and fixed.

File: sceptre/cli/helpers.py [hunk 17]
```diff
@@ -173,6 +189,23 @@ def _generate_text(stream):
     return stream
 
 
+def process_template(
```
jfalkenstein, line 192: While `process_template` as a nested function inside your other functions was a perfectly acceptable name, since you would contextually understand what "processing" meant. However, since you've extracted this function out and put it in `helpers`, it isn't obvious at all what "processing" means. I think we should have a better name on this function that clearly describes what it is supposed to do. Maybe `dump_dict_to_file`?

sceptre/cli/helpers.py
```diff
@@ -173,6 +189,23 @@ def _generate_text(stream):
...
+    template: dict, output_format: str, type: str = "template"
```
jfalkenstein, line 193: It's generally discouraged to shadow builtins like `type`.

sceptre/cli/helpers.py
```diff
@@ -173,6 +189,23 @@ def _generate_text(stream):
...
+
+    :param template: the template.
+    :param output_format: The format to print the output as. Allowed values: \
+    :param type: either template or config used only in the file name.
+    """
+    stack_name = list(template.keys())[0]
+    file_path = Path(".dump") / stack_name / f"{type}.{output_format}"
+    logger.info(f"{stack_name} dumping {type} to {file_path}")
+    write(template[stack_name], output_format, no_colour=True, file_path=file_path)
+    logger.info(f"{stack_name} done.")
```
jfalkenstein, line 206: I don't think we need both of these log messages, at least not both at the `info` level. 

I think having the initial message would be very helpful in the case that writing fails for some reason (not sure why that would be the case, though...). So I think the first log message should be emitted at the `debug` level rather than `info` and the second log message should include the file_path as well and be something like: `f'{stack_name} dumped to file at {file_path}'`.

sceptre/plan/actions.py
```diff
@@ -624,12 +628,13 @@ def _convert_to_url(self, summaries):
 
         return new_summaries
 
+    @deprecation.deprecated("4.0.0", "6.0.0", __version__, "Use dump template instead.")
     @add_stack_hooks
```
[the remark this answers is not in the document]
jfalkenstein: Also, add a `fail_if_not_removed` test on this as well.

sceptre/plan/plan.py
```diff
@@ -440,3 +440,13 @@ def dump_config(self, *args):
         """
         self.resolve(command=self.dump_config.__name__)
         return self._execute(self.config_reader, *args)
+
```
jfalkenstein, line 443: You ought to deprecate `generate` above as well.

sceptre/cli/dump.py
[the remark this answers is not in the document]
dboitnot: I agree that console output should be the default and I agree that a multi-doc YAML output makes sense. 

An alternative would be to pipe the rendered content to STDIN and meta information like the stack name to STDERR.

sceptre/cli/helpers.py
alexharv074: @jfalkenstein the message is proposed by @zaro0508. I would add that without a second logger message after dump has completed, the first will be probably not noticed.

sceptre/cli/helpers.py
alexharv074: fixed

sceptre/cli/helpers.py
alexharv074: fixed

sceptre/plan/plan.py
alexharv074: done

sceptre/cli/helpers.py
alexharv074: have a look now.

sceptre/cli/helpers.py
```diff
@@ -173,6 +189,23 @@ def _generate_text(stream):
     return stream
 
 
+def process_template(
+    template: dict, output_format: str, type: str = "template"
+) -> None:
+    """
+    Helper to write templates and configs used in dump commands.
+
+    :param template: the template.
```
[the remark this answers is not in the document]
alexharv074: You'll have to elaborate ....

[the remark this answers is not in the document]
jfalkenstein: Your `template` variable here isn't actually a template. In other words, your variable naming lies. For a helper function that's meant to stand on its own, the interface must be clear, simple, and functional. But you're making a lot of assumptions here about what `template` is. 

It should take a `stack_name` and a `template_dict` parameter instead of a dict that happens to have a key that is the stack name and a value that is the actual template dict. Your code would be simpler, more readable, and easier to interact with if you did that.

jfalkenstein (dismissed):
I know you won't like this, but `generate` has integration tests and I believe that we should have integration tests on these dump commands as well.

dboitnot (approved):

alexharv074:
> I know you won't like this, but `generate` has integration tests and I believe that we should have integration tests on these dump commands as well.

I've refactored the `generate` test to instead call `dump_template`. As for the other dump commands, they are on a different branch so will have to go in different PRs and this one needs to be merged first.

sceptre/cli/helpers.py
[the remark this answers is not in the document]
alexharv074: I'm not convinced that a variable named `template_dict` of type `dict` is clearer than `template: dict` and to my mind the whole comes out more rather than less messy afterwards. Perhaps we are better off without this helper at all and I revert to how it was before?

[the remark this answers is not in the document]
alexharv074: For example on the caller side:

```python
    for template in output:
        dump_to_file(list(template.keys())[0], template, output_format)
```

We don't want duplication of `list(template.keys())[0]` all over the place for the sake helping the reader of that helper function since most people are going to be reading the CLI functions; not the helpers.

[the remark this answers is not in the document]
jfalkenstein: My point is that you're calling the dict a "template", but it's actually a dict with a single key indicating the stack name and a its value is the actual template. In other words, your variable name is wrong and the interface of the function is more challenging than it ought to be. 

Rather than assuming this arbitrary structure that is a mouthful to describe, I'm just saying take two parameters instead of this complex dict structure that is isn't well-typed.

alexharv074:
@jfalkenstein  have a look now. Here's what I changed:
- implemented `--to-file` making the default behaviour not to create the dump files.
- refactored the process files logic you didn't like moving it all back the dump commands.

sceptre/cli/dump.py
```diff
@@ -18,9 +22,12 @@ def dump_group():
 
 @dump_group.command(name="config")
 @click.argument("path")
+@click.option(
+    "--to-file", is_flag=True, help="If True, also dump the template to a local file."
```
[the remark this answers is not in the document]
jfalkenstein: Never mind. I see what we're doing now.

zaro0508 (changes requested):
`generate` is found in a few locations in the docs.  if we are deprecating that command in this same PR then could we also update to dump in the docs as well? 

I think it might be cleaner to deprecate the generate command in a follow on PR.

sceptre/cli/dump.py
```diff
@@ -39,17 +46,91 @@ def dump_config(ctx, path):
...
+            write(template, output_format, no_colour=True, file_path=file_path)
+            logger.info(f"{stack_name} dump to {file_path} complete.")
+
+        else:
+            write(template, output_format, no_colour=True)
+
+
+@dump_group.command(name="all")
+@click.argument("path")
+@click.option(
```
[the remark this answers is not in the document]
alexharv074: not intentional, fixed.

alexharv074:
> `generate` is found in a few locations in the docs. if we are deprecating that command in this same PR then could we also update to dump in the docs as well?
> 
> I think it might be cleaner to deprecate the generate command in a follow on PR.

A few hits on the word when I `grep -rw generate`, however those are not hits on the subcommand. Are you sure?

sceptre/cli/helpers.py
```diff
@@ -93,6 +102,13 @@ def write(var, output_format="json", no_colour=True):
 
     click.echo(output)
```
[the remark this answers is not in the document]
alexharv074: Ok how about now? (I'm not sure whether people want the text to go to both the screen and file if file_path is specified).

zaro0508:
> A few hits on the word when I grep -rw generate, however those are not hits on the subcommand. Are you sure?

This is what i see 

```
docs/_source/docs/hooks.rst:``before_generate`` or ``after_generate`` - run hook before or after generating stack template.
docs/_source/docs/resolvers.rst:  ``!stack_output other_stack.yaml::OutputName`` and you run the ``dump template`` command, the generated
docs/_source/docs/templates.rst:to generate CloudFormation Template as a `json` string.
docs/_source/docs/templates.rst:  To generate templates using Troposphere you must install the
docs/_source/docs/templates.rst:AWS CDK can be used to programmatically generate templates for your stacks and then Sceptre can
```
I think those references should be changed to the `dump` command


Also didn't we decide to redirect the generate to alias the dump command?  if so it looks like it wasn't done in template.py
```
sceptre/cli/template.py:@click.command(name="generate", short_help="Prints the template.")
sceptre/cli/template.py:def generate_command(ctx, no_placeholders, path):
```

alexharv074:
> This is what i see
> 
> ```
> docs/_source/docs/hooks.rst:``before_generate`` or ``after_generate`` - run hook before or after generating stack template.
> docs/_source/docs/resolvers.rst:  ``!stack_output other_stack.yaml::OutputName`` and you run the ``dump template`` command, the generated
> docs/_source/docs/templates.rst:to generate CloudFormation Template as a `json` string.
> docs/_source/docs/templates.rst:  To generate templates using Troposphere you must install the
> docs/_source/docs/templates.rst:AWS CDK can be used to programmatically generate templates for your stacks and then Sceptre can
> ```
> 
> I think those references should be changed to the `dump` command

@zaro0508 I didn't actually know about the `before_generate` and `after_generate` hooks but I don't see how those can be renamed. `before_dump_template` is both an ugly name and does not make sense in that context. It would have to stay as `before_generate` I think. Remember, it doesn't mean "before you run the generate command" ; it means "before Sceptre generates its template". Generating the template internally, and dumping it out for the caller to look at, are different things.

alexharv074:
> Also didn't we decide to alias the generate to the dump command? if so, it looks like it wasn't done in template.py
> 
> ```
> sceptre/cli/template.py:@click.command(name="generate", short_help="Prints the template.")
> sceptre/cli/template.py:def generate_command(ctx, no_placeholders, path):
> ```

Yes but it was aliased not in the CLI but in the API which is I think the way it should be.

zaro0508:
> Yes but it was aliased not in the CLI but in the API which is I think the way it should be.

What is your reasoning for not aliasing it in the CLI as well?

alexharv074:
> > Yes but it was aliased not in the CLI but in the API which is I think the way it should be.
> 
> What is your reasoning for not aliasing it in the CLI as well?

Well, aliasing in both places will just result in unused code paths. If I point the generate CLI command at the dump_template API, why not just remove the generate API since it is no longer used? 

In the end I don't really mind how it is done or necessarily see the point in aliasing this. The implementation above has the advantage that it seems to be the one that @jfalkenstein agreed to and not to mention that it's the one that I already implemented!

zaro0508 (approved):

PR #1328, opened 2023-04-20
[Resolve #1324] Bugfix for dump config and refactor ConfigReader

### LINKED ISSUE #1324: The dump config command is not loading the j2_environment
### Subject of the issue
Given a stack with custom `j2_environment`, this is not loaded or available during the `dump config` command.

### Your environment
* version of sceptre - latest
* version of python - 3.10
* which OS/distro - Debian

### Steps to reproduce

Create a stack that has a custom Jinja filter called in its config. Try dump config. A message like this will be seen:
```
"No filter named '<some_filter>'."
```
Try `sceptre generate`; that should work fine.

alexharv074:
This is a substantial refactoring of the ConfigReader class to fix the
broken `dump config` command.
    
Changes here include:
  
- Change public `read()` method to expect a single input being the rel_path of the yaml file.  
- change `stack_group_config` from a second arg to `read()` to be a class instance variable instead, which is set up during class initialisation in `__init__() to ensure it is properly shared.
- Clean up some unused code.

TODO.

- I have removed a problematic unit test pending expected detailed discussion about how to test these changes!

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/config/reader.py
+++ b/sceptre/config/reader.py
@@ -374,7 +374,7 @@ class ConfigReader(object):
         return config
 
     def _recursive_read(
-        self, directory_path: str, filename: str, work_in_progress_config: dict
+        self, directory_path: str, filename: str, config: dict
     ) -> dict:
         """
         Traverses the directory_path, from top to bottom, reading in all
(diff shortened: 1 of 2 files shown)


sceptre/config/reader.py
```diff
@@ -102,14 +102,6 @@
...
 )
 
-INTERNAL_CONFIG_ATTRIBUTES = ConfigAttributes(
-    {
-        "project_path",
-        "stack_group_path",
-    },
-    {},
-)
-
```
alexharv074, line 112: This appears to be unused.

sceptre/config/reader.py
```diff
@@ -342,12 +346,10 @@ def read(self, rel_path, base_config=None):
         # Adding properties from class
         config = {
             "project_path": self.context.project_path,
-            "stack_group_path": directory_path,
```
alexharv074, line 345: As far as I could tell, the only place this `stack_group_path` was being used was in the unit tests. So I've removed it.

sceptre/config/reader.py
```diff
@@ -422,8 +419,8 @@ def _render(self, directory_path, basename, stack_group_config):
         :type directory_path: str
         :param basename: The filename of the config file
         :type basename: str
-        :param stack_group_config: The loaded config file for the StackGroup
-        :type stack_group_config: dict
+        :param config_group: The loaded config file for the StackGroup
+        :type config_group: dict
```
alexharv074, line 423: I've done a bit of renaming of variables to ensure that `stack_group_config` is used only once.

sceptre/config/reader.py
```diff
@@ -141,6 +133,42 @@ def __init__(self, context):
 
         self.templating_vars = {"var": self.context.user_variables}
 
+        # Set up root
```
jfalkenstein, line 136: I don't like the idea of the `__init__` doing this much. It's generally a bad pattern for a class' `__init__` method to actually _do_ things. It should only set the class up to do things. This sort of pattern makes the code harder to test, harder to maintain, and less flexible, because you simply cannot instantiate a ConfigReader without all this stuff happening. I'd rather the `__init__` do LESS rather than all this.

jfalkenstein (changes requested):
This is a pretty significant refactor, even if it feels like you're just moving code around, in terms of how all these things operate. I think that we definitely need to refactor the ConfigReader, but I don't think this is the way. Perhaps we can discuss the best approach on #1324?

PR #1333, opened 2023-04-21
[Resolve #1324] Reimplement dump config

### LINKED ISSUE #1324: The dump config command is not loading the j2_environment

alexharv074:
After it was discovered that ConfigReader.read() is a private method, an alternative implementation of dump config is proposed.

Here, we make the following changes:

- Add an attribute `config` to the Stack class which receives the config from the reader.

- The dump config action in the plan is updated to simply return the config attribute.

- For clarify, the read method in ConfigReader is renamed _read to make it clearer that this is a private method.

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/config/reader.py
+++ b/sceptre/config/reader.py
@@ -102,14 +102,6 @@ STACK_CONFIG_ATTRIBUTES = ConfigAttributes(
     },
 )
 
-INTERNAL_CONFIG_ATTRIBUTES = ConfigAttributes(
-    {
-        "project_path",
-        "stack_group_path",
-    },
-    {},
-)
-
 REQUIRED_KEYS = STACK_GROUP_CONFIG_ATTRIBUTES.required.union(
     STACK_CONFIG_ATTRIBUTES.required
 )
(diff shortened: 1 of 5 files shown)


sceptre/stack.py
```diff
@@ -194,6 +194,7 @@ def __init__(
         ignore=False,
         obsolete=False,
         stack_group_config: dict = {},
+        config: dict = {},
```
jfalkenstein, line 197: It's a very bad practice to have mutable default arguments in Python. I know `stack_group_config` is doing it and that shouldn't be doing it either. The better approach is to set the value to `None` and then do like you're doing below in line 215.

sceptre/config/reader.py
```diff
@@ -102,14 +102,6 @@
...
 )
 
-INTERNAL_CONFIG_ATTRIBUTES = ConfigAttributes(
-    {
-        "project_path",
-        "stack_group_path",
-    },
-    {},
-)
-
```
jfalkenstein, line 112: Why have you removed this?

alexharv074: @jfalkenstein  Because it's not used.

jfalkenstein (approved):

jfalkenstein:
Running integration tests on this now...

PR #1337, opened 2023-04-30
[Resolve #1336] Add bugfix in Stack Differ

### LINKED ISSUE #1336: Stack differ raises if a value is None
### Subject of the issue
The Stack Differ raises an exception if a parameter value is None.

### Your environment
* version of sceptre - master branch
* version of python (python --version) - 3.10
* which OS/distro - Ubuntu

### Steps to reproduce
Create a config like this:

```yaml
parameters:
  GlueDatabase:
```

Run 
```
sceptre diff -t difflib myapp.yaml
```

### Expected behaviour
This should work.

### Actual behaviour

```
  File "/usr/local/bin/sceptre", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/click/decorators.py", line 26, in new_func
  File "/app/sceptre/sceptre/cli/helpers.py", line 43, in decorated
    return func(*args, **kwargs)
  File "/app/sceptre/sceptre/cli/diff.py", line 140, in diff_command
    diffs: Dict[Stack, StackDiff] = plan.diff(stack_differ)
  File "/app/sceptre/sceptre/plan/plan.py", line 417, in diff
    return self._execute(*args)
  File "/app/sceptre/sceptre/plan/plan.py", line 29, in wrapped
    return func(self, *args, **kwargs)
  File "/app/sceptre/sceptre/plan/plan.py", line 55, in _execute
  File "/app/sceptre/sceptre/plan/executor.py", line 52, in execute
  File "/usr/local/lib/python3.10/concurrent/futures/_base.py", line 439, in result
    return self.__get_result()
  File "/usr/local/lib/python3.10/concurrent/futures/_base.py", line 391, in __get_result
    raise self._exception
  File "/usr/local/lib/python3.10/concurrent/futures/thread.py", line 58, in run
  File "/app/sceptre/sceptre/plan/executor.py", line 59, in _execute
  File "/app/sceptre/sceptre/hooks/__init__.py", line 95, in decorated
  File "/app/sceptre/sceptre/plan/actions.py", line 997, in diff
    return stack_differ.diff(self)
  File "/app/sceptre/sceptre/diffing/stack_differ.py", line 122, in diff
    generated_config = self._create_generated_config(stack_actions.stack)
  File "/app/sceptre/sceptre/diffing/stack_differ.py", line 150, in _create_generated_config
    parameters = self._extract_parameters_from_generated_stack(stack)
  File "/app/sceptre/sceptre/diffing/stack_differ.py", line 172, in _extract_parameters_from_generated_stack
    formatted_parameters[key] = value.rstrip("\n")
AttributeError: 'NoneType' object has no attribute 'rstrip'
```

alexharv074:
In the case of a parameter dictionary that contains nulls (None) the Boto3 library has been found to take care of converting None to "". This means that launching such a stack in CloudFormation with a parameter dict containing params set to None works fine, whereas this differ raised an exception when it tried to strip newlines.

This patch stops newlines being stripped in the case of a parameter that is set to None.

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/diffing/stack_differ.py
+++ b/sceptre/diffing/stack_differ.py
@@ -169,7 +169,10 @@ class StackDiffer(Generic[DiffType]):
         for key, value in stack.parameters.items():
             if isinstance(value, list):
                 value = ",".join(item.rstrip("\n") for item in value)
-            formatted_parameters[key] = value.rstrip("\n")
+            if isinstance(value, str):
+                formatted_parameters[key] = value.rstrip("\n")
+            elif value is None:
+                formatted_parameters[key] = value
 
         return formatted_parameters
 
@@ -219,6 +222,7 @@ class StackDiffer(Generic[DiffType]):
             self._remove_deployed_default_parameters_that_arent_passed(
                 deployed_template_summary, generated_config, deployed_config
             )
+
         if not self.show_no_echo:
             # We don't actually want to show parameters Sceptre is passing that the local template
             # marks as NoEcho parameters (unless show_no_echo is set to true). Therefore those
--- a/sceptre/diffing/stack_differ.py
+++ b/sceptre/diffing/stack_differ.py
@@ -172,7 +172,7 @@ class StackDiffer(Generic[DiffType]):
             if isinstance(value, str):
                 formatted_parameters[key] = value.rstrip("\n")
             elif value is None:
-                formatted_parameters[key] = value
+                formatted_parameters[key] = ""
 
         return formatted_parameters
 
--- a/tests/test_diffing/test_stack_differ.py
+++ b/tests/test_diffing/test_stack_differ.py
@@ -332,6 +332,17 @@ class TestStackDiffer:
             self.expected_deployed_config, self.expected_generated_config
         )
 
+    def test_diff__deployed_stack_has_default_values_one_of_them_none__compares_different_configs(
+        self,
+    ):
+        self.deployed_parameters["new"] = "default value"
+        self.deployed_parameter_defaults["new"] = None
+        self.parameters_on_stack_config["new"] = ""
+        self.differ.diff(self.actions)
+        self.command_capturer.compare_stack_configurations.assert_called_with(
+            self.expected_deployed_config, self.expected_generated_config
+        )
+
     def test_diff__stack_exists_with_same_config_but_template_does_not__compares_identical_configs(
         self,
     ):


ericabrauer (approved):
LGTM

sceptre/diffing/stack_differ.py
```diff
@@ -169,7 +169,10 @@ def _extract_parameters_from_generated_stack(self, stack: Stack) -> dict:
         for key, value in stack.parameters.items():
             if isinstance(value, list):
                 value = ",".join(item.rstrip("\n") for item in value)
-            formatted_parameters[key] = value.rstrip("\n")
+            if isinstance(value, str):
+                formatted_parameters[key] = value.rstrip("\n")
+            elif value is None:
+                formatted_parameters[key] = value
```
jfalkenstein, line 175: So I know we talked about this, but I just want to triple-check here: If you have created a stack by passing `None` as a parameter, when we describe the stack later, those same parameter values come back as None? Because if they come back as empty strings, this will result in a diff when there is none.

tests/test_diffing/test_stack_differ.py
```diff
@@ -332,6 +332,17 @@ def test_diff__deployed_stack_has_default_values__passes_different_value__compar
...
+        self,
+    ):
+        self.deployed_parameters["new"] = "default value"
+        self.deployed_parameter_defaults["new"] = None
+        self.parameters_on_stack_config["new"] = ""
+        self.differ.diff(self.actions)
+        self.command_capturer.compare_stack_configurations.assert_called_with(
+            self.expected_deployed_config, self.expected_generated_config
+        )
+
```
alexharv074, line 345: @jfalkenstein this test passes with or without my patch so it's evidently not a valid test. If you can see what's wrong that will be most appreciated!

jfalkenstein: Ok, so the lines of code you need to look at are here: https://github.com/Sceptre/sceptre/blob/master/sceptre/diffing/stack_differ.py#L257

Basically, the way the diffing works _right now_ is:
* If a given parameter is defined on the stack config, diff it's value with the deployed config
* If a given parameter is NOT defined on the stack config and the deployed parameter value is identical to its default value, delete the parameter from the deployed config, since we're relying on the default value so there should be no diff detected.

However, the change I think you'll need to make is to add a new condition to that:
* If a parameter is defined on the stack config as `None` and the parameter has a default value, delete that parameter off BOTH the generated config and the deployed config.

Regarding this test, then, I think part of your problem is that you're saying that the default value on the CloudFormation stack for the "new" parameter is `None`, but that's not a valid default value. What you need to test is:

First, you should update the expected_generated_config and expected_deployed_config properties so that the parameters are deepcopied as well. That is important. 

```python
        # The default value for the "new" parameter is "default_value"
        self.deployed_parameter_defaults["new"] = "default_value"
        # It just so happens that the deployed parameter value IS the default value, 
        # meaning we didn't override it
        self.deployed_parameters["new"] = "default value"
        # On the Stack Config, we're setting the value to None, which means we're effectively 
        # NOT setting any value at all.
        self.parameters_on_stack_config["new"] = None
        # We need to modify the expected generated config to be diffed. Look at the property 
        # for more details on how it's created to start with
        expected_generated_config = self.expected_generated_config
        expected_deployed_config = self.expected_deployed_config
        # We should expect that there is no "new" parameter specified at all in the generated_config or the 
        # deployed_config        
        del expected_generated_config.parameters['new']
        del expected_deployed_config.parameters['new']

        self.differ.diff(self.actions)
        self.command_capturer.compare_stack_configurations.assert_called_with(
            self.expected_deployed_config, self.expected_generated_config
        )
```

You'll also need test cases for when:
* There is no default value on the parameter
* There is no deployed config
* When the parameter value on the deployed config is NOT the default value but the parameter on the generated config is None (in which case it defaults to the deployed config).

sceptre/diffing/stack_differ.py
```diff
@@ -169,7 +169,10 @@ def _extract_parameters_from_generated_stack(self, stack: Stack) -> dict:
         for key, value in stack.parameters.items():
             if isinstance(value, list):
                 value = ",".join(item.rstrip("\n") for item in value)
-            formatted_parameters[key] = value.rstrip("\n")
+            if isinstance(value, str):
+                formatted_parameters[key] = value.rstrip("\n")
+            elif value is None:
+                formatted_parameters[key] = ""
```
jfalkenstein, line 175: I actually believe what needs to happen here is that we treat `None` as if there is no value at all and don't add it to the `formatted_parameters` dict at all.

sceptre/diffing/stack_differ.py
```diff
@@ -167,6 +167,12 @@ def _extract_parameters_from_generated_stack(self, stack: Stack) -> dict:
...
         for key, value in stack.parameters.items():
+            # When boto3 receives "None" for a cloudformation parameter, it treats it as if the
+            # value is not passed at all. To be consistent in our diffing, we need to skip Nones
+            # altogether.
+            if value is None:
+                continue
+
             if isinstance(value, list):
                 value = ",".join(item.rstrip("\n") for item in value)
             formatted_parameters[key] = value.rstrip("\n")
```
jfalkenstein, line 178: Do you think we need some sort of proper error message if the value isn't a list of strings, a string, or `None`? I mean, isn't that what you ran into in the first place? That the messaging wasn't clear about what the problem was?

alexharv074: @jfalkenstein didn't we already discuss this idea and conclude that this code path would not be reachable in any other scenario since Sceptre would not have allowed us to create the stack in the first place, e.g. if the value is a dict

```
"Parameter validation failed:\nInvalid type for parameter Parameters[0].ParameterValue, value: {'bar': 'baz'}, type: <class 'dict'>, valid types: <class 'str'>"
```

jfalkenstein: Yeah, that's probably fine then.

jfalkenstein (approved):

PR #1346, opened 2023-06-10
[Resolve #1344] Addressing issue with Generate hook and its deprecation

### LINKED ISSUE #1344: before_generate hook broken by recent deprecation
### Subject of the issue
The `before_generate` hook has been broken by a recent deprecation. 

### Your environment
* which OS/distro

### Steps to reproduce

### Expected behaviour
Tell us what should happen

### Actual behaviour
Tell us what happens instead

jfalkenstein:
This addresses the current issue raised in #1344. 

Here is what this includes:
* The `dump_template` action ALSO triggers `before_generate` and `after_generate`
* The `generate` action ALSO triggers `before_dump_template` and `after_dump_template`
* The StackDiffer invokes `dump_template` rather than `generate`, preventing our own code from triggering deprecation warnings.
* The documentation for hook points is updated to include all current hooks, including notes about the aliasing.
* Deprecation of `generate` is removed after discussion in Slack. This deprecation would be too disruptive to many users, and we agreed instead to support a permanent alias of `generate` => `dump_template`.

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/cli/template.py
+++ b/sceptre/cli/template.py
@@ -65,17 +65,17 @@ def validate_command(ctx, no_placeholders, path):
 @click.argument("path")
 @click.pass_context
 @catch_exceptions
-def generate_command(ctx: click.Context, no_placeholders, path):
+def generate_command(ctx: click.Context, no_placeholders: bool, path: str):
     """
     Prints the template used for stack in PATH.
 
-    This command is aliased to the dump template command; It's the same as running
-    `sceptre dump template`, for legacy support reasons.
+    This command is aliased to the dump template command for legacy support reasons. It's the same
+    as running `sceptre dump template`.
+
     \f
     :param no_placeholders: If True, will disable placeholders for unresolvable resolvers. By
         default, placeholders will be active.
     :param path: Path to execute the command on.
-    :type path: str
     """
     ctx.forward(dump_template)
 
--- a/sceptre/cli/template.py
+++ b/sceptre/cli/template.py
@@ -69,7 +69,8 @@ def generate_command(ctx: click.Context, no_placeholders, path):
     """
     Prints the template used for stack in PATH.
 
-    This command is aliased to running `sceptre dump template` for historical support reasons.
+    This command is aliased to the dump template command; It's the same as running
+    `sceptre dump template`, for legacy support reasons.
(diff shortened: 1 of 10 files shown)


sceptre/plan/actions.py
```diff
@@ -1155,7 +1155,7 @@ def dump_config(self):
         """
         return self.stack.config
 
-    @add_stack_hooks
+    @add_stack_hooks_with_aliases([generate.__name__])
```
jfalkenstein, line 1158: * Running `generate` will invoke the hooks for both `generate` and `dump_template` because the `generate` method invokes `dump_template` and...
* Running `dump_template` will invoke the hooks for both `generate` and `dump_template`

sceptre/hooks/__init__.py
```diff
@@ -98,3 +98,33 @@ def decorated(self, *args, **kwargs):
         return response
 
     return decorated
+
+
+def add_stack_hooks_with_aliases(function_aliases: List[str]):
```
alexharv074, line 103: Isn't it possible to make the original `add_stack_hooks` optionally take a list of aliases in a backwards-compatible way without introducing a new function?

jfalkenstein: Is it possible? Yes. But it would be a bit weird and ugly. A decorator that takes an argument needs to go one level deeper in terms of function scope than a decorator that doesn't. So I'd need to evaluate if the argument is a function or a list and then do different things in each situation. It's just cleaner to define two different functions with explicit names. I don't like fusing together functionality that is better explained if it's separated.

sceptre/cli/template.py
```diff
@@ -67,35 +65,19 @@ def validate_command(ctx, no_placeholders, path):
...
-
-    execution_context = (
-        null_context() if no_placeholders else use_resolver_placeholders_on_error()
-    )
-    with execution_context:
-        responses = plan.generate()
-
-    output = [template for template in responses.values()]
-    write(output, context.output_format)
+    ctx.forward(dump_template)
```
jfalkenstein, line 80: Rather than maintaining duplicate code, we're truly aliasing this to there

alexharv074 (approved):

PR #1352, opened 2023-06-23
[Resolve #1351] Add debugging to Jinja rendering

### LINKED ISSUE #1351: Add more debugging info around Jinja template rendering
### Subject of the issue
In #1321 the case of Jinja template rendering raising hard-to-debug exceptions was made easier to debug by writing debug info to a temp file and providing feedback in debug mode.

In that change, it was missed that Jinja rendering logic is duplicated in the file template handler.

### Your environment
* version of sceptre (sceptre --version) Latest
* version of python (python --version) All
* which OS/distro All

### Steps to reproduce

### Expected behaviour
Tell us what should happen

### Actual behaviour
Tell us what happens instead

alexharv074:
In #1321 the case of Jinja template rendering raising hard-to-debug exceptions was made easier to debug by writing debug info to a temp file and providing feedback in debug mode.

This extends to another Jinja rendering helper function that was missed earlier.

## PR Checklist

## Other Information


sceptre/config/reader.py
```diff
@@ -475,7 +459,7 @@ def _render(self, directory_path, basename, stack_group_config):
             message = f"{Path(directory_path, basename).as_posix()} - {err}"
 
             if logging_level() == logging.DEBUG:
-                debug_file_path = self._write_debug_file(
+                debug_file_path = write_debug_file(
                     json.dumps(self.templating_vars), prefix="vars_"
```
jfalkenstein, line 463: Hey, in all these places where we're dumping templating vars, we should dump with `indent=4`. That is going to make the debug files far more human-readable than a dense blob of unpretty json.

jfalkenstein (commented):
I'm cool with this, but I have one request, that we make the json a bit more readable.

sceptre/config/reader.py
alexharv074: Ok, done.

integration-tests/steps/helpers.py
```diff
@@ -47,6 +48,8 @@ def step_impl(context, exception_type):
         assert isinstance(context.error, AttributeError)
     elif exception_type == "UndefinedError":
         assert isinstance(context.error, jinja2.exceptions.UndefinedError)
+    elif exception_type == "SceptreException":
+        assert isinstance(context.error, SceptreException)
```
alexharv074, line 52: This whole test is nuts and I'd be more in favour of just removing the test so no one else has to waste time on this.

CONTRIBUTING.md
```diff
@@ -186,10 +186,6 @@ during development.
   to work with an AWS account that you have access to.  You can use the same user
   that you use for CircleCi.
 
-_Note_: All integration tests are set up to run in `eu-west-*` region.  If you prefer
-to run in a different region you must update the region in each test before running it
-
-
```
alexharv074, line 192: Considering no other reference in the code to `eu-west` I can only assume that this text above is no longer true.

zaro0508: We are amendable to removing this however we cannot remove it based on your assumption.   Would you be able to prove your assumption?

integration-tests/steps/helpers.py
zaro0508: we are open to the idea of removing the test but only if you can provide a good reason.

pyproject.toml
```diff
@@ -90,6 +90,7 @@ readme-renderer = "^24.0"
 requests-mock = "^1.9.3"
 tox = "^3.23.0"
 troposphere = "^4"
+ipdb = "^0.13.13"
```
zaro0508, line 93: what is this for?

sceptre/template_handlers/helper.py
```diff
@@ -142,5 +149,17 @@ def render_jinja_template(path, jinja_vars, j2_environment):
...
+    try:
+        body = template.render(**jinja_vars)
+    except Exception as err:
+        message = f"{path} - {err}"
+
+        if logging_level() == logging.DEBUG:
+            debug_file_path = write_debug_file(
+                json.dumps(jinja_vars, indent=4), prefix="vars_"
+            )
+            message += f"\nTemplating vars saved to: {debug_file_path}"
```
zaro0508, line 161: errors from this method are usually handled by `print_template_traceback`, would it make sense to put this stuff in `print_template_traceback` instead?

zaro0508 (changes requested):

pyproject.toml
alexharv074: It's the debugger and I've added it to the dev dependies section. It was quite a bit of work to install a debugger so seems like it probably is of benefit to stay here? I can't think of any reason not to?

CONTRIBUTING.md
alexharv074: Ah I know what happened, the code overwrites some of this stuff when you run it, and that's why grep didn't find it? Anyway, the `eu-west` bits really are still there so I'll revert this.

integration-tests/steps/helpers.py
alexharv074: It's just a badly designed test that isn't testing anything useful as far as I can tell. However, not worth discussing too much!

pyproject.toml
zaro0508: My view on dependencies are that they should only be added if used otherwise it should not be there.  Your thoughts @jfalkenstein ?

jfalkenstein: Yeah, I agree. There isn't a good reason to put this into our dependencies. Btw, you can always still `pip install` using poetry in your own environment if you need to throw in an additional thing into your virtual environment. You can either run `poetry run pip install ipdb` or run `poetry shell` then within the shell run `pip install ipdb`. It doesn't need to be in our pyproject.toml (and I don't think it should be).

alexharv074: Ok removed.

sceptre/template_handlers/helper.py
alexharv074: Condensed version of discussion Slack:
1. The `print_template_traceback` function prints a formatted stack trace; this code is writing out the templating vars (and in other contexts the rendered template) to a file for debugging purposes.
2. There is duplication of Jinja rendering in this code base, with separate Jinja rendering implemented in the config reader and templating classes. Meanwhile this debug file is used in both but primarily in the reader class, where the `print_template_traceback` is not available. It could of course be moved, but point (1) would need to be addressed first.

zaro0508: ok, I have taken a closer look and you have convinced me.  thanks.

PR #1357, opened 2023-07-09
[Resolves #1356] Add export=stackoutput and stackoutputexternal features to list outputs

### LINKED ISSUE #1356: Stack outputs in stack groups hide the stack names
### An export format for stack outputs that shows stack name is needed
The main purpose of stack outputs is to consume them in other stacks. Example:

```yaml
VpcId: !stack_output network.yaml::VpcId
```

or in a var file:

```yaml
VpcId: '!stack_output_external my-network-stack::VpcId'
```

In the case of stack groups however, there is no easy way to know which stack exports which outputs without either reading the source, going to the AWS Console, or using AWS CLI commands. Furthermore, even knowing what the stack name of any stack is requires use of Sceptre's `list stacks` command as Sceptre's logic of setting the stack name has never been obvious.

The options we have are:

1. list outputs 
```
% sceptre --output=yaml list outputs
---                       
- direct-connect:                     
  - OutputKey: VPGWId     
    OutputValue: vgw-1fb1e1e1af11da11b
- vpc:                             
  - ExportName: nonprod-DataSubnetCIDRa
    OutputKey: DataSubnetCIDRa     
    OutputValue: 11.11.111.11/11      
  - ExportName: nonprod-PublicSubnetCIDRa
    OutputKey: PublicSubnetCIDRa      
    OutputValue: 11.11.111.111/11
...
```

This command is flawed, in that it does not report true stack names, but names like `direct-connect` and `vpc` that only have meaning inside Sceptre itself.

2. list outputs --export=envvar
```
% sceptre --output=yaml list outputs --export=envvar
export SCEPTRE_VPGWId='vgw-1fb1e1e1af11da11b'
export SCEPTRE_DataSubnetCIDRa='11.11.111.11/11'
export SCEPTRE_PublicSubnetCIDRa='11.11.111.111/11'
export SCEPTRE_DataSubnetCIDRb='11.11.111.11/11'
export SCEPTRE_VPCId='vpc-111111ab111a11e11'
...
```

This command is also flawed, as it does not report the stack name at all.

Note that this same CLI behaviour of listing outputs originates in the initial commit of this project back in 2017. In that original version of the code we had:

```python
@cli.command(name="describe-stack-outputs")
@stack_options
@click.option("--export", type=click.Choice(["envvar"]))
@click.pass_context
@catch_exceptions
def describe_stack_outputs(ctx, environment, stack, export):
    """
    Describes stack outputs.

    Describes ENVIRONMENT/STACK's stack outputs.
    """
    env = get_env(ctx.obj["sceptre_dir"], environment, ctx.obj["options"])
    response = env.stacks[stack].describe_outputs()

    if export == "envvar":
        write("\n".join(
            [
                "export SCEPTRE_{0}={1}".format(
                    output["OutputKey"], output["OutputValue"]
                )
                for output in response
            ]
        ))
    else:
        write(response, ctx.obj["output_format"])
```

### Proposal export=stackoutput and export=stackoutputexternal

In addition to `--export=envvar` options like `--export=stackoutput` and `--export=stackoutputexternal` are needed. They might look like this:

```
% sceptre list outputs --export=stackoutput
!stack_output direct-connect.yaml::VPGWId --- vgw-1fb1e1e1af11da111
!stack_output vpc.yaml::DataSubnetCIDRa --- 11.11.111.11/11
!stack_output vpc.yaml::PublicSubnetCIDRa --- 11.11.111.111/11
!stack_output vpc.yaml::DataSubnetCIDRb --- 11.11.111.11/11
!stack_output vpc.yaml::VPCId --- vpc-111111ab111a11e11
...
```
and

```
% sceptre list outputs --export=stackoutputexternal
!stack_output_external nonprod-direct-connect::VPGWId --- vgw-1fb1e1e1af11da111
!stack_output_external nonprod-vpc::DataSubnetCIDRa --- 11.11.111.11/11
!stack_output_external nonprod-vpc::PublicSubnetCIDRa --- 11.11.111.111/11
!stack_output_external nonprod-vpc::DataSubnetCIDRb --- 11.11.111.11/11
!stack_output_external nonprod-vpc::VPCId --- vpc-111111ab111a11e11
...
```

alexharv074:
This adds features `list outputs --export=stackoutput` and `list outputs --export=stackoutputexternal` that are designed to reveal export outputs in a useful, formatted manner (like `--export=envvar`) as well as the crucial external stack name in the case of `stackoutputexternal` that is needed to consume these outputs.

## PR Checklist

## Other Information


sceptre/cli/list.py
```diff
@@ -89,6 +89,21 @@ def list_outputs(ctx, path, export):
...
                         "text",
                     )
+
+    elif export == "stackoutput":
+        stack_names = {stack.name: stack.external_name for stack in plan.graph}
+        for response in responses:
+            for stack_name, stack in response.items():
+                for output in stack:
+                    write(
+                        "!stack_output_external {0}::{1} [{2}]".format(
```
alexharv074, line 99: Not clear from this diff, but use of format rather than f-string is consistent with the implementation of the case `export == "envvar"`.

sceptre/cli/list.py
```diff
@@ -89,6 +89,21 @@ def list_outputs(ctx, path, export):
...
+                            stack_names[stack_name],
+                            output.get("OutputKey"),
+                            output.get("OutputValue"),
+                        ),
+                        "text",
+                    )
+
```
alexharv074, line 106: Not clear from this diff, but the use of format rather than an f-string is consistent with the implementation of the case `export == "envvar"`.

sceptre/cli/list.py
jfalkenstein, line 99: These are not stack-outputs. They're external stack-outputs. I think this particular functionality seems pretty calibrated to your own particular use-case. But I would argue stack_output_external is less commonly used than `stack_output`. If you're outputting "stack outputs", I wouldn't expect them to be external ones.

alexharv074: agree, have a look now.

sceptre/cli/list.py
```diff
@@ -89,6 +93,39 @@ def list_outputs(ctx, path, export):
...
                         "text",
                     )
+
+    # Format outputs as !stack_output references.
+    elif export == "stackoutput":
+        for response in responses:
+            for stack_name, stack in response.items():
+                for output in stack:
+                    write(
+                        "!stack_output {0}.yaml::{1} [{2}]".format(
```
zaro0508, line 103: what are your thoughts about not including `!stack_output` or `!stack_ouput_external` in the output?

sceptre/cli/list.py
```diff
@@ -89,6 +93,39 @@ def list_outputs(ctx, path, export):
                         ),
                         "text",
                     )
+
+    # Format outputs as !stack_output references.
+    elif export == "stackoutput":
```
zaro0508, line 98: what are you thoughts about changing this to `stack_output` to match `!stack_output`?

sceptre/cli/list.py
```diff
@@ -89,6 +93,39 @@ def list_outputs(ctx, path, export):
...
+                        "!stack_output {0}.yaml::{1} [{2}]".format(
+                            stack_name,
+                            output.get("OutputKey"),
+                            output.get("OutputValue"),
+                        ),
+                        "text",
+                    )
+
+    # Format outputs as !stack_output_external references.
+    elif export == "stackoutputexternal":
```
zaro0508, line 112: what are you thoughts about changing this to `stack_output_external`?

sceptre/cli/list.py
zaro0508: We use `!stack_output_external` more than `!stack_outputs`.  I actually didn't even know about the export flag until now, but this seems pretty useful for my use case.

sceptre/cli/list.py
```diff
@@ -79,6 +80,9 @@ def list_outputs(ctx, path, export):
     plan = SceptrePlan(context)
     responses = [response for response in plan.describe_outputs().values() if response]
 
+    # Legacy. This option was added in the initial commit of the project,
+    # although its intended use case is unclear. It may relate to a feature
+    # that had been removed prior to the initial commit.
```
zaro0508, line 85: i believe this is intended to be used with the [sceptre environment variable resolver](https://docs.sceptre-project.org/4.2.2/docs/resolvers.html#environment-variable).

zaro0508 (changes requested):
there is a test for ouput export with envvars in `tests/test_cli/test_cli_commands.py`.  I would like to see tests for these new export options as well.

sceptre/cli/list.py
alexharv074: Ok, noted. Will amend the comment.

sceptre/cli/list.py
alexharv074: I have tried to be consistent with the original design as much as possible, so since `--export=envvar` prints out Bash code that can be copied/pasted (or eval'ed) this also prints out formatted output that can be copy/pasted into your code. So that's consistent and also convenient in that it saves some keystrokes of typing.

Similarly, I felt that following the case convention (`stackoutput` and `envvar` instead of `stack_output` and `env_var`) will look cleaner and more consistent in the docs and help message. 

But I don't really mind.

sceptre/cli/list.py
alexharv074: See comments above.

sceptre/cli/list.py
alexharv074: Quoting @jfalkenstein offline:

> I think your argument is consistent. On a scale of caring 1-10... I'm at like a 2, though.

Shall we leave it all as-is then @zaro0508 ?

sceptre/cli/list.py
```diff
@@ -46,12 +46,13 @@ def list_resources(ctx, path):
     write(responses, context.output_format)
 
 
+# flake8: noqa: C901
```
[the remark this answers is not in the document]
alexharv074: @zaro0508  Assuming we agree to leave the rest of the implementation as-is, the question of what to do about the linter remains. The linter is complaining that this function (with my additional branches) is now too complex:

Screenshot 2023-08-31 at 3 29 40 pm

I would tend to argue that the linter is wrong! I think it's quite readable and easy to understand and if I am to refactor, we are going to end up with helper functions somewhere that will be dedicated to this particular use-case and not make much sense.

I could explore the possibility of a nested function perhaps? Or just leave it as-is.

sceptre/cli/list.py
jfalkenstein: My name's Jon Falkenstein and I approve this message.

zaro0508: ok, leave this one as as-is is fine.

sceptre/cli/list.py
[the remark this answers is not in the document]
zaro0508: i think flake8 is really complaining about the multiple nested for loops in this function.  hmm, i'm a bit torn because i do agree that it's pretty readable but I don't like the idea of disabling the flake8 check for the entire list_resources function.  How about creating a helper function in this file then disabling the check only on the helper function? 

another option might be to use the [python itertools.product](https://docs.python.org/3.8/library/itertools.html#itertools.product) function to reduce the number of loops?

zaro0508 (changes requested):

sceptre/cli/list.py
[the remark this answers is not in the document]
alexharv074: Concerning your comment here https://github.com/Sceptre/sceptre/pull/1357#discussion_r1311894136 let me have a look at what the options are. Note that it only seems possible to disable this check for the whole file.

sceptre/cli/list.py
alexharv074: Because I haven't done anything yet!

alexharv074: @zaro0508 I am actually looking at this now. I still can't quite see the connection. How does code that exports env vars as `SCEPTRE_some_var` connect to the `!environment_variable` resolver? I would think the resolver simply allows us to use environment variables as inputs, but nothing specifically to do with environment variables whose names are prefixed with `SCEPTRE_`?

The feature reminds me of the `TF_VAR_some_var` feature that Terraform provides. But as far I can tell, Sceptre hasn't implemented anything similar.

sceptre/cli/list.py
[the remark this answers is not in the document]
alexharv074: @zaro0508 Have a look now. I think it looked better before, but I did find a way to satisfy the linter. (I also added tests, but have not amended the comment, see my other comment.)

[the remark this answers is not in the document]
alexharv074: (Also, I don't think the C901 issue is with nested for loops specifically (and my additional has not actually changed the nesting of for loops), but cyclomatic complexity which I understand to be number of logical pathways through a function. I investigated the itertools function but I don't think this is a use-case for that?)

sceptre/cli/list.py
```diff
@@ -80,52 +134,12 @@ def list_outputs(ctx, path, export):
...
-                for output in stack:
-                    write(
-                        "export SCEPTRE_{0}='{1}'".format(
-                            output.get("OutputKey"), output.get("OutputValue")
-                        ),
-                        "text",
-                    )
-
-    # Format outputs as !stack_output references.
+        process_export_envvar(responses)
```
zaro0508, line 138: instead of 3 helper functions why not include the `if` conditions and create just 1 helper function?  i think the helper function provides more context with the `if` conditions.

zaro0508 (changes requested):

sceptre/cli/list.py
alexharv074: Yes, but that just moves the complexity problem somewhere else:

Screenshot 2023-09-03 at 11 22 02 pm

Any other ideas?

zaro0508: i think it's fine as is because i have no other ideas.  I'm wondering if @jfalkenstein might have something up his sleeve?

alexharv074: I still prefer the first implementation, because I think it is easier to understand and more consistent with the rest of the functions in this CLI. I would have no problem with a `noqa` line. But I really don't mind.

@jfalkenstein, your thoughts?

sceptre/cli/list.py
alexharv074: Is this point resolved @zaro0508 i.e. we really don't know what the `SCEPTRE_env_vars` are for and no reason for a comment in the code not to say so? Otherwise, happy to change or remove the comment as you see fit.

zaro0508: yes, i'm fine to keep as is

sceptre/cli/list.py
zaro0508: how about the idea of single helper function with `if`+`for`s+`noqa`?

jfalkenstein: So here's my two-cents on what _I'd_ do, based mostly on the Single Responsibility Principle:
1. I'd make a generator called `iterate_stack_outputs` that iterates tuples of `(stack_name, output_dict)`. This basically flattens out the loops within loops within loops you keep repeating.
2. I'd make three functions: `write_envvar(output_dict)`, `write_stack_output_resolver(stack_name, output_dict)`, and `write_external_stack_output_resolver(stack_name, output_dict)`

Thus, you're separating the concerns of (1) looping and (2) writing, one for each output. 

The generator in point 1 above would look something like this:

```python
def iterate_stack_outputs(responses: list[dict[str, list[dict]]]):
    for response in responses:
        for stack_name, values in response.items()
            for value in values:
                yield stack_name, value
```

This way you have a single iterator that flattens it all out.

alexharv074: @jfalkenstein , @zaro0508 Updated as per Jon's suggestions. Thoughts? I can live with any of these solutions, although I think I am still in favour of the original for the same reasons. I think we've made the code a bit harder to understand in this version despite that it is true that we have refactored out some duplication using the iterator.

alexharv074: I wouldn't mind this one so much if it would let me use nested functions but sadly it still fails the complexity check if I move these to be nested functions.

zaro0508: I'm not a fan of @jfalkenstein suggestion.  I would definitely have a harder time understanding it vs the original simple implementation.  I would prefer to go with the initial simpler implementation.  With that being said i'm also ok with  @jfalkenstein's suggestion as long as we add some comments explaining that we implemented it that way to    avoiding `noqa` failures.

zaro0508 (approved):
LGTM.  Thank you for being patient on this one @alexharv074

PR #1372, opened 2023-08-27
[Resolves #1213] Add shell option to cmd hook

### LINKED ISSUE #1213: !cmd hook runs in `sh`, which is not my default shell
### Subject of the issue
When I use a `!cmd` hook, it appears to be run by `sh` because I receive this error: 

```shell
/bin/sh: 1: [[: not found
```

Is there a way to force the underlying `subprocess.run()` call to use Python? I know I can do this in Python by using arguments that look like `["bash", "-c", "your command goes here"]`, and, in fact, if I simply pass in `["your command goes here"]` in Python, it will run in bash because it is my default shell. Is `sceptre` really running my hook using `sh`? If so, how can I make it use `bash` instead?

### Your environment
**Sceptre, version 2.4.0**
**Python 3.8.10**
* which OS/distro
**Ubuntu 20.04.3 LTS on 4.4.0-19041-Microsoft #1237-Microsoft x86_64** 

### Steps to reproduce
Tell us how to reproduce this issue. Please provide sceptre project files if possible,

Here is a hook that causes the behavior:
```yaml
hooks:
    before_update:
        - !cmd 'if [[ "{{ var.update_lambdas | default("false") }}" == "true" ]]; then lambdas/deploy-lambdas.sh -u "$(./get-stack-output.sh "{{ environment }}/{{ region }}/LambdaS3Bucket.yaml" LambdaBucketName)" HandleServiceAccountAppGitHubWebhook; fi;'
```

### Expected behaviour
The hook should run successfully using my default shell, which is `bash`.

### Actual behaviour
I see the error `/bin/sh: 1: [[: not found` instead.

iainelder:
The cmd hook now accepts named arguments `run` and `shell` to run a command in a
given shell. For backwards compatibility it still supports the unnamed
argument to run a command in the default shell.

The documentation shows the expected output for examples of each argument style.

The unit tests no longer depend on the implementation of Cmd and test only the
behavior. This makes it easier to change how the hook controls the subprocess.

## PR Checklist

- [x] ~Commit message~ PR title starts with `[Resolve #issue-number]`.

## Other Information

### DIFF
--- a/docs/_source/docs/hooks.rst
+++ b/docs/_source/docs/hooks.rst
@@ -57,7 +57,7 @@ Syntax:
    # Another shell.
    <hook_point>:
      - !cmd
-         command: <shell_command>
+         run: <shell_command>
          shell: <shell_name_or_path>
 
 Pass the command string as the only argument to use the default shell.
@@ -85,7 +85,7 @@ A successful command prints its standard output messages between status messages
    [2023-09-03 01:06:31] - test - Updating Stack
    [2023-09-03 01:06:31] - test - No updates to perform.
 
-Pass named arguments to use a different shell. Here the command string is called ``command`` and the
+Pass named arguments to use a different shell. Here the command string is called ``run`` and the
 shell executable is called ``shell``.
 
 Write the executable name as you would type it at an interactive prompt to start the shell. For
@@ -97,7 +97,7 @@ absolute path such as ``/bin/bash``.
    hooks:
      before_update:
        - !cmd
-           command: echo "Hello, $0!"
+           run: echo "Hello, $0!"
            shell: bash
 
 .. code-block:: text
@@ -115,7 +115,7 @@ You can use PowerShell in the same way.
    hooks:
      before_update:
        - !cmd
-           command: Write-Output "Hello, Posh!"
+           run: Write-Output "Hello, Posh!"
            shell: pwsh
 
 .. code-block:: text
@@ -126,7 +126,7 @@ You can use PowerShell in the same way.
    [2023-09-04 00:44:34] - test - Updating Stack
    [2023-09-04 00:44:34] - test - No updates to perform.
 
-If the shell command fails, so does Sceptre. The command output sits between Sceptre's status
+If the shell command fails, so does Sceptre. Its output sits between Sceptre's status
 messages and a Python traceback.
 
 .. code-block:: yaml
--- a/docs/_source/docs/hooks.rst
+++ b/docs/_source/docs/hooks.rst
@@ -52,13 +52,13 @@ Syntax:
 
    # Default shell.
    <hook_point>:
-     - !cmd <command string>
+     - !cmd <shell_command>
 
    # Another shell.
    <hook_point>:
      - !cmd
-         args: <command string>
-         executable: <shell command>
+         args: <shell_command>
+         executable: <shell_name_or_path>
 
 Pass the command string as the only argument to use the default shell.
 
--- a/docs/_source/docs/hooks.rst
+++ b/docs/_source/docs/hooks.rst
@@ -57,8 +57,8 @@ Syntax:
    # Another shell.
    <hook_point>:
      - !cmd
-         args: <shell_command>
-         executable: <shell_name_or_path>
+         command: <shell_command>
+         shell: <shell_name_or_path>
 
 Pass the command string as the only argument to use the default shell.
 
@@ -95,8 +95,8 @@ in system path, you can write ``bash``; otherwise, you need to write the absolut
    hooks:
      before_update:
        - !cmd
-           args: "if foo; then baz; else qux; fi"
-           executable: "bash"
+           command: "if foo; then baz; else qux; fi"
+           shell: "bash"
 
 This example fails because none of the ``foo``, ``baz``, and ``quz`` commands exist. Its standard error messages sit between Sceptre's status messages and a Python traceback.
 
--- a/docs/_source/docs/hooks.rst
+++ b/docs/_source/docs/hooks.rst
@@ -50,15 +50,15 @@ Syntax:
 
 .. code-block:: yaml
 
-  # Default shell.
-  <hook_point>:
-    - !cmd <command string>
+   # Default shell.
+   <hook_point>:
+     - !cmd <command string>
 
-  # Another shell.
-  <hook_point>:
-    - !cmd
-        args: <command string>
-        executable: <shell command>
+   # Another shell.
+   <hook_point>:
+     - !cmd
+         args: <command string>
+         executable: <shell command>
 
 Pass the command string as the only argument to use the default shell.
 
@@ -70,19 +70,19 @@ quotes and backslashes to escape filenames with spaces in them.
 
 .. code-block:: yaml
 
-  hooks:
-    before_update:
-      - !cmd "echo 'Hello, world!'"
+   hooks:
+     before_update:
+       - !cmd "echo 'Hello, world!'"
 
 This example succeeds. Its standard output messages appear between status messages from Sceptre.
 
 .. code-block::
 
-  [2023-09-03 01:06:28] - test - Launching Stack
-  [2023-09-03 01:06:29] - test - Stack is in the CREATE_COMPLETE state
-  Hello, world!
-  [2023-09-03 01:06:31] - test - Updating Stack
-  [2023-09-03 01:06:31] - test - No updates to perform.
+   [2023-09-03 01:06:28] - test - Launching Stack
+   [2023-09-03 01:06:29] - test - Stack is in the CREATE_COMPLETE state
+   Hello, world!
+   [2023-09-03 01:06:31] - test - Updating Stack
+   [2023-09-03 01:06:31] - test - No updates to perform.
 
 Pass named arguments ``args`` and ``executable`` to use a different shell.
 
@@ -92,23 +92,23 @@ in system path, you can write ``bash``; otherwise, you need to write the absolut
 
 .. code-block:: yaml
 
-  hooks:
-    before_update:
-      - !cmd
-          args: "if foo; then baz; else qux; fi"
-          executable: "bash"
+   hooks:
+     before_update:
+       - !cmd
+           args: "if foo; then baz; else qux; fi"
+           executable: "bash"
 
 This example fails because none of the ``foo``, ``baz``, and ``quz`` commands exist. Its standard error messages sit between Sceptre's status messages and a Python traceback.
 
 .. code-block:: text
 
-  [2023-09-03 02:00:21] - test - Launching Stack
-  [2023-09-03 02:00:22] - test - Stack is in the CREATE_COMPLETE state
-  bash: foo: command not found
-  bash: qux: command not found
-  Traceback (most recent call last):
-  <snip>
-  subprocess.CalledProcessError: Command 'if foo; then baz; else qux; fi' returned non-zero exit status 127.
+   [2023-09-03 02:00:21] - test - Launching Stack
+   [2023-09-03 02:00:22] - test - Stack is in the CREATE_COMPLETE state
+   bash: foo: command not found
+   bash: qux: command not found
+   Traceback (most recent call last):
+   <snip>
+   subprocess.CalledProcessError: Command 'if foo; then baz; else qux; fi' returned non-zero exit status 127.
 
 
 asg_scaling_processes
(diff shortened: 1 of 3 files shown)


sceptre/hooks/cmd.py
```diff
@@ -19,8 +19,16 @@ def run(self):
...
+            executable = None
+        elif isinstance(self.argument, dict):
+            args = self.argument["args"]
+            executable = self.argument["executable"]
+
         try:
-            subprocess.check_call(self.argument, shell=True, env=envs)
+            subprocess.check_call(args, shell=True, env=envs, executable=executable)
         except TypeError:
             raise InvalidHookArgumentTypeError(
```
iainelder, line 33: The error message needs to change to reflect the new options. And what type of inputs will be invalid now? I hope after I learn about the unit tests it will be clear how to rewrite this.

iainelder:
See my [sceptre-dev-notes](https://github.com/iainelder/sceptre-dev-notes/blob/main/dev-notes/cmd-hook-options.md) repo for the notes on the devleopment and testing. If any of that is useful to put in the main repo I can add it here.

iainelder:
Well, tox gives me some failing tests, so I guess I should start here!

```text
Results (15.28s):
     635 passed
       2 failed
         - tests/test_hooks/test_cmd.py:18 TestCmd.test_run_with_non_str_argument
         - tests/test_hooks/test_cmd.py:23 TestCmd.test_run_with_str_argument
ERROR: InvocationError for command /home/isme/.local/bin/poetry run pytest --cov=sceptre --cov-append --cov-report=term-missing --junitxml=test-reports/junit-py310.xml (exited with code 1)
```

sceptre/hooks/cmd.py
```diff
@@ -19,10 +19,15 @@ def run(self):
...
-                "arguments of type string.".format(self.argument)
-            )
+
+        if self.argument is None:
+            raise InvalidHookArgumentTypeError()
+
+        if isinstance(self.argument, str) or self.argument is None:
+            args = self.argument
+            executable = None
+        elif isinstance(self.argument, dict):
```
zaro0508, line 29: passing in a dict is new so could you please add some docs for this?  an example would probably help as well.

also there should probably an else condition?  what if user passes in a list?  or you could probably keep the try/catch to handle all the other wacko things that users do.

tests/test_hooks/test_cmd.py
```diff
@@ -11,27 +10,34 @@
...
         self.cmd = Cmd(stack=self.stack)
 
     def test_run_with_non_str_argument(self):
         self.cmd.argument = None
         with pytest.raises(InvalidHookArgumentTypeError):
             self.cmd.run()
 
-    @patch("sceptre.hooks.cmd.subprocess.check_call")
-    def test_run_with_str_argument(self, mock_call):
+    def test_run_with_str_argument(self, capfd):
```
zaro0508, line 33: remember to test running with a dict arg as well.

sceptre/hooks/cmd.py
```diff
@@ -19,10 +19,15 @@ def run(self):
...
-        except TypeError:
-            raise InvalidHookArgumentTypeError(
-                'The argument "{0}" is the wrong type - cmd hooks require '
-                "arguments of type string.".format(self.argument)
-            )
+
+        if self.argument is None:
+            raise InvalidHookArgumentTypeError()
+
+        if isinstance(self.argument, str) or self.argument is None:
```
zaro0508, line 26: doesn't the previous condition already check `self.argument is None`?

zaro0508 (changes requested):

sceptre/hooks/cmd.py
iainelder: Good catch. I added the previous condition but didn't change it here.

sceptre/hooks/cmd.py
iainelder: Indeed, I'll add some tests for invalid inputs so that I can handle those properly.

tests/test_hooks/test_cmd.py
iainelder: Yes, I'm looking at how to do that. I've tried to make my tests depend only on the behavior and not the implementation. Just now the test doesn't know that subprocess is being used.

But I don't know how to to test that the `executable` option really causes a different shell to be launched.

I think I might have to use a mock after all and make the tests depend on the `subprocess` module. That module is the obvious way to implement this hook, so maybe it's not so bad.

iainelder:
I've addressed all the feedback about the code. Now I need to add documentation for the new features.

iainelder:
@zaro0508 , how does it look now?

Should I add some tests that run the sceptre CLI on some YAML files? I've just done that testing in my own terminal.

docs/_source/docs/hooks.rst
```diff
@@ -44,23 +44,72 @@ Available Hooks
...
 
-   before_create:
-     - !cmd "echo hello"
+   hooks:
+     before_update:
+       - !cmd
+           command: "if foo; then baz; else qux; fi"
+           shell: "bash"
+
+This example fails because none of the ``foo``, ``baz``, and ``quz`` commands exist. Its standard error messages sit between Sceptre's status messages and a Python traceback.
```
zaro0508, line 101: seems odd to me to use an example that fails.  i'm not sure I understand what this example is trying to demonstrate specifically?  why not just do  `command: "echo $0"` which should echo "bash"?

zaro0508 (approved):
LGTM, just had a question in docs.

docs/_source/docs/hooks.rst
iainelder: Yes, good point. I will split this into two examples to make it clearer.

* one that echoes the shell name and uses Bash syntax to prove we are using the Bash shell
* one that fails (could be any shell) to show what Sceptre does with the shell error

iainelder: @zaro0508 I used your simpler example and added a new one to show the error behavior. I added a PowerShell example too since it's a common choice on Windows. How does it look now?

iainelder:
To avoid confusing people who try this on an old version, I want the new syntax to cause an error in Sceptre v4.2.2.

When `command` is the first named argument, Sceptre v4.2.2 executes `sh -c command shell`. This has no output and returns 0, which makes it look like Sceptre just ignored the hook.

As a compromise I suggest we replace `command` with `run` because it breaks in a more obvious way. The goal here is to avoid confusing people who try the new syntax on an old version of Sceptre.

[GitHub Actions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun) uses `run` to mean the same thing, so we would be in good company.

```pycon
>>> subprocess.check_call({"run": "echo $0", "shell": "bash"}, shell=True)
shell: 1: run: not found
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/subprocess.py", line 364, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '{'run': 'echo $0', 'shell': 'bash'}' returned non-zero exit status 127.
```

sceptre/hooks/cmd.py
```diff
@@ -13,16 +13,33 @@ def __init__(self, *args, **kwargs):
...
+            shell = None
+
+        elif (
+            isinstance(self.argument, dict)
+            and set(self.argument) == {"run", "shell"}
+            and isinstance(self.argument["run"], str)
+            and isinstance(self.argument["shell"], str)
+        ):
+            command_to_run = self.argument["run"]
+            shell = self.argument["shell"]
```
zaro0508, line 36: i didn't even consider windows.  What happens if users attempt to use a windows shell on linux or the other way around?  should we bring back the try/catch or add some error handling around `shell` inputs as well?

docs/_source/docs/hooks.rst
```diff
@@ -44,23 +44,105 @@ Available Hooks
...
+Write the executable name as you would type it at an interactive prompt to start the shell. For
+example, if Bash is in the system path, you can write ``bash``; otherwise, you need to write the
+absolute path such as ``/bin/bash``.
+
+.. code-block:: yaml
+
+   hooks:
+     before_update:
+       - !cmd
+           run: echo "Hello, $0!"
```
zaro0508, line 100: this parameter change is not backwards compatible.  I think this change should be in a separate PR with a separate discussion thread

zaro0508 (changes requested):

sceptre/hooks/cmd.py
iainelder: I'm not against adding a try/catch around the shell invocation. I just don't know what would be useful to catch yet. It might be useful later to wrap the stock Python exception with a Sceptre exception for nicer presentation of the errors, but I had considered that out of scope for now.

I don't think it's our business to know which shells are available in the environment. If the user tells Sceptre to use a shell that isn't available, then Sceptre should fail. I'll add a test and documentation to show what that failure looks like. This is useful to know how it behaves when I do something silly like type `bsah` instead of `bash`.

(PowerShell ships with Windows, but it's no longer just a shell for Windows. I [installed it on Ubuntu](https://learn.microsoft.com/en-us/powershell/scripting/install/installing-powershell-on-linux?view=powershell-7.3) to test the `pwsh` part of the documentation! Similarly, I believe [Windows for Subsystem for Linux (WSL)](https://learn.microsoft.com/en-us/windows/wsl/install) users can use the Bash shell.)

docs/_source/docs/hooks.rst
iainelder: Can you tell me what "backwards compatible" means for this feature?

The latest released version of Sceptre is v4.2.2 and supports only the string argument. The keyword arguments, whatever we decide to call them, are a new feature. I thought that as long as we continue to support the string argument then we satisfy the backwards-compatible requirement.

Over the weekend I renamed `command` to `run`, but only in the branch.

zaro0508: oops, I misread.  i thought you wanted to replace `!cmd` with `!run`.  sorry about that.  replacing `command` is fine since it's never been used before.

sceptre/hooks/cmd.py
zaro0508: hmm, `FileNotFoundError` for invalid shell name is not very helpful.  would you be able to provide a more helpful message?

zaro0508 (changes requested):

docs/_source/docs/hooks.rst
iainelder: Ah, okay. Yes, there's no need to change the name of the hook. Apparently that's controlled in poetry.toml.

https://github.com/Sceptre/sceptre/blob/33404ee5486ff345ed738044aeb1cc3b71fdb129/pyproject.toml#L30

docs/_source/docs/hooks.rst
zaro0508: looks good

sceptre/hooks/cmd.py
iainelder: The full error message is `FileNotFoundError: [Errno 2]` No such file or directory: '/bin/bsah'`. It does print the bad shell name. But we could improve the message with a Sceptre error that tells the user to check the shell input.

Since we've started talking about custom error handling, it would be great to suppress the stack trace for these kinds of errors that we know how to handle. Currently Sceptre will print about 50 lines of unhelpful stack trace between the first useful error message and the last useful error message. I'll see what I can do about that.

zaro0508: ohh, the full message looks good to me.  It's up to you but I don't think you need to add supression of the stack trace on this PR.  i'm ok if you want to do it as a separate change.

zaro0508 (changes requested):
could you please update the subject of this PR to include the issue, something like `[Resolves #1213] Add options to cmd hook`?

This branch is out of date so you will need to update (merge) this PR with master.

sceptre/hooks/cmd.py
iainelder: Thanks. I agree about using another PR. The stack trace suppression happens elsewhere in the code. I don't want to make this PR any wider.

iainelder:
I renamed the issue and merged the main branch.

I made the error message checking of the tests more robust.

How does it look now?

iainelder:
I understand that I need to squash all the commits before you merge it. I'll have a look at the other merge commit messages.

iainelder:
I updated the title and description with my proposal for the final commit message.

zaro0508 (approved):
nice work @iainelder

PR #1374, opened 2023-09-23
[Resolves #841] Add documented name as entry point for ASG hook

### LINKED ISSUE #841: ASGScalingProcesses registered as 'asg_scheduled_actions', expected to be 'asg_scaling_processes'
I noticed that the default hook for ASG processes is registered as `            "asg_scheduled_actions ="` and not as `asg_scaling_processes`, see https://github.com/Sceptre/sceptre/blob/b7585fd7bf9607230f9d4324166b49cd59cab3ee/setup.py#L59

In the docs (2.2.1) and also the code (https://github.com/Sceptre/sceptre/blob/9a36309ed1daa15ef65504fa550adf161e6d7c6f/sceptre/hooks/asg_scaling_processes.py#L39-L40) it's referenced as `asg_scaling_processes`, but using `!asg_scaling_processes` in `hooks` results in:
```
yaml.constructor.ConstructorError: could not determine a constructor for the tag '!asg_scaling_processes'
  in "<unicode string>", line 40, column 7:
        - !asg_scaling_processes 'suspend: ...
          ^
```

iainelder:
[Your PR description here]

## PR Checklist

## Other Information

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,6 +22,7 @@ classifiers = [
 
 [tool.poetry.plugins."sceptre.hooks"]
 "asg_scheduled_actions" = "sceptre.hooks.asg_scaling_processes:ASGScalingProcesses"
+"asg_scaling_processes" = "sceptre.hooks.asg_scaling_processes:ASGScalingProcesses"
 "cmd" = "sceptre.hooks.cmd:Cmd"
 
 [tool.poetry.plugins."sceptre.resolvers"]


iainelder:
[My comment on #841](https://github.com/Sceptre/sceptre/issues/841#issuecomment-1732367411) explains the change.

Can you think of a way to test this in the current framework? Do we need to?

I don't think a unit test isn't appropriate for this change. A test would need a real template as input and it would need to check the behavior of the CLI.

If we are going to deprecate the old name, should detect its use and show a warning?

A deprecation notice would complicate the hook implementation, and may not be worth the hassle. A user can transition with a global search and replace.

iainelder:
[My dev notes for the change](https://github.com/iainelder/sceptre-dev-notes/blob/main/dev-notes/update-asg-hook.md).

zaro0508 (commented):
i think we should just remove the old name (`asg_scheduled_action`) in this PR and merge it when we do the next major release (5.0).

iainelder:
That makes sense to me. I suppose when you prepare the major release you can note it as a breaking change.

I've pushed another commit that removed the old name.

How does it look?

zaro0508 (approved):

iainelder:
@alexharv074 requested that it remain compatible with the current release family 4.0.

PR #1393, opened 2023-11-28
Update docker container for CI

zaro0508:
Update circleci docker container with python 3.7 removed and python 3.12 added

depends on https://github.com/Sceptre/sceptre-circleci/pull/19

PR #1394, opened 2023-11-28
replace circleci docker image

zaro0508:
Before setting up Poetry for this project we had to create a custom docker container[1] with all the virtualenvs for tests.  Poetry is now handling the virtualenv for tests so we no longer need to use the custom docker container for running tests, we can just use the stock circleci container.

[1] https://github.com/Sceptre/sceptre-circleci

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ executors:
   docker-publisher:
     environment:
       IMAGE_NAME: sceptreorg/sceptre
-      CIRCLECI_DOCKER_IMAGE: cimg/python:3.12.0
+      CIRCLECI_DOCKER_IMAGE: cimg/python:3.12.0-node
     docker:
       - image: circleci/buildpack-deps:stretch
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,13 +4,14 @@ executors:
   docker-publisher:
     environment:
       IMAGE_NAME: sceptreorg/sceptre
+      CIRCLECI_DOCKER_IMAGE: cimg/python:3.12.0
     docker:
       - image: circleci/buildpack-deps:stretch
 
 aliases:
   - &docs-job
     docker:
-      - image: sceptreorg/sceptre-circleci:2.0.0
+      - image: $CIRCLECI_DOCKER_IMAGE
         environment:
           REPOSITORY_PATH: '/home/circleci/docs'
           DEPLOYMENT_GIT_SSH: 'git@github.com:Sceptre/sceptre.github.io.git'
@@ -34,7 +35,7 @@ aliases:
 jobs:
   build:
     docker:
-      - image: sceptreorg/sceptre-circleci:2.0.0
+      - image: $CIRCLECI_DOCKER_IMAGE
     steps:
       - checkout
       - restore_cache:
@@ -57,7 +58,7 @@ jobs:
 
   unit-tests:
     docker:
-      - image: sceptreorg/sceptre-circleci:2.0.0
+      - image: $CIRCLECI_DOCKER_IMAGE
     steps:
       - checkout
       - restore_cache:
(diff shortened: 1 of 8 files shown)


.pre-commit-config.yaml
```diff
@@ -43,7 +43,7 @@ repos:
         # supported by your project here, or alternatively use
         # pre-commit's default_language_version, see
         # https://pre-commit.com/#top_level-default_language_version
-        language_version: python3.11
+        language_version: python3.10
```
zaro0508, line 46: Not sure why but pre-commit failed on circleci when set to python3.11 
 https://app.circleci.com/pipelines/github/Sceptre/sceptre/2010/workflows/e64f2159-65b5-4b30-9a1d-0ed329fefaf3/jobs/10978

iainelder:
I'm a bit lost now among so many PRs.

Does this change make the [sceptre-circleci repo](https://github.com/Sceptre/sceptre-circleci) and the [PR we just worked on](https://github.com/Sceptre/sceptre-circleci/pull/19) obsolete?

zaro0508:
> Does this change make the [sceptre-circleci repo](https://github.com/Sceptre/sceptre-circleci) and the [PR we just worked on](https://github.com/Sceptre/sceptre-circleci/pull/19) obsolete?

Yes, that is correct @iainelder

.circleci/config.yml
```diff
@@ -10,7 +10,7 @@ executors:
 aliases:
   - &docs-job
     docker:
-      - image: sceptreorg/sceptre-circleci:2.0.0
```
iainelder, line 13: What about using the `cimg/python:3.12` image, which tracks the Python feature version, rather than `cimg/python:3.12.0`, which tracks the bugfix version?

[The push history of the `3.11` tags](https://hub.docker.com/r/cimg/python/tags?page=1&name=3.11&ordering=last_updated) shows that `3.11` and `3.11.6` have the same digest.

iainelder: Doesn't it need the `node` variant?

.pre-commit-config.yaml
iainelder: The error message:

```text
RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.11'
```

The first output of pre-commit shows that, perhaps arbitrarily, pre-commit chose python3.10 because the project doesn't support the default python3.12.0.

```text
The currently activated Python version 3.12.0 is not supported by the project (>=3.8,<3.12).
Trying to find and use a compatible version.
Using python3.10 (3.10.6)
```

Rather than change the language version of black, we should probably pull in some of changes from #1393 that this PR replaces.

* Change pyproject.toml to support Python 3.12
* Upgrade dependencies (at least PyYAML to 6.0.1. I already saw an [error like this ](https://github.com/python-poetry/poetry/issues/8287) using 6.0.0)
* Extend flake8's ignored rules (I don't know why they started flaring up either. We can address it later)

iainelder: I can reproduce the black error locally like this.

```bash
docker run -it cimg/python:3.12.0 bash -c '
git clone https://github.com/Sceptre/sceptre .
git fetch origin pull/1394/head:pr
git switch pr
sed -i "46s/language_version: python3.10/language_version: python3.11/" .pre-commit-config.yaml
poetry install --all-extras -v
poetry run pre-commit run black --all-files --show-diff-on-failure
'
```

zaro0508: done.

.circleci/config.yml
zaro0508: > Doesn't it need the `node` variant?

All of the tests passed so i guess not.

zaro0508:
i didn’t realize that but you are correct @iainelder.  Unit test fails after setting `skip_missing_interpreters = false` in this PR using `cimg/python:3.12`.  It passes when setting `skip_missing_interpreters = false` in PR #1393 using `sceptreorg/sceptre-circleci:2.1.0`.  i guess we do still need the sceptre-circleci docker images, will close this PR and reopen #1393.

PR #1395, opened 2023-12-08
ci: use matrix for unit tests

branchvincent:
Trying out CI changes from https://github.com/Sceptre/sceptre-file-resolver/pull/12

* Replace https://github.com/Sceptre/sceptre-circleci image with CircleCI's [official python image](https://circleci.com/developer/images/image/cimg/python)
* Use a matrix to parallelize unit tests

## PR Checklist

## Other Information


.circleci/config.yml
```diff
@@ -30,68 +27,62 @@ aliases:
...
+      - python/install-packages:
+          pkg-manager: poetry
+          args: --all-extras
       - run:
           name: Unit Tests
           command: |
-            poetry run tox -e clean
-            poetry run tox
-            poetry run tox -e report
+            poetry run pytest --cov=sceptre --cov-report=term-missing --cov-report=html --junitxml=test-reports/junit.xml
```
zaro0508, line 70: now that the CI is no longer running with tox maybe we should change [skip_missing_interpreters](https://github.com/Sceptre/sceptre/blob/master/tox.ini#L4) to `true` ?

.circleci/config.yml
```diff
@@ -198,6 +183,9 @@ workflows:
           filters:
             branches:
               only: /^pull\/.*/
+          matrix:
+            parameters:
+              python: ["3.8", "3.9", "3.10", "3.11", "3.12"]
```
zaro0508, line 188: you wouldn't happen to know a way to keep this in sync with [python versions in tox](https://github.com/Sceptre/sceptre/blob/master/tox.ini#L3) would you?

.circleci/config.yml
```diff
@@ -30,68 +27,62 @@ aliases:
             poetry run ./.circleci/add-known-hosts.sh
             chmod +x .circleci/github-pages.sh
             poetry run ./.circleci/github-pages.sh
+          environment:
+            REPOSITORY_PATH: /home/circleci/docs
+            DEPLOYMENT_GIT_SSH: git@github.com:Sceptre/sceptre.github.io.git
```
zaro0508, line 32: i don't see these ENV Vars used anywhere?

zaro0508 (changes requested):
Just a few minor issues otherwise this is great @branchvincent .  Thanks.

.circleci/config.yml
iainelder: Yes, I agree. Tox is still useful for local unit testing and the skip action makes it easier to use.

I wonder now whether we even need tox for local testing. Can you run a CircleCI test confirguration locally?

branchvincent: updated `skip_missing_interpreters`

> Can you run a CircleCI test confirguration locally?

With Docker [yes](https://circleci.com/docs/how-to-use-the-circleci-local-cli/#run-a-job-in-a-container-on-your-machine), but imo tox is still easier/lighterweight for local dev

.circleci/config.yml
branchvincent: These lines were just moved from above (they're used in the `github-pages.sh` script) https://github.com/Sceptre/sceptre/blob/4361e7d8b0b942406ec663e622647bc875a4faff/.circleci/config.yml#L14-L16

.circleci/config.yml
branchvincent: Agreed that would be nice, but didn't find a simple solution after some searching. Seems like you could use https://circleci.com/docs/dynamic-config/ to generate the parameters from the tox.ini file

branchvincent:
Unpinned the `python` version based on https://github.com/Sceptre/sceptre-file-resolver/pull/12#discussion_r1421032436

.circleci/config.yml
zaro0508: actually, how about if we just run unit tests with tox, `poetry run tox`?  with `skip_missing_interpreters` enabled tox should only run tests on available python versions.  This should have the same affect as pytest except that we would not be ignoring the tox python versions list.

.circleci/config.yml
zaro0508: as commented above, run unit tests thru tox instead of pytest?

zaro0508 (changes requested):

.circleci/config.yml
zaro0508: docs job failed :(   https://app.circleci.com/pipelines/github/Sceptre/sceptre/2034/workflows/885ac577-7398-406f-a33d-3542580761cd/jobs/11094
```
Building docs in /home/circleci/docs/*******.github.io/dev
./.circleci/github-pages.sh: line 63: sphinx-apidoc: command not found
```

iainelder: @zaro0508 the step to install with `--all-extras` has changed, and I think that option is being ignored now. `sphinx-apidoc` is part of the optional sphinx package. I'm still investigating.

zaro0508:
@branchvincent, wondering whether you've given any thought to my idea of running tox instead of pytest?

.circleci/config.yml
iainelder: @branchvincent , why do you want to use `pytest` here instead of `tox`? I thought it was fine to just use `tox` here with `skip_missing_interpreters=true` because of what @zaro0508 said.

.circleci/config.yml
iainelder: For now I think it's easier to just define the list of Python versions in two places instead of trying to figure out how to generate one from the other.

.pre-commit-config.yaml
```diff
@@ -39,11 +39,6 @@ repos:
     rev: 23.1.0
     hooks:
       - id: black
-        # It is recommended to specify the latest version of Python
-        # supported by your project here, or alternatively use
-        # pre-commit's default_language_version, see
-        # https://pre-commit.com/#top_level-default_language_version
-        language_version: python3.11
```
iainelder, line 46: Why don't we need this any more? (I never really understood why we needed it before either.)

.circleci/config.yml
iainelder: I tracked down the step "Install dependencies with poetry using project pyproject.toml" in the [build job 11085](https://app.circleci.com/pipelines/github/Sceptre/sceptre/2034/workflows/885ac577-7398-406f-a33d-3542580761cd/jobs/11085).

The step command is `poetry install --no-ansi --all-extras`. The output shows that it installs the `sphinx` package.

```
  • Installing sphinx (5.1.1)
```

So that's not the problem.

.pre-commit-config.yaml
alexharv074: @iainelder  This is probably a question for @zaro0508 as the revision history shows that he added this line when he set up Poetry in the first place. Considering that we support many versions of Python in this project, I would lean in favour of removing this line altogether as proposed in this PR.

.circleci/config.yml
branchvincent: oops, fixed (I had remove the `restore_cache` step since I didn't think it needed it)

.pre-commit-config.yaml
branchvincent: This was failing in CI since the `executor: python/default` image only has python 3.8. As the comment suggests, it's not needed since the file also specifies `default_language_version` (I'm assuming this snippet was just copy/pasted from [black's docs](https://black.readthedocs.io/en/stable/integrations/source_version_control.html#version-control-integration))

.circleci/config.yml
branchvincent: (sorry just getting back to this)

I think I did this since the container has multiple pythons (one as expected matching the image tag but also an OS-provided one):

```console
$ docker run --rm -it cimg/python:3.12
$ which -a python3
/home/circleci/.pyenv/shims/python3
/usr/bin/python3
/bin/python3
$ which -a python3 | xargs -I{} echo {} -V | sh
Python 3.12.1
Python 3.10.6
Python 3.10.6
```

So just `poetry run tox` would run tests for both versions (which isn't ideal since we may see a failing `ci/circleci: unit-tests-3.12` job and learn that actually it was a failure for 3.10). So we'd want `poetry run tox -e py312`, let me see what that would look like

.circleci/config.yml
iainelder: Glad it was something "simple" like the cache. I had resorted to reading the Oython orb source to figure out what was going on.

The docs job looks like a good one to simplify after we complete this work.

thanks!

.circleci/config.yml
branchvincent: updated to use `tox`

branchvincent:
@zaro0508 sorry for the delay, I updated to use tox and should have also fixed the docs job

.circleci/config.yml
```diff
@@ -9,89 +12,87 @@ executors:
...
+      - run: poetry self update --no-ansi
+      - python/install-packages:
+          pkg-manager: poetry
+          args: --all-extras
       - run:
           name: Unit Tests
           command: |
-            poetry run tox -e clean
-            poetry run tox
+            poetry run tox -e py$(echo <<parameters.python>> | tr -d .)
```
iainelder, line 75: I think `tr -d .` deletes `.` characters. Why do you need to do that? (I don't remember what tox's output looks like.)

iainelder: Ah, it's in the command substitution. Does it strip the dot from `3.8` to make it `38`?

branchvincent: yup exactly!

.pre-commit-config.yaml
iainelder: It's "black" magic from the documentation, because it doesn't tell you why you would want to override the Python version.

Yes, it makes sense to remove this then.

iainelder:
The [build-and-test](https://app.circleci.com/pipelines/github/Sceptre/sceptre/2054/workflows/c8d78c15-5539-4b0a-a5f7-a9892ca6bbb5) workflow of [CircleCI pipeline 2054 suceeded](https://app.circleci.com/pipelines/github/Sceptre/sceptre/2054).

This PR doesn't appear to trigger the "build-test-and-deploy" workflow.

How do we trigger "build-test-and-deploy"? We need that one to check whether the docs job is fixed.

.circleci/config.yml
zaro0508, line 75: would be great if we can add some comments here describing why this is needed. it would help give context to others debugging or updating the CI workflow.

zaro0508 (approved):

.circleci/config.yml
```diff
@@ -9,89 +12,87 @@ executors:
...
-          key: sceptre-v1-{{ checksum "pyproject.toml" }}-{{ checksum "poetry.lock" }}-{{ checksum ".pre-commit-config.yaml" }}-{{ checksum "tox.ini" }}
+          key: sceptre-v1-{{ checksum "poetry.lock" }}-{{ checksum "tox.ini" }}
+      # Update poetry for https://github.com/python-poetry/poetry/issues/7184
+      - run: poetry self update --no-ansi
+      - python/install-packages:
+          pkg-manager: poetry
+          args: --all-extras
       - run:
           name: Unit Tests
           command: |
```
iainelder, line 74: I attempt to explain it. Does it make sense? Can you make it more concise?

```suggestion
          # Normally a single tox invocation runs the tests on all available
          # Python versions.
          #
          # Now CircleCI runs parallel test pipelines, one for each supported
          # Python version.
          #
          # See `matrix.parameters.python` for the Python version values. Each
          # pipeline uses that value as its default Python version.
          #
          # There is one tox invocation for each pipeline. The `-e` option runs
          # tox on just the default Python for that pipeline.
          #
          # The container may make more Python versions available, and we don't
          # want tox to run the tests on those because they may cause false
          # failures.
          #
          # For example, if the container for Python 3.12 would use Python 3.12
          # by default. It make also have an earlier Python version installed
          # such as 3.10 that was used for bootstrapping the container image.
          # Configured this way, tox ignores the version 3.10 in the pipeline
          # for version 3.12.
          command: |
```

.circleci/config.yml
iainelder: [See my comment suggestion on the line just above](https://github.com/Sceptre/sceptre/pull/1395/files/b6fa015e4c430eda7252e1f7839bf55db5662fab#r1433045343).

iainelder (approved):
[The latest pipeline for the `build-test-and-deploy` workflow succeeded](https://app.circleci.com/pipelines/github/Sceptre/sceptre/2056/workflows/787a1571-44ae-4705-8212-8411bac0bb41).

LGTM!

iainelder:
For your info, @branchvincent :

To trigger the `build-test-and-deploy` workflow @zaro0508 fetches the PR branch into the main repo as a normal branch. Someone with write permissions on the main repo can trigger that workflow.

The `build-test-and-deploy` workflow reads secrets to run the `integration-test` job. So you also need permission to read the secrets to run the complete workflow.

Thanks for all your help here!

zaro0508:
great job @branchvincent and thanks a bunch!

branchvincent:
@iainelder @zaro0508 happy to help and thank you for the thoughtful reviews!

PR #1398, opened 2023-12-17
Clean up in Makefile and CONTRIBUTING.md

alexharv074:
This does some clean up in the `Makefile` and `CONTRIBUTING.md` to bring these two into alignment.

Notably:
- Fixes up incomplete `.PHONY` declaration.
- Updates `CONTRIBUTING.md` to add some additional commands that are helpful during development and debugging.
- Adds `ARGS` to `Makefile` to allow extra args to be passed to poetry as required.
- Adds instructions on how to temporarily add the Python debugger for debugging purposes.

[Your PR description here]

## PR Checklist

- [x] Wrote a good commit message & description [see guide below].
- [ ] Commit message starts with `[Resolve #issue-number]`.
- [ ] Added/Updated unit tests.
- [ ] Added/Updated integration tests (if applicable).
- [x] All unit tests (`poetry run tox`) are passing.
- [x] Used the same coding conventions as the rest of the project.
- [ ] The new code passes pre-commit validations (`poetry run pre-commit run --all-files`).
- [x] The PR relates to _only_ one subject with a clear title.
      and description in grammatically correct, complete sentences.

## Approver/Reviewer Checklist

- [ ] Before merge squash related commits.

## Other Information

[Guide to writing a good commit](http://chris.beams.io/posts/git-commit/)

PR #1403, opened 2023-12-27
Update documentation for docs

zaro0508:
Update the docs to show contributors how to generate Sceptre docs using poetry instead of make. The CI does not run the `make` command and neither do developers anymore.  All build, test, and deployment tasks use poetry and all the tasks have been documented therefore in the interest of making the project less of a burden to maintain we remove the makefile.

Note that the docs/Makefile is used to generated docs and that one is still needed, the makefile in the root folder is the one that's no longer needed and is being removed. 

depends on #1398


CONTRIBUTING.md
```diff
@@ -145,9 +144,11 @@ documentation in HTML format.
 To build the documentation locally:
 
 ```bash
-$ make docs
+  $ poetry run make html --directory docs
```
jfalkenstein, line 147: How does this work if the makefile doesn't exist?

zaro0508: ohh right, poetry is running make, however all of the dependencies are provided by poetry (in pyproject.toml) so the make tool is being used however the makefile is not used therefore it's not needed

jfalkenstein: ah. Ok, that makes more sense.

jfalkenstein (approved):

PR #1416, opened 2024-01-16
Refactor GH workflows

zaro0508:
Refactor Github action from a single workflow to multiple workflows. The purpose is to make the workflows easier to follow and understand.  Explanation of workflows..

* check: runs before a merge (i.e. PRs) or after a push (merge or forced push). Mainly for running quick validation jobs.
* gate: runs after a merge or a forced push to any branch.  Allows admins to trigger long running integration tests as a gate to merging into the main branch.
* release: runs on a tag to publish Sceptre to registries and repositories


.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,47 @@
+name: integration-tests
+
+env:
+  AWS_REGION: us-east-1
+  AWS_ROLE_DURATION: 3600
+  # role generated from https://github.com/Sceptre/sceptre-aws/blob/master/config/prod/gh-oidc-sceptre-tests.yaml
+  AWS_ROLE: arn:aws:iam::743644221192:role/gh-oidc-sceptre-tests
+
+on:
+  workflow_run:
```
zaro0508, line 10: I think there are problems with setting up [workflow dependencies](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run).

Even when setting the [github.event.workflow_run.conclusion](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow) flag it does not gather up conclusions from all dependent workflows as suggested in this discussion  https://github.com/orgs/community/discussions/26238

.github/workflows/staging.yaml
```diff
@@ -0,0 +1,34 @@
+name: staging
+
+on:
+  workflow_run:
```
zaro0508, line 4: I think there are problems with setting up [workflow dependencies](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run).

Even when setting the [github.event.workflow_run.conclusion](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow) flag it does not gather up conclusions from all dependent workflows as suggested in this discussion  https://github.com/orgs/community/discussions/26238

Not to mention that having to set `github.event.workflow_run.conclusion` is kinda ugly and is one of the exact things we wanted to avoid for this refactor.

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,47 @@
+name: integration-tests
+
+env:
+  AWS_REGION: us-east-1
```
tarkatronic, line 4:
```suggestion
  AWS_REGION: eu-west-1
```
As noted in #1420 this is where integration tests need to be run

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,47 @@
...
+    workflows:
+      - tests
+    branches: "*"
+    types:
+      - completed
+
+jobs:
+  integration-tests:
+    needs:
+      - unit-tests
```
tarkatronic, line 20: Will this dependency work between separate workflow files? Since this workflow is triggered only by the completed `tests` workflow run, it would only need a single conditional:
```suggestion
    if: ${{ github.event.workflow_run.conclusion == 'success' }}
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,45 @@
...
+          tags: sceptreorg/sceptre:${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+  pypi-release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install Poetry
+        uses: snok/install-poetry@v1
+      - name: Install dependencies
+        run: poetry install --no-interaction --all-extras
```
tarkatronic, line 43: This shouldn't be strictly necessary for a release should it? The `publish` would be all you really need to do.

.github/workflows/staging.yaml
tarkatronic: I'm curious, what's the purpose of this particular workflow? It seems to be basically a duplicate of the `release`. Or is this just meant to push non-release builds to Docker Hub?

Basically I'm wondering, how essential is this particular workflow, and how hard do we need to try debugging this particular problem?

tarkatronic (commented):
Just a few comments and questions!

.github/workflows/staging.yaml
zaro0508: the purpose is to push a docker container with an `edge` tag. edge tag refers to the latest from master.  this would allow users to run sceptre with the latest code (from master branch).

.github/workflows/integration-tests.yaml
zaro0508: as i understand it the result of `github.event.workflow_run.conclusion` is from it's dependency defined by `needs` param.

.github/workflows/release.yaml
zaro0508: hmm, maybe?  i'll test and see.

zaro0508:
i've made quite a few changes to PR #1420 since your review @tarkatronic.  let's get that merged first and i'll redo this PR.  With the job dependencies setup in #1420 i'm still not sure it's possible or even worth it to split this setup into multiple workflows.

.github/workflows/push.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+on:
+  workflow_run:
+    workflows:
+      - check
+    types:
+      - completed
+  push:
+    branches:
+      - '*'
```
zaro0508, line 17: this workflow isn't getting triggered from my testing.  i'm not sure whether it's because this snippet isn't working or whether this workflow needs to be bootstrapped before in GH actions before it will work.

zaro0508:
@tarkatronic i've redone this PR with updated files.  could you please take another look?

.github/workflows/push.yaml
zaro0508: i've tested and verified it should be..
```
on:
  workflow_run:
    workflows:
      - check
    types:
      - completed
    branches:
      - master
```

.github/workflows/check.yaml
```diff
@@ -0,0 +1,79 @@
...
+      envlist: ${{ steps.generate-envlist.outputs.envlist }}
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+      - name: Install Poetry
+        uses: snok/install-poetry@v1
+      - name: Install dependencies
+        run: poetry install --no-interaction --all-extras
+      - id: generate-envlist
+        run: poetry run tox --gh-matrix
```
tarkatronic, line 45: Just curious, is it actually necessary to install everything here? Wouldn't it be sufficient to install just `tox` and the `tox-gh-matrix` plugin?

Not an absolutely necessary change, but might speed this up a bit!

tarkatronic (approved):
Looks great!

Fun fact for the future, it may be possible to do away with the `PYPI_API_TOKEN` secret, using PyPI's "Trusted Publishers" setup: https://docs.pypi.org/trusted-publishers/

.github/workflows/check.yaml
zaro0508: yes, you are correct. we can just `pip install tox` then run `tox --gh-matrix` however I think i prefer to run everything thru poetry so i think it's probably just better to leave as is.

zaro0508:
> Fun fact for the future, it may be possible to do away with the PYPI_API_TOKEN secret, using PyPI's "Trusted Publishers" setup: https://docs.pypi.org/trusted-publishers/

Yeah, i notice that.  The problem with that setup is that it needs to be configure every time we create a new repo (think of sceptre plugin) in this org which means that it's another manual bootstrapping step.  Tokens on the other hand can be shared at the org level so less manual setup.

tarkatronic (approved):

PR #1418, opened 2024-01-17
[Resolves #1417] Add FormattedYaml template handler

### LINKED ISSUE #1417: Provide a formatted_yaml template handler
### Subject of the issue
When generating templates from Jinja2, it is difficult and time-consuming to use Jinja2 line trimming to produce a neatly-formatted output template. This often leads to messy code seen in commands like `sceptre generate`.

It is proposed to add a template handler `formatted_yaml` that subclasses the `File` template handler and reformats white spacing and line breaks in a standard way.

alexharv074:
This adds the `formatted_yaml` template handler that subclasses the `file` template handler to reformat the YAML in a clean and standard way.

Note that this feature was written by @mrowlingfox  

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/template_handlers/formatted_yaml.py
+++ b/sceptre/template_handlers/formatted_yaml.py
@@ -1,0 +1,68 @@
+"""
+This module contains a template handler that can load files from disk and
+format the content.
+"""
+import re
+import os
+from sceptre.template_handlers.file import File
+
+
+class FormattedYaml(File):
+    """
+    Template handler that can load files from disk.
+    Supports JSON, YAML, Jinja2 and Python.
+
+    This class extends the `File` class from the
+    `sceptre.template_handlers.file` module. It provides additional
+    functionality to format YAML content by removing trailing whitespaces
+    and blank lines, adding line breaks before lines that don't start with
+    spaces, tabs, or specific prefixes, and adding a newline at the end
+    of the file.
+    """
+
+    # Class attribute for line start exceptions
+    no_newline_prefixes = ["---", "AWSTemplateFormatVersion", "Description"]
+
+    def handle(self):
+        """
+        Handles the template by returning the template contents.
+
+        Returns:
+            str or bytes: The template contents.
+        """
+        contents = super().handle()
+        if isinstance(contents, str):
+            return self._format_yaml_content(contents)
+        return contents
+
+    def _format_yaml_content(self, input_string: str) -> str:
+        """
+        Formats the input string by removing trailing whitespaces
+        and blank lines, adding a line break before every line that
+        doesn't start with spaces, tabs, or specific prefixes, and adding
+        a newline at the end of the file.
+
+        Args:
+            input_string (str): The input string that needs to be formatted.
+
+        Returns:
+            str: The formatted string.
+        """
+        # Remove trailing whitespaces and blank lines
+        result = re.sub(r"^\s*\n", "", input_string, flags=re.MULTILINE)
+        result = re.sub(r"\s+$", "", result, flags=re.MULTILINE)
+
+        # Construct the pattern for line start exceptions
+        exceptions_pattern = "|".join(
+            re.escape(prefix) for prefix in self.no_newline_prefixes
+        )  # noqa: E501  pylint: disable=line-too-long
+        pattern = r"^(?!" + exceptions_pattern + r")([^\s\t].*)$"
+
+        # Add a line break before every line that doesn't start with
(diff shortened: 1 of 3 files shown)


alexharv074:
An example of the output is given in the tests here https://github.com/Sceptre/sceptre/pull/1418/files#diff-a6582d117234ff569fb21051fba54957585389bec88b5965066b1e3c404e43e9R53-R70

sceptre/template_handlers/formatted_yaml.py
```diff
@@ -0,0 +1,68 @@
...
+            str: The formatted string.
+        """
+        # Remove trailing whitespaces and blank lines
+        result = re.sub(r"^\s*\n", "", input_string, flags=re.MULTILINE)
+        result = re.sub(r"\s+$", "", result, flags=re.MULTILINE)
+
+        # Construct the pattern for line start exceptions
+        exceptions_pattern = "|".join(
+            re.escape(prefix) for prefix in self.no_newline_prefixes
+        )  # noqa: E501  pylint: disable=line-too-long
```
ericabrauer, line 58: Since the line seems to have been adjusted, can this noqa be removed?

sceptre/template_handlers/formatted_yaml.py
```diff
@@ -0,0 +1,68 @@
...
+        pattern = r"^(?!" + exceptions_pattern + r")([^\s\t].*)$"
```
ericabrauer, line 59: Is there any benefit to using an f-string here?

ericabrauer (approved):
Looks good to me! I just had the one question on noqa on the actual handler. 

Thanks for putting this together!

sceptre/template_handlers/formatted_yaml.py
```diff
@@ -0,0 +1,68 @@
...
+
+        Returns:
+            str or bytes: The template contents.
+        """
+        contents = super().handle()
+        if isinstance(contents, str):
+            return self._format_yaml_content(contents)
+        return contents
+
+    def _format_yaml_content(self, input_string: str) -> str:
```
zaro0508, line 38: hmm, what if someone wants their own custom format?

sceptre/template_handlers/formatted_yaml.py
```diff
@@ -0,0 +1,68 @@
...
+    def handle(self):
+        """
+        Handles the template by returning the template contents.
+
+        Returns:
+            str or bytes: The template contents.
+        """
+        contents = super().handle()
+        if isinstance(contents, str):
+            return self._format_yaml_content(contents)
```
zaro0508, line 35: shouldn't this formatting only happen for jinja based templates?

sceptre/template_handlers/formatted_yaml.py
zaro0508: we use flake8 and the setting is [max-line-length = 120](https://github.com/Sceptre/sceptre/blob/master/.flake8#L16) which seems like it should not be too long?

zaro0508 (changes requested):
Thanks for the contribution.  While this is a good idea I think this handler is a bit opinionated and doesn't provide much more functionality than the File handler itself.  I think this handler might be better suited living outside of Sceptre core.  We are happy to setup a repo in the Sceptre org for this handler.  With that being said I'm open to discussion and possibily reconsidering my stance on inclusion of this handler into this repo.

sceptre/template_handlers/formatted_yaml.py
alexharv074: I guess it should happen for anyone who chooses to call the `formatted_yaml` template handler. The intended use case is Jinja-based templates although in theory it could work on any YAML templates. I agree though that some more sanity-checking here would not hurt. I'll have a think.

sceptre/template_handlers/formatted_yaml.py
alexharv074: Having had some bad experiences in the matter in the past, I would probably advise them not to waste time on it! However, the option would be there to just subclass File as we have done and write their own template handler. Or perhaps better to just use the standard File template handler if specific output formatting is desired.

sceptre/template_handlers/formatted_yaml.py
alexharv074: Yes seems like Black was smart enough to reformat the code but not smart enough to remove the noqa bit!

alexharv074:
To be created as a separate project.

PR #1425, opened 2024-01-23
Remove an external resolver integration test

zaro0508:
An AWS profile was used to run integration tests in circleci.  We have moved to github actions and we've also moved to using GH OIDC instead of an AWS profile.  A test that validates the use of a profile was passing in circleci but is now failing in GH actions.  Since we no longer use a profile we cannot run tests to validate the usage of profiles therefore it has been removed.


alexharv074 (approved):
LGTM

PR #1427, opened 2024-01-24
reconfigure pre-commit validation

zaro0508:
We have setup pre-commit.ci[1] to offload the execution of pre-commit. This is supposed to be faster and easier to maintaine than running it with GH actions.

[1] https://pre-commit.ci/


tarkatronic (approved):
This is awesome, I love it!

branchvincent (approved):

PR #1430, opened 2024-01-30
CI: add integration test workflow

zaro0508:
Setup a workflow to trigger an integration test run from a review comment.

Integration tests take a long time to complete therefore it is typically only run as a final check before merging to the master branch.

The idea is to allow users to trigger an integration test run only after all other tests have passed AND a human review approval has been provided. Adding an `/integration-tests` string to a PR review will trigger the integration test run.

### DIFF
--- a/.github/workflows/comment-integration-tests.yaml
+++ b/.github/workflows/comment-integration-tests.yaml
@@ -1,0 +1,12 @@
+name: comment-integration-tests
+
+on:
+  pull_request_review:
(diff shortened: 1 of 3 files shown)


zaro0508:
unfortunately this doesn't work because github action does not allow access to the github token when triggered from a PR.  The workflow fails with.. "Credentials could not be loaded, please check your action inputs: Could not load credentials from any providers"

PR #1438, opened 2024-02-07
Remove classifiers

zaro0508:
Removing classifiers in poetry.toml because poetry will auto generate the classifiers[1]

[1] https://python-poetry.org/docs/pyproject/#classifiers

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,17 +10,6 @@ authors = ["Sceptre <sceptreorg@gmail.com>"]
 description = "An AWS Cloud Provisioning Tool"
 keywords = ["aws", "cloud", "devops", "infrastructure", "tools", "cli"]
 license = "Apache-2.0"
-classifiers = [
-  "Development Status :: 5 - Production/Stable",
-  "Intended Audience :: Developers",
-  "Natural Language :: English",
-  "Environment :: Console",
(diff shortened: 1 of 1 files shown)


tarkatronic (approved):
Should we remove the `Environment` classifier? I'm not certain that one can be auto-generated either.

zaro0508:
> Should we remove the `Environment` classifier? I'm not certain that one can be auto-generated either.

hmm, the docs just has a general statement that it will auto generate `classifiers`.  how do you know which ones it can and cannot auto generate?

tarkatronic:
> > Should we remove the `Environment` classifier? I'm not certain that one can be auto-generated either.
> 
> hmm, the docs just has a general statement that it will auto generate `classifiers`. how do you know which ones it can and cannot auto generate?

From my understanding of that statement, only the `License` and `Programming Language :: Python :: <version>` classifiers will be auto-generated.

zaro0508:
> From my understanding of that statement, only the `License` and `Programming Language :: Python :: <version>` classifiers will be auto-generated.

ok, so how does poetry infer the license without the classifier? does it look in the root folder for a license file?

zaro0508:
i've fixed the classifier.  take another look @tarkatronic ?

tarkatronic (approved):
👍

PR #1445, opened 2024-02-25
[Resolves #1187] Add configuration for inheritance strategies for list and dict configs

### LINKED ISSUE #1187: feature request: append stack_tag inheritance
Currently stack_tags inheritance strategy is override which means that stack tags can be defined in a Sceptre StackGroup and then overridden in the a child StackGroup or Stack config.    I would like to make a feature request to append stack_tags on a dependencies. 

Example:
Assume Scepter project A, B and C with dependency C -> B 

root StackConfig config/config.yaml
```
stack_tags:
  - country: US
```

config/prod/A.yaml
```
stack_tags:
  - city: anaheim    # stack A tags are 'city:anaheim', 'country:US'
```

config/prod/C.yaml
```
dependencies:
  - prod/B.yaml
stack_tags:
  - county: collin   # stack C tags are 'county:collin', 'country:US'
 ``` 

config/prod/B.yaml
```
stack_tags:
  - city: boston    # stack B tags are 'city:boston', 'county:collin', 'country:US'
```

### PR DESCRIPTION
**Pull request by okcleary**
Add configuration for inheritance strategies for list and dict configs.

This resolves #1187 by allowing selecting a dict_merge strategy at a stack group/stack level, while remaining backwards compatible with existing stacks by retaining the default of child_wins.

This does not however address the example from the original issue of including tags of dependencies (or dependents?), as I don't believe this fits with the existing model (or makes sense as a use case).

This PR additionally adds the same support to all list and dict based configs.

The config for inheritance strategy itself is also inheritable, but not configurable.

From the original issue example:

Example:
Assume Sceptre project A, B and C with dependency C -> B

root StackConfig config/config.yaml

```
stack_tags_inheritance: dict_merge
stack_tags:
  - country: US
```

config/prod/A.yaml

```
stack_tags:
  - city: anaheim    # stack A tags are 'city:anaheim', 'country:US'
```

config/prod/C.yaml

```
dependencies:
  - prod/B.yaml
stack_tags:
  - county: collin   # stack C tags are 'county:collin', 'country:US'
```

config/prod/B.yaml

```
stack_tags:         # In contrast to original issue example, dependencies do not impact inherited values
  - city: boston    # stack B tags are 'city:boston', 'country:US'
```

## PR Checklist

## Other Information

### DIFF
--- a/docs/_source/docs/stack_config.rst
+++ b/docs/_source/docs/stack_config.rst
@@ -85,7 +85,7 @@ dependencies
 ~~~~~~~~~~~~
 * Resolvable: No
 * Can be inherited from StackGroup: Yes
-* Inheritance strategy: Appended to parent's dependencies
+* Inheritance strategy: Appended to parent's dependencies. Configurable with ``dependencies_inheritance`` parameter.
 
 A list of other Stacks in the environment that this Stack depends on. Note that
 if a Stack fetches an output value from another Stack using the
@@ -110,13 +110,15 @@ dependencies_inheritance
 
 This configuration will override the default inheritance strategy of `dependencies`.
 
-Valid values for this config are: ``list_join``, and ``child_wins``.
+The default value for this is ``list_join``.
+
+Valid values for this config are: ``list_join``, or ``child_wins``.
 
 hooks
 ~~~~~
 * Resolvable: No (but you can use resolvers _in_ hook arguments!)
 * Can be inherited from StackGroup: Yes
-* Inheritance strategy: Overrides parent if set
+* Inheritance strategy: Overrides parent if set. Configurable with ``hooks_inheritance`` parameter.
 
 A list of arbitrary shell or Python commands or scripts to run. Find out more
 in the :doc:`hooks` section.
@@ -129,7 +131,9 @@ hooks_inheritance
 
 This configuration will override the default inheritance strategy of `hooks`.
 
-Valid values for this config are: ``list_join``, and ``child_wins``.
+The default value for this is ``child_wins``.
+
+Valid values for this config are: ``list_join``, or ``child_wins``.
 
 ignore
 ~~~~~~
@@ -180,7 +184,7 @@ notifications
 
 List of SNS topic ARNs to publish Stack related events to. A maximum of 5 ARNs
 can be specified per Stack. This configuration will be used by the ``create``,
-``update``, and ``delete`` commands. More information about Stack notifications
+``update``, or ``delete`` commands. More information about Stack notifications
 can found under the relevant section in the `AWS CloudFormation API
 documentation`_.
 
@@ -256,7 +260,7 @@ parameters
 ~~~~~~~~~~
 * Resolvable: Yes
 * Can be inherited from StackGroup: Yes
-* Inheritance strategy: Overrides parent if set
+* Inheritance strategy: Overrides parent if set. Configurable with ``parameters_inheritance`` parameter.
 
 .. warning::
 
@@ -266,7 +270,7 @@ parameters
    environment variable resolver.
 
 A dictionary of key-value pairs to be supplied to a template as parameters. The
-keys must match up with the name of the parameter, and the value must be of the
+keys must match up with the name of the parameter, or the value must be of the
 type as defined in the template.
 
 .. note::
@@ -325,7 +329,9 @@ parameters_inheritance
 
 This configuration will override the default inheritance strategy of `parameters`.
 
-Valid values for this config are: ``dict_merge``, and ``child_wins``.
+The default value for this is ``child_wins``.
+
+Valid values for this config are: ``dict_merge``, or ``child_wins``.
 
 protected
 ~~~~~~~~~
@@ -426,7 +432,7 @@ sceptre_user_data
 ~~~~~~~~~~~~~~~~~
 * Resolvable: Yes
 * Can be inherited from StackGroup: Yes
-* Inheritance strategy: Overrides parent if set
+* Inheritance strategy: Overrides parent if set. Configurable with ``sceptre_user_data_inheritance`` parameter.
 
 Represents data to be passed to the ``sceptre_handler(sceptre_user_data)``
 function in Python templates or accessible under ``sceptre_user_data`` variable
@@ -440,7 +446,9 @@ sceptre_user_data_inheritance
 
 This configuration will override the default inheritance strategy of `sceptre_user_data`.
 
-Valid values for this config are: ``dict_merge``, and ``child_wins``.
+The default value for this is ``child_wins``.
+
+Valid values for this config are: ``dict_merge``, or ``child_wins``.
 
 stack_name
 ~~~~~~~~~~
@@ -481,7 +489,7 @@ stack_tags
 ~~~~~~~~~~
 * Resolvable: Yes
 * Can be inherited from StackGroup: Yes
-* Inheritance strategy: Overrides parent if set
+* Inheritance strategy: Overrides parent if set. Configurable with ``stack_tags_inheritance`` parameter.
 
 A dictionary of `CloudFormation Tags`_ to be applied to the Stack.
 
@@ -493,7 +501,9 @@ stack_tags_inheritance
 
 This configuration will override the default inheritance strategy of `stack_tags`.
 
-Valid values for this config are: ``dict_merge``, and ``child_wins``.
+The default value for this is ``child_wins``.
+
+Valid values for this config are: ``dict_merge``, or ``child_wins``.
 
 stack_timeout
 ~~~~~~~~~~~~~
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: docs/_source/docs/stack_group_config.rst [hunk 1]
```diff
@@ -185,6 +185,17 @@ For example, if you wanted the ``dev`` StackGroup to build to a different
...
+Inheritance Strategy Override
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The inheritance strategy of some properties may be overridden by the stack group config.
+
+Strategy options:
+
+ - ``list_join``: Child configs are appended to parent configs.
+ - ``dict_merge``: Child configs will be merged with parent configs, the child keys taking precedence.
+ - ``child_wins``: Overrides parent if set.
```
okcleary, line 197: I chose to use the function names directly as the options, but it's possible a more "user friendly" might be better.

sceptre/config/reader.py
```diff
@@ -352,11 +374,8 @@ def _read(self, rel_path, base_config=None):
 
         # Parse and read in the config files.
         this_config = self._recursive_read(directory_path, filename, config)
-
-        if "dependencies" in config or "dependencies" in this_config:
-            this_config["dependencies"] = CONFIG_MERGE_STRATEGIES["dependencies"](
-                this_config.get("dependencies"), config.get("dependencies")
-            )
+        # Apply merge strategies with the config that includes base_config values.
+        this_config.update(self._get_merge_with_stratgies(config, this_config))
```
okcleary, line 378: "dependencies" is handled in the common merge function.

This change is safe because all default strategies except for dependencies are `child_wins`, so behaviour doesn't change without the user selecting a strategy.

tests/test_config_reader.py
```diff
@@ -295,7 +297,7 @@ def test_construct_stacks_constructs_stack(
             sceptre_user_data={},
             hooks={},
             s3_details=sentinel.s3_details,
-            dependencies=["child/level", "top/level"],
+            dependencies=["top/level", "child/level"],
```
okcleary, line 300: This changed simply because the previous merge call for dependencies used the opposite order of inputs to the strategy function compared to the one done in `_recursive_read`.

alexharv074:
Thanks for this contribution @okcleary I have checked it out and will test it locally for a while to make sure everything works for us.

alexharv074:
@okcleary Thanks for your contribution. This looks like a great feature and we have discussed in the Slack channel that I am currently running your branch locally to make sure nothing breaks for me. Could you help me to review this PR by explaining to me how we can be sure that the change is backwards-compatible and will not break anything for others?

docs/_source/docs/stack_config.rst
```diff
@@ -106,6 +121,16 @@ hooks
 A list of arbitrary shell or Python commands or scripts to run. Find out more
 in the :doc:`hooks` section.
 
+hooks_inheritance
```
zaro0508, line 124: how about updating the hooks info (above) to reference this key?
 
```
* Inheritance strategy: Overrides parent if set. Configurable with ``hooks_inheritance`` parameter.
```

sceptre/config/reader.py
```diff
@@ -395,16 +414,43 @@ def _recursive_read(
...
+            override_key = f"{config_key}_inheritance"
+            if override_key in CONFIG_MERGE_STRATEGIES:
+                name = CONFIG_MERGE_STRATEGIES[override_key](
+                    left.get(override_key), right.get(override_key)
+                )
+                if not name:
+                    pass
+                elif name not in strategies.SELECTABLE:
+                    raise SceptreException(
+                        f"{name!r} is not a valid inheritance strategy"
```
zaro0508, line 440: i have not seen `!r` before but the doc says that it's not needed in fstrings? https://peps.python.org/pep-0498/#s-r-and-a-are-redundant

zaro0508 (changes requested):
nice addition @okcleary, thanks for the contribution

okcleary:
> @okcleary Thanks for your contribution. This looks like a great feature and we have discussed in the Slack channel that I am currently running your branch locally to make sure nothing breaks for me. Could you help me to review this PR by explaining to me how we can be sure that the change is backwards-compatible and will not break anything for others?

Thanks! There are two points to consider here for the backwards-compatibility of this change.

The first one is the decision to not change any of the existing merge strategies - using a different strategy is optin by the user, and the existing default strategy of each key is untouched by this PR.

The second point is the change to merging the stackgroup config, seen here:

https://github.com/Sceptre/sceptre/pull/1445/files#diff-f46036fa4ab9b5ddee081163999046ae153ec9a37458c96fc312a3e8d6cf53f4L356

Currently only the only key merged is `dependencies` which is done explicitly.  This PR changes it so the merge behaves in the same way as the stack merges - considering all keys that could be mergeable.  This change is backwards compatible because all the other keys besides dependencies have `child_wins` as the default strategy so the behaviour is the same as before.

docs/_source/docs/stack_group_config.rst
okcleary: Yes something like this, although perhaps `merge` instead of `prepend` as I think it will differentiate it better from `append`.

I'm happy to change this over if it's preferred?

sceptre/config/reader.py
okcleary: Hmm, I hadn't seen that section before - I read that as, `!r` is not strictly necessary because you can just do `f"{repr(name)}".

This is a personal habit of mine, the rational here is to clearly present what the incorrect variable value was because it will use the repr of the value.  So for example if the value was an empty string the error would be `"'' is not a valid inheritance strategy"` or if for some reason the value had trailing whitespace `"'dict_merge ' is not a valid inheritance strategy"`.

I'm happy to remove `!r` and just let it be the str of name if that's preferable?

docs/_source/docs/stack_group_config.rst
zaro0508: i think i would prefer friendlier names, however I'm not sure what that would look like exactly.  do we use `merge` instead of `list_join` and `dict_merge`?  if so that might be more confusing since we would be using one word to describe something similar but subtly different?

sceptre/config/reader.py
zaro0508: i think it's more common without thhe `!r` therefore probably easier to understand for most people so i would remove.

docs/_source/docs/stack_group_config.rst
okcleary: I do like that, it seems like a clean UX.  Personally I don't see the dependency in behaviour to be much of a problem - as long as the user has a basic understanding of YAML, because lists and maps simply work differently when combined.

It feels like most words we could use would have the same problem? i.e most common words here conceptually apply to both operations.

zaro0508: ok, lets go with `merge` for both list_join and dict_merge then. thanks

okcleary: I've updated this in the latest commit.

zaro0508 (approved):
LGTM, just a few doc fix ups and it'll be good to go.

docs/_source/docs/stack_group_config.rst
```diff
@@ -192,9 +192,8 @@ The inheritance strategy of some properties may be overridden by the stack group
 
 Strategy options:
 
- - ``list_join``: Child configs are appended to parent configs.
- - ``dict_merge``: Child configs will be merged with parent configs, the child keys taking precedence.
- - ``child_wins``: Overrides parent if set.
+* ``merge``: Child config is merged with parent configs.  For maps child keys take precedence.
```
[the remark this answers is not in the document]
okcleary: Mmm, this was my attempt to not use python terminology - by map I'm referring to a YAML mapping (https://yaml.org/spec/1.2.2/#mapping).

Should I just omit it completely by instead describing it like:

> Child config is merged with parent configs, with child taking precedence for conflicting keys.

sceptre/config/strategies.py
```diff
@@ -71,3 +71,26 @@ def child_wins(a, b):
     :returns: b
     """
     return b
+
+
+def child_or_parent(a, b):
+    """
+    Returns second arg if is not empty, else the first.
```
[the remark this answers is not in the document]
okcleary: Sorry, not sure I follow. Would this be better:

> Returns the the first non-falsy parameter.

docs/_source/docs/stack_group_config.rst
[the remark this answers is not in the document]
zaro0508: that sounds better.  i also think the use of the word `dictionaries` instead of maps would be more recognizable as well.

sceptre/config/strategies.py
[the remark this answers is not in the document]
zaro0508: i think this is more clear? `Returns the second arg if it is not empty, else the first`

okcleary:
Thanks @zaro0508, I've updated with the suggested documentation changes.

zaro0508 (approved):

alexharv074 (approved):
Thanks for the explanation, makes sense @okcleary

PR #1452, opened 2024-03-23
[Resolve #1451] Stop suppressing ClientError in describe_outputs

### LINKED ISSUE #1451: List outputs is eating ExpiredToken messages and possibly others
### Subject of the issue
I have noticed that when my auth token expires, e.g.

```
# aws s3 ls 
An error occurred (ExpiredToken) when calling the AssumeRole operation: The security token included in the request is expired
```

Some Sceptre commands such as `list outputs` eat this and fail silently. Others don't. For example:

```
# sceptre --merge-vars --dir=network --var-file=/app/platform-sceptre/stacks/../data/test/common-env.yaml --var-file=/app/platform-sceptre/stacks/../data/test/network-e2e.yaml launch network.yaml -y 
[2024-03-23 10:08:08] - vpc - Launching Stack
"An error occurred (ExpiredToken) when calling the AssumeRole operation: The security token included in the request is expired"
```

Compared to

```
# sceptre --merge-vars --dir=network --var-file=/app/platform-sceptre/data/test/common-env.yaml --var-file=/app/platform-sceptre/data/test/network.yaml list outputs network.yaml --export=envvar 
# echo $? 
0
```

### Your environment
* version of sceptre (sceptre --version) latest
* version of python (python --version) 3.10
* which OS/distro Ubuntu

### Steps to reproduce
See above.

### Expected behaviour
Should see that the token expired.

### Actual behaviour
No output.

alexharv074:
This resolves a bug in `sceptre outputs` where all `ClientError` exceptions (e.g. `ExpiredToken`) are suppressed, for no apparent reason.

In 682a3061139e8d5f24b63c754164d5a5f8284d50, (October 2018, commit summary "Move Stack operations from Stack to StackActions"), a big refactor had occurred. That refactor had introduced a public method `describe` in the `StackActions` class.

A month later, in 182f0b64f1bfbb439c7e6f7778a3ce0d0774d187 (November 2018, commit summary "Implement algorithm to execute Stacks from StackGraph"), another big refactor occurred. That refactor introduced an essentially duplicate method `_describe` called by `describe_output` method, and then added code to catch all `ClientError` exceptions and return an empty list if one is raised.

So far as I can tell, there was no good reason to introduce either the new method, or catch all `ClientError` exceptions. The suppression of exceptions may have been accidental, a hack to get the unit tests to pass (or for a reason I have not thought of yet!).

Note that at the time of 182f0b64f1bfbb439c7e6f7778a3ce0d0774d187, the `describe_outputs` method that suppressed exceptions was not called anywhere other than in the CLI `outputs` commands:

```
% git co 182f0b64f1bfbb439c7e6f7778a3ce0d0774d187 
% grep -wr describe_outputs sceptre 
sceptre/plan/plan.py:    def describe_outputs(self, *args):
sceptre/plan/plan.py:        self._resolve(command=self.describe_outputs.__name__)
sceptre/plan/actions.py:    def describe_outputs(self):
sceptre/cli/list.py:        in plan.describe_outputs().values() if response
```

This patch attempts to fix all of this up:

- Duplicate method removed in favour of the public `describe` method. I have opted to keep the public method as the differ class is calling it, which means it should be public.

- The Nov 2018 change to `describe_outputs` which introduced the present bug is reverted.

- Duplicate logic to handle the case of "stack does not exist" removed.

- Tweak call to the `describe` method in the differ.

## PR Checklist

## Other Information


tests/test_actions.py
```diff
@@ -515,22 +515,6 @@ def test_delete_with_non_existent_stack(
...
-        )
-
-    def test_describe_events_sends_correct_request(self):
-        self.actions.describe_events()
-        self.actions.connection_manager.call.assert_called_with(
-            service="cloudformation",
-            command="describe_stack_events",
-            kwargs={"StackName": sentinel.external_name},
-        )
-
```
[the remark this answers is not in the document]
zaro0508: can you please tell us which existing tests make these redundant?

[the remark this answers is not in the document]
alexharv074: I think I was dreaming! I have restored the tests.

zaro0508 (commented):
Generally LGTM, however the integration tests are failing for this change.. https://github.com/Sceptre/sceptre/actions/runs/8405338399/job/23307454058

alexharv074:
> Generally LGTM, however the integration tests are failing for this change.. https://github.com/Sceptre/sceptre/actions/runs/8405338399/job/23307454058

I can't remember if this is because I already fixed it or what but I can't see any failing pipelines. All good to merge this now?

zaro0508:
> I can't remember if this is because I already fixed it or what but I can't see any failing pipelines. All good to merge this now?

@alexharv074 please take a look at my last comment, it contains a link to the integration test run.

zaro0508 (changes requested):
@alexharv074 please take a look at my last comment, it contains a link to the integration test run.

alexharv074:
@zaro0508 

> @alexharv074 please take a look at my last comment, it contains a link to the integration test run.

I could be off base here but I do not think these 'describe stack' tests are needed and have tentatively deleted them. The scenarios like "the user describes resources in stack_group" refers to things that the user in fact cannot do and these are in effect integration tests on private methods. The tests were previously passing _because_ the describe method called was swallowing up all exceptions and this is the bug this PR attempts to fix. 

The second problem here is that keeping these tests would require us to now somehow test for exceptions being raised (remembering that these tests are really testing a private method and not something that the user can actually do). It looks like this is possible e.g. https://stackoverflow.com/questions/27894993/handling-exceptions-in-python-behave-testing-framework but would inevitably be a bit hacky.

I personally feel that these code paths are already tested by unit tests and therefore deleting these integration test cases could be acceptable.

integration-tests/features/describe-stack-group.feature
```diff
@@ -1,25 +0,0 @@
-Feature: Describe stack_group
```
zaro0508, line 1: the steps for these tests are in `integration-tests/steps/stacks.py`.  should we remove those as well?

integration-tests/steps/stack_groups.py
```diff
@@ -204,13 +204,6 @@ def step_impl(context, stack_group_name, status):
         assert response in expected_response
 
 
-@then("no resources are described")
```
zaro0508, line 207: there are checks in this file for tests in `integration-tests/features/describe-stack-group-resources.feature`.  are those needed anymore? if not can we remove them?

zaro0508 (changes requested):
Generally LGTM.  Just seeing if there is an opportunity to do more cleanup?

integration-tests/steps/stack_groups.py
alexharv074: Good pickups! Another 154 lines could be deleted.
```
% git diff --cached --stat 
 integration-tests/steps/stack_groups.py | 115 -------------------------------------------------------------------------------------------------------------------
 integration-tests/steps/stacks.py       |  39 ---------------------------------------
 2 files changed, 154 deletions(-)
```

zaro0508 (approved):

PR #1469, opened 2024-06-01
[Resolve #915] Support ChangeSetType CREATE

### LINKED ISSUE #915: Create changeset for new stack
We'd like to create a changeset in our CI/CD pipeline before we execute it, so we can review changes and execute the change set manually. CloudFormation gives us the opportunity to `deploy` without executing via `--no-execute-changeset`, or to create a changeset with either CREATE or UPDATE `--change-set-type`. I don't think Sceptre has this ability? When we create a changeset for a stack that doesn't exist we get:

`"An error occurred (ValidationError) when calling the CreateChangeSet operation: Stack [stack-name-goes-here] does not exist"`

And `launch` doesn't have an equivalent to  `--no-execute-changeset`.

alexharv074:
This adds logic supporting the case of creating change sets for a stack that does not already exist. That is, it adds support for change sets of ChangeSetType CREATE. https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation/client/create_change_set.html

To add some context, there is no obvious reason why Sceptre does not already support change sets of type CREATE since the logic for detecting stack status already exists for the `launch` command. The project's initial commit already has this inconsistency and essentially the same code. It is probably an oversight.

The changes have already been tested by Rohit S in this Slack thread
https://og-aws.slack.com/archives/C01JNN8RGBB/p1716869822031499

In addition, it is necessary to add logic to handle the `REVIEW_IN_PROGRESS` state, both for the case of the `launch` and `create` action. 
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-describing-stacks.html

Creation of a change set on a non-existent stack actually creates a stack and places it into the `REVIEW_IN_PROGRESS` state. If Sceptre is then to `launch` this stack, it is necessary to firstly delete the stack, just as it already does in the cases of states `CREATE_FAILED` and `ROLLBACK_COMPLETE`.

## PR Checklist

## Other Information

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
     hooks:
       - id: yamllint
   - repo: https://github.com/awslabs/cfn-python-lint
-    rev: v0.86.4
+    rev: v1.2.5a9
     hooks:
       - id: cfn-python-lint
         args:
@@ -28,7 +28,7 @@ repos:
         exclude: |
           (?x)(
             ^integration-tests/sceptre-project/config/|
-            ^integration-tests/sceptre-project/templates/jinja/|
+            ^integration-tests/sceptre-project/templates/|
             ^tests/fixtures-vpc/config/|
             ^tests/fixtures/config/|
             ^temp/|
@@ -47,7 +47,7 @@ repos:
         language_version: python3.10
         args: ['--check']
   - repo: https://github.com/sirosen/check-jsonschema
-    rev: 0.28.2
+    rev: 0.28.4
     hooks:
       - id: check-github-workflows
       - id: check-github-actions
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
     hooks:
       - id: yamllint
   - repo: https://github.com/awslabs/cfn-python-lint
-    rev: v1.2.5a9
+    rev: v0.86.4
     hooks:
       - id: cfn-python-lint
         args:
@@ -47,7 +47,7 @@ repos:
         language_version: python3.10
         args: ['--check']
   - repo: https://github.com/sirosen/check-jsonschema
-    rev: 0.28.4
+    rev: 0.28.2
     hooks:
       - id: check-github-workflows
       - id: check-github-actions
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
     hooks:
       - id: yamllint
   - repo: https://github.com/awslabs/cfn-python-lint
-    rev: v0.86.4
+    rev: v1.2.5a9
     hooks:
       - id: cfn-python-lint
         args:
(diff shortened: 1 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -17,7 +17,7 @@ repos:
     hooks:
       - id: yamllint
   - repo: https://github.com/awslabs/cfn-python-lint
-    rev: v1.2.5a9
+    rev: v0.86.4
```
alexharv074, line 20: These changes are necessary from this PR https://github.com/Sceptre/sceptre/pull/1470 I merged that branch in here so that the build could be green.

.pre-commit-config.yaml
```diff
@@ -28,7 +28,7 @@ repos:
         exclude: |
           (?x)(
             ^integration-tests/sceptre-project/config/|
-            ^integration-tests/sceptre-project/templates/jinja/|
+            ^integration-tests/sceptre-project/templates/|
```
zaro0508, line 31: this has already been committed to master.  could you please update this PR with master?

zaro0508 (changes requested):

zaro0508 (changes requested):
Integration tests are failing for this PR, https://github.com/Sceptre/sceptre/actions/runs/9507089977/job/26205762444

alexharv074:
@zaro0508  integration test fixed:
Screenshot 2024-06-14 at 10 27 22 am

sceptre/plan/actions.py
```diff
@@ -212,7 +212,11 @@ def launch(self) -> StackStatus:
 
         if existing_status == "PENDING":
             status = self.create()
-        elif existing_status in ["CREATE_FAILED", "ROLLBACK_COMPLETE"]:
+        elif existing_status in [
```
zaro0508, line 215: the docs for the [create command](https://docs.sceptre-project.org/latest/docs/cli.html?highlight=CHANGE_SET_NAME#sceptre-create) provides info about change sets while the doc and CLI help for the `launch` command does not.  Could you please add some documentation to launch?

zaro0508 (changes requested):
Generally LGTM, just need a little documentation.

sceptre/plan/actions.py
alexharv074: @zaro0508  I've had a look and I'm afraid I'm not sure exactly what the issue is. My understanding is the create command can also create a change set; the launch command I assume cannot. The docs look right to me?

zaro0508: Am I missing something?  The launch command doesn't create the change set however doesn't this enable the launch command to deploy change sets?

alexharv074: @zaro0508 Yes, I think you could be misunderstanding what I am changing here. All that changes is an additional state, `REVIEW_IN_PROGRESS`, is added to the list of possible existing statuses that launch needs to be able to handle. Actually, the fact that it is currently not handled is arguably a bug anyway and this change to the launch method therefore is just a trivial bugfix. 

> The launch command doesn't create the change set however doesn't this enable the launch command to deploy change sets?

No. If launch is used on a stack that has a change set, that stack will exist and be in state `REVIEW_IN_PROGRESS`. The launch command will not execute that change set; it will delete this 'placeholder' stack that gets created by the create changeset command, and then launch a new one in its place.

If the caller doesn't want this, they would instead use the changeset execute command rather than launch.

So all will continue to behave as documented, and as expected.

zaro0508 (approved):

PR #1472, opened 2024-06-04
[Resolves #937] Reset Jinja2 variable cache between config files

### LINKED ISSUE #937: Jinja variables cached between StackGroups

alexharv074:
From https://github.com/NathanWilliams

Fix for: https://github.com/Sceptre/sceptre/issues/937

This resets the Jinja2 templating_vars in config/reader.py to prevent variables leaking across StackGroups.

I am adopting this stale PR:
https://github.com/Sceptre/sceptre/pull/938

## PR Checklist

## Other Information


zaro0508 (approved):
The issue for this hints that using a non deterministic set is the problem... 

```
If a variable is omitted in one stack group's config, the incorrect value from another group is used.
(Note: Sceptre uses a set for its file list, which is non-deterministic, as such the following bug will only be seen on some runs
```

I'm wondering if the better fix is to use an more deterministic data structure, like a ordered set?

alexharv074:
> I'm wondering if the better fix is to use an more deterministic data structure, like a ordered set?

@zaro0508  That's not a fix for this bug but it would have assisted the original team who reported the bug in being able to consistently reproduce it. The nature of the bug here is simply that keys belonging to other stack groups leaked into the configs of stack groups they did not belong to and this would occur with or without ordering. Maybe track a separate issue of making the code a bit more predictable by using ordered sets where applicable?

PR #914, opened 2020-07-01
[Resolves #359] Implements use previous paramater feature

RobReus:
Resolves #359 

This new feature will make it possible to create and maintain stacks using Sceptre while also having the possibility to modify parameters outside of Sceptre, without Sceptre then overwriting/reverting these changes.

An example use-case: create and maintain your containerized environment using Sceptre, but after the initial deploy, maintain the Docker image tag using a CICD pipeline. When configuring the image tag parameter correctly, it will be given an initial value upon creation, but updates will not overwrite the value.

It is as simple as this:
```yaml
parameters:
  Param:
    initial_value: "foobar"
    use_previous_value: true
```

It is fully backwards compatible and will not break existing functionality.

## PR Checklist

- [X] Wrote a good commit message & description [see guide below].
- [X] Commit message starts with `[Resolve #issue-number]`.
- [X] Added/Updated unit tests.
- [X] Added/Updated integration tests (if applicable).
- [X] All unit tests (`make test`) are passing.
- [X] Used the same coding conventions as the rest of the project.
- [X] The new code passes flake8 (`make lint`) checks.
- [X] The PR relates to _only_ one subject with a clear title.
      and description in grammatically correct, complete sentences.

## Approver/Reviewer Checklist

- [ ] Before merge squash related commits.

PR #1471, opened 2024-06-02
[Resolve #1468] Cast parameters to strings

alexharv074:
In already-resolved and merged issue #1443 we implemented parameter validation.

Subsequent discussion led to a consensus that it would be better to not only validate parameters, but also cast where applicable to string. This will lead to a cleaner user experience for most Sceptre users where parameters passed as ints, floats and bools in their YAML values files are cast to strings first and then sent to CloudFormation, since CloudFormation's API is unable to handle other types.

This PR changes the `_ensure_parameters` method to `_cast_parameters` in the Stack class to handle casting to strings where applicable instead of only validating.

## PR Checklist

- [x] Wrote a good commit message & description [see guide below].
- [x] Commit message starts with `[Resolve #issue-number]`.
- [x] Added/Updated unit tests.
- [ ] Added/Updated integration tests (if applicable).
- [x] All unit tests (`poetry run tox`) are passing.
- [x] Used the same coding conventions as the rest of the project.
- [x] The new code passes pre-commit validations (`poetry run pre-commit run --all-files`).
- [x] The PR relates to _only_ one subject with a clear title.
      and description in grammatically correct, complete sentences.

## Approver/Reviewer Checklist

- [ ] Before merge squash related commits.

## Other Information

[Guide to writing a good commit](http://chris.beams.io/posts/git-commit/)

PR #1476, opened 2024-06-16
[Resolve #359] Support use-previous-parameters

### LINKED ISSUE #359: Don't override existing parameters
```
[failed] create change set custom/oar-ecs-cluster/eu-west-1/ecs-cluster
An error occurred (ValidationError) when calling the CreateChangeSet operation: Parameters: [NexusDockerAuth] must have values
```

We've manually set a `NoEcho` parameter because it contains a secret key. As your documents highlight, we shouldn't commit the value to configuration.

It'd be really helpful if sceptre didn't try and override parameter values if they are not provided, that way we can just lean on the values stored in AWS.

Currently we'd have to fiddle around with CLI arguments and must then store secrets in both AWS and wherever we're running secptre.

alexharv074:
I am adopting this stale PR by @RobReus 
https://github.com/Sceptre/sceptre/pull/914

Resolves https://github.com/Sceptre/sceptre/issues/359

This new feature will make it possible to create and maintain stacks using Sceptre while also having the possibility to modify parameters outside of Sceptre, without Sceptre then overwriting/reverting these changes.

An example use-case: create and maintain your containerized environment using Sceptre, but after the initial deploy, maintain the Docker image tag using a CICD pipeline. When configuring the image tag parameter correctly, it will be given an initial value upon creation, but updates will not overwrite the value.

It is as simple as this:

```yaml
parameters:
  Param:
    initial_value: "foobar"
    use_previous_value: true
```

It is fully backwards compatible and will not break existing functionality.

## PR Checklist

## Other Information


alexharv074:
@RobReus  Can you recall the extent to which you tested this feature yourself, besides the unit tests you have added?

alexharv074:
Some of this logic around validating the parameters would need to be moved to `_ensure_parameters` inside the `Stack` class.

alexharv074:
To avoid potential rework here, this PR would need to go in first. https://github.com/Sceptre/sceptre/pull/1471

alexharv074:
I have changed my mind about merging this considering that it adds some complexity in the code we have no known users who would benefit from the feature.

PR #1494, opened 2024-08-03
[Resolve #1493] Complete DisableRollback implementation

### LINKED ISSUE #1493: DisableRollback not fully implemented in stack updates
### Subject of the issue
DisableRollback is handled only during the creation of a stack and not for stack updates. This appears to be an oversight. It is already half-implemented in the CLI.

### Your environment
* which OS/distro

### Steps to reproduce

### Expected behaviour
Update stack should handle DisableRollback if passed. 
### Actual behaviour
Nothing.

alex-harvey-z3q:
This adds missing code to complete the implementation of the DisableRollback feature.

The implementation in 99c839a2b began this feature by adding disable rollback to the create and update CLI commands, but it was forgotten to update the stack actions. 

This adds the missing code and another unit test.

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/cli/update.py
+++ b/sceptre/cli/update.py
@@ -41,6 +41,7 @@ def update_command(
     :type verbose: bool
     :param yes: A flag to answer 'yes' to all CLI questions.
     :type yes: bool
+    :param disable_rollback: A flag to disable cloudformation rollback.
     """
 
     context = SceptreContext(
--- a/sceptre/plan/actions.py
+++ b/sceptre/plan/actions.py
@@ -139,6 +139,10 @@ class StackActions:
                     {"Key": str(k), "Value": str(v)} for k, v in self.stack.tags.items()
                 ],
             }
+
+            if self.stack.disable_rollback:
+                update_stack_kwargs.update({"DisableRollback": self.stack.disable_rollback})
+
             update_stack_kwargs.update(self.stack.template.get_boto_call_parameter())
             update_stack_kwargs.update(self._get_role_arn())
             response = self.connection_manager.call(
--- a/sceptre/plan/actions.py
+++ b/sceptre/plan/actions.py
@@ -141,7 +141,9 @@ class StackActions:
             }
 
             if self.stack.disable_rollback:
-                update_stack_kwargs.update({"DisableRollback": self.stack.disable_rollback})
+                update_stack_kwargs.update(
+                    {"DisableRollback": self.stack.disable_rollback}
+                )
 
             update_stack_kwargs.update(self.stack.template.get_boto_call_parameter())
             update_stack_kwargs.update(self._get_role_arn())
--- a/tests/test_actions.py
+++ b/tests/test_actions.py
@@ -309,7 +309,9 @@ class TestStackActions(object):
                 "DisableRollback": True,
             },
         )
-        mock_wait_for_completion.assert_called_once_with(sentinel.stack_timeout, boto_response=ANY)
+        mock_wait_for_completion.assert_called_once_with(
+            sentinel.stack_timeout, boto_response=ANY
+        )
 
     @patch("sceptre.plan.actions.StackActions._wait_for_completion")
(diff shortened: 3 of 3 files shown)


alexharv074:
Integration test run https://github.com/Sceptre/sceptre/actions/runs/10226398355

sceptre/plan/actions.py
```diff
@@ -139,6 +139,12 @@ def update(self):
                     {"Key": str(k), "Value": str(v)} for k, v in self.stack.tags.items()
                 ],
             }
+
+            if self.stack.disable_rollback:
```
zaro0508, line 143: there's code in the above `create` method that says.. [can specify either DisableRollback or OnFailure , but not both](https://github.com/Sceptre/sceptre/blob/master/sceptre/plan/actions.py#L86).  Wouldn't this need to be aware of `OnFailure` as well?

zaro0508 (changes requested):

sceptre/plan/actions.py
alex-harvey-z3q: I think we should leave it as-is for now. Here's the problem:
In the case of [`create_stack`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation/client/create_stack.html), the Boto3 docs say:
> DisableRollback (boolean) –
> 
> Set to true to disable rollback of the stack if stack creation failed. You can specify either DisableRollback or OnFailure, but not both.
>
> Default: false

In the case of [`update_stack`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation/client/update_stack.html) however, there is no option for `OnFailure` and the docs simply say:

> DisableRollback (boolean) –
>
> Preserve the state of previously provisioned resources when an operation fails.
>
> Default: False

Ideally, I think the original feature that added `DisableRollback` to Sceptre should have implemented the same logic as specified in Boto3, however we did not. Instead we have simply given `DisableRollback` precedence over `OnFailure`. 

Now it gets better! Because although the code indicates in some places that we have support for `on_failure`, it has never been connected to the CLI! This situation appears to go back to 2018.

Thus there is no actual option for anyone to do something silly like

```
sceptre create --disable-rollback=true --on-failure=do-nothing
```

Because we have never actually implemented `on-failure` so far as I can tell. This deserves to be tracked as a separate tech debt ticket.

sceptre/cli/update.py
```diff
@@ -41,6 +41,7 @@ def update_command(
     :type verbose: bool
     :param yes: A flag to answer 'yes' to all CLI questions.
     :type yes: bool
+    :param disable_rollback: A flag to disable cloudformation rollback.
```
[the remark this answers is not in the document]
alex-harvey-z3q: What you did for the create command is shown here:
```python
def create_command(ctx, path, change_set_name, yes, disable_rollback: Optional[bool]):
    """
    Creates a stack for a given config PATH. Or if CHANGE_SET_NAME is specified
    creates a change set for stack in PATH.
    \f

    :param path: Path to a Stack or StackGroup
    :type path: str
    :param change_set_name: A name of the Change Set - optional
    :type change_set_name: str
    :param yes: A flag to assume yes to all questions.
    :type yes: bool
    :param disable_rollback: A flag to disable cloudformation rollback.
    """
    context = SceptreContext(
        command_path=path,
        command_params=ctx.params,
        project_path=ctx.obj.get("project_path"),
        user_variables=ctx.obj.get("user_variables"),
        options=ctx.obj.get("options"),
        ignore_dependencies=ctx.obj.get("ignore_dependencies"),
    )
```
Your thinking appears to be not to document the type in the docstring, but instead to use type hints to document the type. I agree with what you did there, and have copied that while updating the update command.

sceptre/plan/actions.py
zaro0508: yikes! although i guess nobody actually cared to complain about `on-failure`

zaro0508 (approved):

PR #1522, opened 2024-10-03
[Resolve #1519] Improve error message

### LINKED ISSUE #1519: When running sceptre diff, we get an `UnboundLocalError` due to lack of permissions, can we make the error clearer to the user?
### Subject of the issue
When running `sceptre diff some/stack.yaml` we get an `UnboundLocalError`, caused by lack of permissions, can we make the error clearer to the user?

```
  File "/home/jak/.pyenv/versions/3.12.3/envs/env1/lib/python3.12/site-packages/sceptre/plan/actions.py", line 1077, in diff
    return stack_differ.diff(self)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jak/.pyenv/versions/3.12.3/envs/env1/lib/python3.12/site-packages/sceptre/diffing/stack_differ.py", line 126, in diff
    deployed_config = self._create_deployed_stack_config(stack_actions)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jak/.pyenv/versions/3.12.3/envs/env1/lib/python3.12/site-packages/sceptre/diffing/stack_differ.py", line 206, in _create_deployed_stack_config
    print(description)
          ^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'description' where it is not associated with a value
```

### Your environment
* Sceptre, version 4.5.2
* Python 3.12.3
* Ubuntu 24.04, WSL, 5.15.153.1-microsoft-standard-WSL2

### Steps to reproduce
We use AWS SSO to manage access to accounts, and obtain temporary access to AWS via SCIM integration with Azure.

We encounter this error when running `sceptre diff some/stack.yaml`, when the temporary access is not active.
i.e. We do not currently have AWS permissions 
```
$ aws --profile dev sts get-caller-identity
An error occurred (ForbiddenException) when calling the GetRoleCredentials operation: No access
```

### Expected behaviour
We expect sceptre to fail because we don't have access, but we also expect it to make it clear the reason is because of receiving 403 Forbidden when interacting with the AWS API's.

### Actual behaviour
We see an `UnboundLocalError`, which is unhelpful and caused us to dig into the sceptre source code to see what was wrong.

### Cause of issue
In `_create_deployed_stack_config` of `sceptre/diffing/stack_differ.py` there is a try/except block, but the except only checks for `err.response["Error"]["Message"].endswith("does not exist")`

When the error is a permission error, `err` is instead:
```
{'Error': {'Message': 'No access', 'Code': 'ForbiddenException'}, 'ResponseMetadata': {'RequestId': '0bc1*******', 'HTTPStatusCode': 403, 'HTTPHeaders': {'date': 'Wed, 02 Oct 2024 10:35:33 GMT', 'content-type': 'application/json', 'content-length': '86', 'connection': 'keep-alive', 'access-control-expose-headers': 'RequestId, x-amzn-RequestId', 'requestid': '0bc********', 'server': 'AWS SSO', 'x-amzn-requestid': '0bc*********'}, 'RetryAttempts': 0}}
```

In this case, the code continues on to line 206 
```
stacks = description["Stacks"]
```
However `description` was never set and so we get `UnboundLocalError`

### Suggested Fix
I'm happy to raise a PR, if you approve of the following suggestion:
We can add another check for when `err.response["Error"]["Code"] == "ForbiddenException"` and raise an error in this case, or perhaps we should just allow `err` to be thrown since the code can not continue from here?
I.e.
```python
def _create_deployed_stack_config(
    self, stack_actions: StackActions
) -> Optional[StackConfiguration]:
    try:
        description = stack_actions.describe()
    except ClientError as err:
        # This means the stack has not been deployed yet
        if err.response["Error"]["Message"].endswith("does not exist"):
            return None
        else:
            raise err
  stacks = description["Stacks"]
```
Let me know how you'd prefer the fix to function, but I think it would be good to get the root issue (`403 Forbidden / No access`) up to the user.

Thanks,
Jak

dboitnot:
Resolves #1519 by:
- Raising un-handled exceptions which (I believe) should be the default behavior.
- Detecting ForbiddenException and wrapping it in a more informative message

## PR Checklist

## Other Information

### DIFF
--- a/sceptre/diffing/stack_differ.py
+++ b/sceptre/diffing/stack_differ.py
@@ -201,7 +201,7 @@ class StackDiffer(Generic[DiffType]):
             if err.response["Error"]["Code"] == "ForbiddenException":
                 raise SceptreException(
                     "ForbiddenException: Confirm your current AWS profile is authenticated",
-                    "and has the necessary access."
+                    "and has the necessary access.",
                 )
 
             # This means the stack has not been deployed yet
--- a/sceptre/diffing/stack_differ.py
+++ b/sceptre/diffing/stack_differ.py
@@ -197,9 +197,18 @@ class StackDiffer(Generic[DiffType]):
         try:
             description = stack_actions.describe()
         except ClientError as err:
+            # Check for AWS access exceptions
+            if err.response["Error"]["Code"] == "ForbiddenException":
+                raise SceptreException(
+                    "ForbiddenException: Confirm your current AWS profile is authenticated."
+                )
+
             # This means the stack has not been deployed yet
             if err.response["Error"]["Message"].endswith("does not exist"):
                 return None
+
+            # Unknown error, raise it as-is
+            raise err
 
         stacks = description["Stacks"]
         for stack in stacks:
--- a/sceptre/diffing/stack_differ.py
+++ b/sceptre/diffing/stack_differ.py
@@ -200,7 +200,8 @@ class StackDiffer(Generic[DiffType]):
             # Check for AWS access exceptions
             if err.response["Error"]["Code"] == "ForbiddenException":
                 raise SceptreException(
-                    "ForbiddenException: Confirm your current AWS profile is authenticated."
+                    "ForbiddenException: Confirm your current AWS profile is authenticated",
+                    "and has the necessary access."
                 )
 
             # This means the stack has not been deployed yet
--- a/tests/test_diffing/test_stack_differ.py
+++ b/tests/test_diffing/test_stack_differ.py
@@ -458,9 +458,9 @@ class TestStackDiffer:
         self.local_no_echo_parameters.append("hide_me")
 
         expected_generated_config = self.expected_generated_config
-        expected_generated_config.parameters[
-            "hide_me"
-        ] = StackDiffer.NO_ECHO_REPLACEMENT
+        expected_generated_config.parameters["hide_me"] = (
+            StackDiffer.NO_ECHO_REPLACEMENT
+        )
 
         self.differ.diff(self.actions)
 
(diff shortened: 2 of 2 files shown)


alex-harvey-z3q:
In general I am a bit concerned that while I support the general idea, the implementation here just will not scale to the handling of all exceptions in all places in the code and will not therefore be maintainable. E.g. the PR suggests we handle `ForbiddenException` but what about `ExpiredTokenException` and `NoCredentialsError` and every possible other exception the client can throw if you haven't set up your AWS config properly? 

I wonder if you should start by reviewing this much earlier PR 
https://github.com/Sceptre/sceptre/pull/997

dboitnot:
> In general I am a bit concerned that while I support the general idea, the implementation here just will not scale to the handling of all exceptions in all places in the code and will not therefore be maintainable.

I completely agree that this is addressing a narrow case and that a more general solution is needed. I would argue that:

1. The original implementation was broken, returning None when it should have raised the exception.
2. We needn't let perfect be the enemy of good. This PR fixes the bug and addresses an additional case which is affecting real users.
3. Misleading error messages are expensive and permissions issues are common.

C'mon, it's been forever since I submitted a PR. :laughing:

alex-harvey-z3q:
@dboitnot Here are some more thoughts:

- Agree that the bug should be fixed and I think I agree about re-raising the exception.
- At a high level design point of view, I think we need to think carefully about whether we want to in general handle exceptions by sending hints back to the caller about how to setup their AWS and so on. Because if we handle this one lone exception this way, and then don't do the same everywhere else, that just gives the tool the feeling of an inconsistent design.
- If we do choose to do this, then I think the exception should be handled here https://github.com/Sceptre/sceptre/blob/master/sceptre/plan/actions.py#L309-L320

alex-harvey-z3q:
Closing in favour of https://github.com/Sceptre/sceptre/pull/1530

PR #1571, opened 2026-01-26
Plugin support for Sceptre CLI path resolving

wieshka:
Path processor plugin system for Sceptre CLI.

This module provides a plugin architecture that allows external packages to
register path processors that can transform or expand path arguments passed
to Sceptre CLI commands.

This PR enabled existence of https://github.com/wieshka/sceptre-wildcard-plugin - a Sceptre plugin which adds glob/wildcard support for sceptre CLI commands.


wieshka:
Looking forward to pick this up as I just realised we are all internally using just my outdated fork.

